PR #7, opened 2018-11-30
Update guides to include Origami patterns.

notlee:
Proposals may not always be a new component. They may also be broader Origami patterns.
- [Primary Mixin](https://github.com/Financial-Times/origami-proposals/issues/6) (current proposal)
- [Remove the ability to customise classes across components](https://github.com/Financial-Times/origami-proposals/issues/4) (accepted proposal)

This PR updates the guides accordingly.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -3,39 +3,39 @@
 
 There are several ways to contribute to this proposal process:
 
-  - [Discuss existing component proposals which are under consideration](#discuss-existing-proposals)
-  - [Propose a new component](#propose-a-new-component)
+  - [Discuss existing proposals which are under consideration](#discuss-existing-proposals)
+  - [Propose a new component or pattern](#propose-a-new-component-or-pattern)
   - [Review a proposal](#review-a-proposal) (Origami team only)
 
 
 ## Discuss Existing Proposals
 
-You should feel free to discuss any existing component proposals which are in the [proposal backlog], and are in the "Proposed" or "Under Consideration" columns. Some examples of useful contributions to the discussion:
+You should feel free to discuss any existing proposals which are in the [proposal backlog], and are in the "Proposed" or "Under Consideration" columns. Some examples of useful contributions to the discussion:
 
-  - Asking questions about a component proposal
+  - Asking questions about a proposal
   - Answering questions from other contributors
+  - Sharing research relating to the proposal
   - Sharing examples or demos of a component or pattern that's similar
-  - Sharing research relating to the proposed component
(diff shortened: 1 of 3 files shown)


.github/CONTRIBUTING.md
```diff
@@ -3,39 +3,39 @@
...
-## Propose a New Component
+## Propose a New Component or Pattern
 
-Anybody can propose a new component for inclusion in Origami. This can be a completely new component or an existing one which would benefit from being maintained by the Origami team. We ask that you complete the steps below:
+Anybody can propose a new component or pattern for inclusion in Origami. This can be a completely new component, an existing component which would benefit from being maintained by the Origami team, or a pattern that would impact multiple components. We ask that you complete the steps below:
 
   1. ### Check the Registry
 
-     Before proposing a new component, please check and search the [Origami Registry] so that you're sure we haven't already built a similar component. If you find something and need changes made to it, then please open an issue on the component repository.
+     If proposing a new component, please check and search the [Origami Registry] so that you're sure we haven't already built a similar component. If you find something and need changes made to it, then please open an issue on the component repository.
```
gvonkoss: Should we add an `if` for a pattern as well? As  in, describe what kind of pattern and why it would be useful/applicable? Or would that be part of the issue 🤔

.github/CONTRIBUTING.md
```diff
@@ -44,23 +44,23 @@ Anybody can propose a new component for inclusion in Origami. This can be a comp
...
 
   3. ### Acceptance
 
-     If, after discussing the component, it's decided that Origami should take on the work then we should outline our reasoning in a comment on the issue. Then we move it to the "Accepted" column and add the appropriate labels.
+     If, after discussing the proposal, it's decided that Origami should take on the work then we should outline our reasoning in a comment on the issue. Then we move it to the "Accepted" column and add the appropriate labels.
 
-	 If the component is not going to be accepted then we should also provide some reasoning and move to the "Not Accepted" column. It's important to also note that if somebody in future has a need for the same component then we can reopen the proposal and accept further evidence and reasoning.
+	  If the proposal is not going to be accepted then we should also provide some reasoning and move to the "Not Accepted" column. It's important to also note that if somebody in future has a need for the same proposal then we can reopen the proposal and accept further evidence and reasoning.
```
gvonkoss:
```suggestion
	 If the proposal is not going to be accepted then we should also provide some reasoning and move to the "Not Accepted" column. It's important to also note that if somebody in future has a need for the same proposal then we can reopen the proposal and accept further evidence and reasoning.
```

gvonkoss: SPACES

PR #27, opened 2020-03-03
Move issue templates to folder

chee:
So this can be a holistic origami discussion repo!


chee:
hmm, this repo has no CODEOWNERS

chee:
hmm, this repo has no CODEOWNERS !

chee:
@notlee the `pattern` and `component` proposals have come from the current ISSUE_TEMPLATE.md, split here into two things instead of the `if/else` clauses in the current issue template. i.e., pattern is _not_ generic, it is specific to component patterns!

When you are creating an issue for now you will probably most often click the "Open a blank issue" link:

a screenshot of the "open a blank issue" link


We can add more templates as we notice patterns of issue-types emerge, but for now this preserves (and slightly improves?) the previous Origami Component Proposal system and lets us get on with our other tasks :)

PR #64, opened 2020-06-08
Fix proposal paths in contribution guide

chee:


notlee:
should a new proposal go under `accepted`? Is the idea that we assume it will be accepted and if not it's moved to a different directory like `rejected`?

chee:
the proposal should be added to `accepted`, we accept a proposal by merging it.

if it's not accepted then the PR will be closed!

PR #73, opened 2020-08-07
Proposal: Move all Origami component build tools under one repo

chee:
rendered: https://github.com/Financial-Times/origami/blob/8e7b86fd631968b22746e076caffeba88f13a85f/proposals/accepted/0073-origami-build-system.md

### DIFF
--- a/proposals/accepted/0073-origami-build-system.md
+++ b/proposals/accepted/0073-origami-build-system.md
@@ -27,6 +27,7 @@ origami-build-system
 │   │   ├── create-origami-component
 │   │   ├── eslint-config-origami-component
 │   │   ├── origami-build-tools
+│   │   ├── origami-bundle-size-cli
 │   │   ├── origami-ci-tools
 │   │   ├── remark-preset-lint-origami-component
 │   │   ├── sass
@@ -51,7 +52,9 @@ When package `x` depends on package `y`, its package.json dependencies section
 looks like this:
 
 ```json
-"dependencies":
+"dependencies": {
+	"y": "file:./y"
+}
 ```
 
 Athloi is more convenient to use if all the subpackages are at the same version,
--- a/proposals/accepted/0073-origami-build-system.md
+++ b/proposals/accepted/0073-origami-build-system.md
@@ -1,40 +1,146 @@
-# put your proposal title here
-
-> Provide a brief one line summary of your proposal.
+# create an origami component build system workspace
+let's put all of the packages related to building and testing origami components
+in one repository.
 
 ## motivation
+when making a change to something like eslint-config-origami-component, we often
+need to make the change there, and in origami-build-tools, and in origami-ci.
 
-> Explain why you think Origami should accept the proposal.
->
-> e.g. When proposing a new component:
-> - What evidence do you have that it's needed by multiple teams across the FT?
-> - What evidence do you have that it meets the needs of the users of those teams?
->
-> This section is required
+these are distinct packages with distinct purposes, but they are worked on
+together and released together.
+
+a concept called "workspaces" is a good approach for this.
 
 ## explanation
+We use [athloi](https://github.com/Financial-Times/athloi) to manage our npm
+workspaces, and Cargo to manage our rust workspaces.
 
-> Provide a detailed explanation of your proposal
-> This section is required
+We have a directory structure like this:
+
+```
+origami-build-system
+├── Cargo.toml
+├── origami.json
+├── package.json
+├── packages
+│   ├── node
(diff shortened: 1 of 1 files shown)


notlee:
>Athloi is more convenient to use if all the subpackages are at the same version, so we do a major release of v11 for every package when first making the change.

Does this mean going forward all packages have to be at the same major?

proposals/accepted/0073-origami-build-system.md
```diff
@@ -0,0 +1,146 @@
...
+origami-build-system
+├── Cargo.toml
+├── origami.json
+├── package.json
+├── packages
+│   ├── node
+│   │   ├── create-origami-component
+│   │   ├── eslint-config-origami-component
+│   │   ├── origami-build-tools
+│   │   ├── origami-ci-tools
```
chee, line 30:
```suggestion
│   │   ├── origami-build-tools
│   │   ├── origami-bundle-size-cli
│   │   ├── origami-ci-tools
```

proposals/accepted/0073-origami-build-system.md
```diff
@@ -0,0 +1,146 @@
...
+"workspaces": [
+	"packages/node/*"
+]
+```
+
+When package `x` depends on package `y`, its package.json dependencies section
+looks like this:
+
+```json
+"dependencies":
```
JakeChampion, line 54: Is this meant to have more in it such as this?
```suggestion
"dependencies": {
	"y": "file:./y"
}
```

chee: oh, woops! yes it is :D

proposals/accepted/0073-origami-build-system.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+Athloi is more convenient to use if all the subpackages are at the same version,
```
JakeChampion, line 60: I like the idea of having all packages at the same version, it makes it simpler to notice if they are out of sync in a project but I do also worry that it might annoy other projects which depend on the eslint config and keep getting dependabot/renovate pull-requests to update the version even though nothing inside that specific package has changed.

chee:
> > Athloi is more convenient to use if all the subpackages are at the same version, so we do a major release of v11 for every package when first making the change.
> 
> Does this mean going forward all packages have to be at the same major?

they don't _have_ to be, but it's more convenient if every package is on the same version and released together. the exact same version, not just the same major version. otherwise you need to make heavy use of athloi's `--filter`, and we'd need to be clever about making sure any changes cause version bumps to cascade out to the dependents.

lerna offers an [independent mode](https://github.com/lerna/lerna#independent-mode) which allows you to keep packages at different versions. we would still need to have a system for causing version bumps to cascade out to the dependents.

athloi's approach is less versatile, but more simple. the packages are part of a system, and the system gets a release.

proposals/accepted/0073-origami-build-system.md
```diff
@@ -0,0 +1,146 @@
...
+2. rewrite scrumple's package build (modeling it on the ft
+[sass](https://github.com/Financial-Times/sass) build)
+3. write a github action to build scrumple + the npm packages in the right order
+  (athloi takes care of the topological ordering of the npm build)
+4. do a major prerelease from this repository
+5. if bugs, fix bugs and goto 4
+6. do a major release of everything!
+7. archive all the other repos
+
+scrumple's `package.json#scripts.build` can maybe kick off the cargo build? not sure.
```
JakeChampion, line 98: is that to make it work via `athloi` or for another reason?

chee: no, this is just for the convenience of making sure there is a scrumple binary ready for you when you are working

it would be ideal to create a system where you could use `cargo watch`  and have the debug binary placed in the right spot so `obt` picks it up, so you can make code changes to obt and scrumple together.

proposals/accepted/0073-origami-build-system.md
chee: we can create a mildly clever script that checks if a package has had a change since the last major version, and use `athloi --filter` to only _publish_ those that have changed. `git log` can help us here

chee:
should i do some more exploration on this? like create a demo repo?

chee:
> > should i do some more exploration on this? like create a demo repo?
> 
> Yes, shall we put that work into the next cycle?

Yes!

JakeChampion:
I added the `pending` label to make sure we don't forget to discuss this in the next planning :-)

chee:
> I added the `pending` label to make sure we don't forget to discuss this in the next planning :-)

good call

PR #76, opened 2020-08-18
Let's put our component .github directory in a repository of its own, and include it with git subtree

chee:
rendered: https://github.com/Financial-Times/origami/blob/f9f54cecca9be58e1d5f7cac06b00598ec3f4ced/proposals/accepted/0076-dot-github-subtree.md


proposals/accepted/0076-dot-github-subtree.md
```diff
@@ -0,0 +1,138 @@
...
+
+
+<a id="a-generator"></a>
+
+### a generator
+
+instead of a repo that is subtree'd in, we create a command-line computer program that can be run in a repo to <span class="underline">generate</span> the origami `.github` directory for a particular project type. this is essentially the same as the proposal except the templates would all be kept on the project's main branch.
+for users who are diverging from our `.github` directory, they would see the changes in their `git diff` and could decide to keep them or not. we can also provide flags such as `--codeowner="origami-core"`, or `--without-issue-template` to let them generate the right one for their project.
+
+now that i'm typing it, this seems like it might actually be a better idea and maybe i should be proposing that instead?
```
JakeChampion, line 86: I think I like this idea more than git subtree.

We could have a comment at the top of all the generated files which states they were generated from a tool and where to go to update the canonical file.

chee: yes! that's a great idea.

notlee: I like this idea too, but how would it look if we wanted to add a workflow to all components? Wouldn't we still be in a position of running a shell script on our local machines to update all the things?

JakeChampion:
Loving this proposal :-)

One question I have which I think the proposal has not answered (not sure if it needs to answer it). How would we ensure that all the actively maintained components have their `.github` folder up-to-date? In the alternative approaches section it was noted that using Git Submodules would mean Dependabot would work to keep them up-to-date.

proposals/accepted/0076-dot-github-subtree.md
chee: You're right. the only difference would be the PRs would be updating the tool and running ~generate~ again, and the changes would be happening in a public repo with history, rather than in a throwaway script. but i think it is mostly the same, and actually not enough better. I'm going to reject this proposal. Please see my closing comment.

chee:
> Loving this proposal :-)
> 
> One question I have which I think the proposal has not answered (not sure if it needs to answer it). How would we ensure that all the actively maintained components have their `.github` folder up-to-date? In the alternative approaches section it was noted that using Git Submodules would mean Dependabot would work to keep them up-to-date.


Yes, this wouldn't help with that at all. Now that I've given it more thought, we should not do this and I'm going to reject my proposal. Please see my closing comment.

chee:
I am going to reject this proposal. There are two separate things it is trying to solve:

1.  Shared files such as `CODEOWNERS`, `ISSUE_TEMPLATE/` etc. these are actually brought in by `create-origami-component`, and when forking the template. there's no reason to try and solve them again.
2.  Workflows shared across components. It doesn't solve this well enough.
    
It also has, as Jake [pointed out](https://github.com/Financial-Times/origami/pull/76#issuecomment-677768717), a huge drawback in that it doesn't help us keep them up to date, that's still a manual process. Lee also mentioned to me on Slack that that's one of the things that really need fixed. I think that's it's a deal-breaker that this proposal doesn't help with that.

## a new proposal

I'm going to research a new proposal. By default, a GitHub workflow runs all of its jobs in parallel. Only when a job has dependencies does it wait to run. Instead of syncing all of our workflows for components, maybe we should have *only one*.

The `origami-component-action`.

We may be able to replace all of the workflows in our components with a single workflow that listens on all events (at least `issues`, `pull_request` and `release`) and calls on `origami-component-action`, which has jobs for each of the sub-actions that are currently defined in every repository.

We can also add shell scripts to that action using the `composite` run steps type

<https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobs>
<https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action>

Issue #54, opened 2020-05-07
Review how we manage and deploy component specifc assets


notlee:
Review how we build components with unique static assets such as images.

**Background:**

At the time of writing (May 2020) the vast majority of our components pull their images from our shared icon or logo [image sets](https://registry.origami.ft.com/components?imageset=true&active=true&maintained=true). 

However [o-comments](https://github.com/Financial-Times/o-comments/pull/164) has a unique image asset that is hosted in an s3 bucket (somewhere) and served through the [Image Service](https://www.ft.com/__origami/service/image/v2/). It's the best approach we have at the moment but it's not ideal: It relies on a component author having a secure and reliable way to host their own assets; the asset could be deleted accidentally as there's no explicit connection; the asset isn't within version control for restoring and history of the component; and it reduces our ability to develop offline (although along with fonts, polyfill.io requests, and potential build service requests in demos we can't really anyway).

We used to use [o-assets](https://github.com/Financial-Times/o-assets) to manage individual component assets as required [in the spec](https://origami.ft.com/spec/v1/components/#subresources), but it has a few problems:
- For local development, assumes assets are in `bower_components` by default, but: Origami components are also published to npm now and the asset directory cannot be changed to `node_modules` because the component directory has a `@financial-times` namespace; the asset directory may or may not be included in the npm bundle by [occ](https://github.com/Financial-Times/origami-component-converter/); and
- Requires apps in development and production to serve `bower_components/node_modules` source code so requests can be made for the assets (potentially with a url mapping behind a router by configuring o-assets variables).

**Quick brain bump of requirements:**
We need a way to include unique component assets (let's assume any static file be it images, videos, json blobs, font files) which:
- work for the Origami Build Service
- work for bower and npm users in development
- allows users to self host component assets; and or
- doesn't require projects to include an asset build step (Origami hosted assets)

notlee:
**Possible Solution(s). This is not fleshed out and feels very convoluted. Just wanted to get down some initial ideas and start a conversation.**

- Spec that assets must go in an `assets` directory.
- Before release, automate a commit to [set the components version](https://github.com/Financial-Times/origami/issues/48) as a JS and Sass variable.
- On tag publish the `assets` directory to an Origami S3 bucket by name and version.
- Use a simplified `o-assets`, which:
   - has a Sass and JS function to include assets
        - takes the component name
        - takes the asset path relative to the asset directory
        - outputs an absolute path to the Origami S3 bucket by default
[...]

PR #86, opened 2020-10-09
Proposal: npm-only origami components

chee:
Origami Components are written using the [Bower](https://bower.io/) package manager.
However  components may be included in projects in one of three ways:
- [Bower](https://bower.io/)
- [npm](https://www.npmjs.com/) (beta, not recommended)
- [Origami Build Service](https://www.ft.com/__origami/service/build/v2/)

This work-in-progress proposal is to drop Bower support, for both developing and consuming components.
See the proposal: https://github.com/Financial-Times/origami/blob/npm/proposals/accepted/0000-npm.md


proposals/accepted/0000-npm.md
```diff
@@ -43,15 +43,17 @@ We'll be bumping the `origamiVersion` from `1` to `"2.0.0"`, so our services and
...
 	- Ensure users are prepared to switch to npm and dart within x months, we'll need to maintain multiple versions of components in the meantime; and
 	- Communicate the value across the business
+- Origami CI Tools
+	- Update `origami-ci-tools` so it may publish v2 components, e.g. [without using occ](https://github.com/Financial-Times/origami-ci-tools/blob/9083d70dbba7da78a4ffcb8bc84211720e844cca/commands/release.js#L42)
+- Test Component
+	- Create v2 test components to support tool and service work
+	- Publish a `o-test-component` version with Sass and JavaScript, which conforms to spec v2, and depends on a new `o-test-component-dependency` v2 component
 - Build Tools
 	- Support v1 and v2 components conditionally by checking `origamiVersion`:
 	- Don't run `bower install` for v2 components
```
chee, line 53: `obt` actually already skips `bower install` if there is no `bower.json`:
https://github.com/Financial-Times/origami-build-tools/blob/master/lib/tasks/install-bower.js#L35-L40

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,102 @@
+# Proposal: npm-only origami components
+
+## motivation
+
+there is so much motivation!!
+
+- Teams across the FT are moving away from `bower`, and the Origami Team are meant to serve those teams.
+
+- Often<sup>[citation needed]</sup> the only reason someone<sup>[who]</sup> is installing `bower` is because they need to use Origami.
```
JakeChampion, line 9: Now that we have Origami Components on npm, we can see how many projects are using that instead of Bower, my assumption is that new greenfield projects are using npm more than Bower and that we can use that as an indication that it was Origami holding projects back from ditching Bower.

chee: good idea

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,102 @@
...
+
+- The `origami-component-converter` project that builds npm versions of our bower components is really confusing for users, and is horrible to maintain!
+
+- There are new ideas we have to make Origami easier to use. Switching to dart-sass, making hot-reloadable templates, multirepos, \[...others], that will be much easier to do when using npm and not needing to support bower anymore.
```
JakeChampion, line 13: This is because the majority of the frontend ecosystem of tooling has dropped support for Bower, which means we've had to make our own tooling such as Scrumple, which means more stuff to build and maintain just because we are using Bower

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,102 @@
...
+
+- dependencies we want to use that exist only on npm, we just don't use at the moment
```
JakeChampion, line 15: Examples of such packages are:
- `date-fns`
- `lit-html` maybe?
- `@popperjs/core` for replacing the o-tooltip positioning logic

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,102 @@
...
+
+- we currently maintain our own bundler because there aren't other modern bundlers that support both npm and bower.
```
JakeChampion, line 17: This is a real issue because making a bundler is a complex piece of work and something which needs maintenance and feature work, which means less time spent on making/maintaining/improving Origami Components.

chee: :100:

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,102 @@
...
+We stop supporting bower!
+
+We delete `bower.json` from our components and we convert them to npm components, moving the dependencies into the `package.json`.
+
+We'll be bumping the `origamiVersion` from `1` to `"2.0.0"`, so our services and tools know to process the component as an npm component.
+
+## work required
+
+- Specification
+	- Create Draft Specification
```
JakeChampion, line 34: Would it be possible, feasible for a proof-of-concept to be created which showcases how this future Origami Component world will work for both component development and for component usage? That would likely be a good way to help elicit feedback on the proposed approach

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,102 @@
...
+		- Require npm 7+
+		- Replace Bower references with npm
+		- Require `peerDependencies` are used, never `dependencies`, to ensure flatness
+		- @todo any other `package.json` restrictions to add or remove?
+		- Add that demos are not part of the Origami API explicitly, to support future work
+		- Require dart-sass, to support future work
```
JakeChampion, line 40: We would need to spec out a minimum set of sass load paths required to make Origami Components work.

If we want components to be imported by their full npm package name (E.G. `@financial-times/o-buttons`) then the minimum set of sass load paths would be `node_modules`.

If we want components to be imported by their component name (E.G. `o-buttons`) then the minimum set of sass load paths would be `node_modules/@financial-times` and `node_modules`. `node_modules` would still be required because of sass dependencies such as sass-mq which are used within our components

proposals/accepted/0000-npm.md
chee: the initial `o-test-component` scheme mentioned [here](https://github.com/Financial-Times/origami/pull/86/files/bfea21573908c8c754935f2a122d7572bbb53bf7#diff-2d5a7818cbb02856907b8845f3e6500b47a9468569c9619e3c26467368e984d7R48-R50), i.e. publishing a test component with no deps and another test component that depends on it, will be the closest we have for a proof-of-concept. that'll be where we see how the approach works.

after that, i think we should also publish betas of every component and use them somewhere like `registry-ui` to demonstrate things like a major cascade. that will be a good time to get lots of feedback and let people see it in action on a real app.

consumers of the build service shouldn't be affected.
bower users will have to migrate to npm or the build service.
npm users who are using npmv7 shouldn't notice a change either, because the `peerDependencies` should be installed automatically.

component development will initially not change, i think? this proposal will only mean you're working with the npm version and not the bower version, which won't necessarily . demos will remain the same, you'll still use `obt`, &c. 

the bigger changes to development and consumption will happen when we start doing things like moving to sass modules and changing how demos are defined, or creating an origami-components multirepo.

proposals/accepted/0000-npm.md
chee: we mention in [the current "Origami via npm" doc](https://origami.ft.com/docs/tutorials/npm/#how-to-get-origami-sass-to-work) that people should have `node_modules/@financial-times` and `node_modules` in their includePaths

my preference on this is that in Origami we start importing sass by the full npm package name "@financial-times/o-table" and recommend setting the includePaths to `["node_modules"]`.

Consumers who have set it the path to `["node_modules", "node_modules/@financial-times"]` will be able to continue importing as `o-buttons` and can migrate whenever they like.

chee:
proposal for this proposal:
let's remove this section from spec v2: https://origami.ft.com/spec/v1/components/markup/#templates

chee:
also i'm thinking maybe we don't need semver for the documentation, we can probably just use `v2.0`. we probably don't need patch-level for the spec

chee:
also maybe we don't need semver for the spec. `v2.0` should be enough, we probably don't need patch level.

proposals/accepted/0000-npm.md
chee, line 13:
```suggestion
- There are new ideas we have to make Origami easier to use that will be much easier to do when using npm and not needing to support bower anymore, because the tools for doing them do not exist on, or support, bower.
```

proposals/accepted/0000-npm.md
chee, line 9:
```suggestion
```

let's remove this line, i think there's enough about it in the line above

proposals/accepted/0000-npm.md
chee, line 17:
```suggestion
- we currently maintain our own bundler because there aren't other modern bundlers that support both npm and bower. A bundler is a complex piece of work, which means less time spent on work that more directly helps people
```

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,102 @@
...
+	- Version the specification, so the v1 specification is still available
+	- Publish the v2 specification
+	- Add a `origamiVersion` migration guide
+	- Update all documentation and tutorials to reflect v2. With references to v1, i.e. as asides or links to a document explaining the difference, where needed whilst users migrate
+- Components
+	- Update all Origami team maintained components to support `origamiVersion` v2, including `README` updates e.g. to remove any existing bower/npm reference and use the correct `import` path
+	- Work with other teams to update components they maintain. It may be they are less willing to maintain two versions and will release a major at the same time as upgrading end projects
+- Other Origami Projects
+	- Any project which an `origami.json` which isn't affected should release a `minor` which updates the `origamiVersion` to `v2`. This is so it appears up-to-date and is returned when filtering for `v2`, i.e. in the Registry or via Repo Data.
+
```
chee, line 91:
```suggestion

First steps we can take immediately, and in any order, are:

- Update the build service to use npm when the origamiVersion is `^2.0.0`.
- Update the build tools to correctly build demos, and assume npm is in use, when the `origamiVersion` is `^2.0.0`.
- Start a PR on the origami-website to add the v2 specification with the changes highlighted above.
- Create three test v2 components
  - A: one with no dependencies
  - B: one that has a peerDependency on A
  - C: one with peerDependencies on A and B
```

JakeChampion: >- Update the build service to use npm when the origamiVersion is `^2.0.0`.
This is difficult to do because it requires knowing the origamiVersion of a component before downloading and installing the component.
One idea I've had is that we have a list in the build service of the latest version of every component which has had an origamiVersion of 1 and state that new components have to be origamiVersion 2 or whatever the latest version is.
What do you think @Financial-Times/origami-core? Any ideas on other solutions for this?

chee: hmm.. we can surface the `origamiVersion` in repo data, build service could do a request for all v2 components? 

i think your idea sounds good.

this is a time when the whole component system being in a multirepo would be good because we could publish the metadata of everything into a .json file for the build service any time a component was updated

notlee: yeah I think that's good Jake, the list should only include majors so we can release patches to previous majors if needed

proposals/accepted/0000-npm.md
JakeChampion: If we renamed the scss entry file from `main.scss` to `_index.scss` then the way to import/use the component would be `@import 'o-icons';` instead of `@import 'o-icons/main';`

We could even choose to not rename the entry file at first, instead we could add `_index.scss` as a new file which only contains `@import './main';`, that way our consumers can use both `@import 'o-icons';` and `@import 'o-icons/main';`

Documentation on using index files --> https://sass-lang.com/documentation/at-rules/use#index-files

JakeChampion:
> also maybe we don't need semver for the spec. `v2.0` should be enough, we probably don't need patch level.

Agreed 👍

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,102 @@
...
+	- Add tests to cover Origami Build Tools features with v1 and v2 components
+- Build Service
+	- Install the component, e.g:
+		- Attempt an npm install, fall back to bower on error; or
+		- Make a request to Repo Data and install via npm if v2; or
+		- Install via npm by default. Maintain a list of all v1 component major versions to install via bower
+	- Build the component:
+		- Check the Origami version in origami.json
+		- Conditionally build the component based on the src for v1 or npm bundle for v2
+		- @todo what does the above involve?
```
JakeChampion, line 62: Do we want to allow a bundle which contains both spec v1 and v2 components?

Do we want to allow a spec v2 component to depend on a spec v1 component?

Do we want to allow a spec v1 component to depend on a spec v2 component?

JakeChampion:
Does this mean we would stop using `occ` (Origami Component Converter) and stop publishing a `dist` directory?

JakeChampion:
I wanted to test that this proposal of using peerDependencies works so I cloned all the components and converted them to this proposal format and then tested those components on polyfill-service and everything worked :-)

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,101 @@
...
+
+- There are new ideas we have to make Origami easier to use that will be much easier to do when using npm and not needing to support bower anymore, because the tools for doing them do not exist on, or support, bower.
+
+- dependencies we want to use that exist only on npm, we just don't use at the moment
+
+- we currently maintain our own bundler because there aren't other modern bundlers that support both npm and bower. A bundler is a complex piece of work, which means less time spent on work that more directly helps people
+
+- the javascript ecosystem has moved on, and we haven't.
+
+- the npm client is now ready to support us. npmv7's automatic install of `peerDependencies` will allow us to maintain a flat origami tree in apps.
```
notlee, line 20: - It makes it easier to setup a new project with Origami, since npm is already used by projects, engineers are more familiar with npm, and npm is better supported by the JavaScript ecosystem (as mentioned) without special setup requirements.
- Onboarding is also easier since npm is already used by projects anyway and engineers are more familiar with npm

notlee: - Our documentation assumes bower but some people are already using npm, which can lead to confusion and wasted time. By decoming npm only are documentation becomes more simple.

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,101 @@
...
+
+- The `origami-component-converter` project that builds npm versions of our bower components is really confusing for users, and is horrible to maintain!
+
+- There are new ideas we have to make Origami easier to use that will be much easier to do when using npm and not needing to support bower anymore, because the tools for doing them do not exist on, or support, bower.
+
+- dependencies we want to use that exist only on npm, we just don't use at the moment
+
+- we currently maintain our own bundler because there aren't other modern bundlers that support both npm and bower. A bundler is a complex piece of work, which means less time spent on work that more directly helps people
+
+- the javascript ecosystem has moved on, and we haven't.
```
notlee, line 18:
```suggestion
- The javascript ecosystem has moved on, and we haven't. We may not be able to use some opensource projects for use within components, without complication. Meaning we need to build more aspects of a component from scratch. The same is true for tooling, e.g. the bundler mentioned previously.
```

proposals/accepted/0000-npm.md
chee: that's cool, I'm into adding a _index that imports main!!

proposals/accepted/0000-npm.md
JakeChampion: Good additions, let's add them to the doc 👍

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,101 @@
...
+
+- the npm client is now ready to support us. npmv7's automatic install of `peerDependencies` will allow us to maintain a flat origami tree in apps.
+
+## explanation
+
+We stop supporting bower!
+
+We delete `bower.json` from our components and we convert them to npm components, moving the dependencies into the `package.json`.
```
JakeChampion, line 26: We will need to update the implementation of o-assets to start using `node_modules` instead of `bower_components`.

Currently even the npm-built version of o-assets uses `bower_components`, which I think means it has never really worked correctly for our npm users.


https://unpkg.com/browse/@financial-times/o-assets@3.4.8/browser.js

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,101 @@
...
+- Teams across the FT are moving away from `bower`, and the Origami Team are meant to serve those teams.
+
+
+- The `origami-component-converter` project that builds npm versions of our bower components is really confusing for users, and is horrible to maintain!
+
+- There are new ideas we have to make Origami easier to use that will be much easier to do when using npm and not needing to support bower anymore, because the tools for doing them do not exist on, or support, bower.
+
+- dependencies we want to use that exist only on npm, we just don't use at the moment
+
+- we currently maintain our own bundler because there aren't other modern bundlers that support both npm and bower. A bundler is a complex piece of work, which means less time spent on work that more directly helps people
```
notlee, line 16: we can also deprecate our bower registry

notlee: this should be a task

proposals/accepted/0000-npm.md
JakeChampion: A proposal was made by Lee about updating o-assets' implementation already -- https://github.com/Financial-Times/origami/issues/54

proposals/accepted/0000-npm.md
JakeChampion: Draft spec is here -- https://github.com/Financial-Times/origami-website/pull/273

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,101 @@
...
+We'll be bumping the `origamiVersion` from `1` to `"2.0.0"`, so our services and tools know to process the component as an npm component.
+
+## work required
+
+- Specification
+	- Create Draft Specification
+		- Require npm 7+
+		- Replace Bower references with npm
+		- Require `peerDependencies` are used, never `dependencies`, to ensure flatness
+		- @todo any other `package.json` restrictions to add or remove?
```
JakeChampion, line 37:
```suggestion
		- All packages which export client-side JavaScript **must** set the [browser](https://docs.npmjs.com/files/package.json#browser) field in `package.json` to have the name of the entry point file.
		- @todo any other `package.json` restrictions to add or remove?
```

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,102 @@
...
+
+- Specification
+	- Create Draft Specification
+		- Require npm 7+
+		- Replace Bower references with npm
+		- Require `peerDependencies` are used, never `dependencies`, to ensure flatness
+		- All packages which export client-side JavaScript **must** set the [browser](https://docs.npmjs.com/files/package.json#browser) field in `package.json` to have the name of the entry point file.
+		- @todo any other `package.json` restrictions to add or remove?
+		- Add that demos are not part of the Origami API explicitly, to support future work
+		- Require dart-sass, to support future work
```
JakeChampion, line 40:
```suggestion
		- Rewrite sass imports of origami components to use the full npm package. E.G. `@financial-times/o-colors` instead of `o-colors`. This will enable a simpler sass configuration for our users.
		- Require dart-sass, to support future work
```

proposals/accepted/0000-npm.md
chee: :+1:

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,104 @@
...
+	- It may be that no other changes are required since repo data already supports non-component projects. Add tests to confirm.
+	- E.g. that for a v2 component `package` is added to the `manifests` key, bower is null as with non-component projects
+	- E.g. npm dependencies, including peer dependencies, are returned in repository/version entities
+	- E.g. v1 behaviour doesn't change
+	- Add an `origamiVersion` filter to the [repos endpoint](https://origami-repo-data.ft.com/v1/docs/api/repositories#get-v1-repos)
+- Registry UI
+	- Test the registry displays `o-test-component`, `o-test-component-dependency` correctly. This may not be the case if, for example, the Registry assumes the [repo data registry/version entity](https://origami-repo-data.ft.com/v1/docs/api/repositories#entity-repo) has a bower manifest. It doesn't appear to make such an assumption. The registry may just work since it also deals with non-components. E.g. it will [already display npm dependencies](https://github.com/Financial-Times/origami-registry-ui/blob/42fac8202a5c9f7f960ec4091bbedd78903cc9c4/lib/routes/components.js#L283) provided Repo Data is updated.
```
notlee, line 76: Component pages include bower related install instructions. These should remain the same for v1 components but show npm instructions for v2 components. The registry should be able to do this with data from repo data
Screenshot 2020-10-22 at 16 55 52

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,122 @@
...
+we can continue being an old-fashioned drag and suffer the consequences
+
+## supporting examples
+
+i don't know... every other modern component system i guess
+
+## notes
+
+### migration steps for products coming from bower
+
```
JakeChampion, line 107:
```suggestion

- Firstly: update to the latest version of the Origami Components which is available via Bower.
```

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,123 @@
...
+E.G:
+```diff
+-import oTracking from 'o-tracking';
++import oTracking from '@financial-times/o-tracking';
+```
+- remove `bower_components` from the sass include-paths and add `node_modules`
+- rewrite all sass `@include`s of Origami packages to use the full npm package name.
+```diff
+-@include 'o-buttons';
++@include '@financial-times/o-buttons';
```
JakeChampion, line 122:
```suggestion
-@import 'o-buttons';
+@import '@financial-times/o-buttons';
```

proposals/accepted/0000-npm.md
JakeChampion:
```suggestion
	- Create Draft Specification -- https://github.com/Financial-Times/origami-website/pull/273
```

proposals/accepted/0000-npm.md
JakeChampion:
```suggestion
- the npm client is now ready to support us. npmv7's automatic install of `peerDependencies` will allow us to maintain a flat origami tree in apps.
- It makes it easier to setup a new project with Origami, since npm is already used by projects, engineers are more familiar with npm, and npm is better supported by the JavaScript ecosystem (as mentioned) without special setup requirements.
- Onboarding is also easier since npm is already used by projects anyway and engineers are more familiar with npm
- Our documentation assumes bower but some people are already using npm, which can lead to confusion and wasted time. By becoming npm only our documentation becomes more simple.
```

proposals/accepted/0000-npm.md
notlee: Should this be separate on our backlog?

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,143 @@
+# Proposal: npm-only origami components
+
+## motivation
+
+there is so much motivation!!
```
notlee, line 5:
```suggestion
There is so much motivation:
```
Fun is not allowed

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,143 @@
...
+
+- The javascript ecosystem has moved on, and we haven't. We may not be able to use some opensource projects within components, without complication. Meaning we need to build more aspects of a component from scratch. The same is true for tooling, e.g. the bundler mentioned previously.
+
+- The npm client is now ready to support us. npmv7's automatic install of `peerDependencies` will allow us to maintain a flat origami tree in apps.
+
+- It makes it easier to setup a new project with Origami, since npm is already used by projects, engineers are more familiar with npm, and npm is better supported by the JavaScript ecosystem (as mentioned) without special setup requirements.
+
+- Onboarding is also easier since npm is already used by projects anyway and engineers are more familiar with npm
+
+- Our documentation assumes bower but some people are already using npm, which can lead to confusion and wasted time. By becoming npm only are documentation becomes more simple.
```
notlee, line 25:
```suggestion
- Our documentation assumes bower but some people are already using npm, which can lead to confusion and wasted time. By becoming npm only Origami documentation becomes more simple.
```

proposals/accepted/0000-npm.md
notlee: I think so, since its already a problem

proposals/accepted/0000-npm.md
notlee: >Do we want to allow a bundle which contains both spec v1 and v2 components?

[No.](https://github.com/Financial-Times/origami-build-service/pull/383#discussion_r508463024) 

>Do we want to allow a spec v2 component to depend on a spec v1 component?

I think yes? At least its not worth the effort to stop? If it builds ok with `obt` and passes tests, fine?

>Do we want to allow a spec v1 component to depend on a spec v2 component?

That seems an unrealistic scenario but: I think yes but strongly discouraged? At least its not worth the effort to stop? If it builds ok with `obt` and passes tests, fine?

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,138 @@
...
+	- Display a v2 badge on the component listing and in the right hand sidebar on the component page.
+- Origami Workshop
+	- Confirm the [origami workshop](https://github.com/Financial-Times/origami-workshop) CLI works with `origamiVersion` v2 components
+- Create Origami Component
+	- Update [create-origami-component](https://github.com/Financial-Times/create-origami-component) to create a v2 component
+- Origami Website
+	- Version the specification, so the v1 specification is still available
+	- Publish the v2 specification
+	- Add a `origamiVersion` migration guide
+	- Update all documentation and tutorials to reflect v2. With references to v1, i.e. as asides or links to a document explaining the difference, where needed whilst users migrate
```
JakeChampion, line 93:
```suggestion
	- Update all documentation and tutorials to reflect v2. With references to v1, i.e. as asides or links to a document explaining the difference, where needed whilst users migrate
- Bower Registry
	- Deprecate it
```

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,143 @@
...
+
+- There are new ideas we have to make Origami easier to use that will be much easier to do when using npm and not needing to support bower anymore, because the tools for doing them do not exist on, or support, bower.
+
+- Dependencies we want to use that exist only on npm, we just don't use at the moment.
+
+- We currently maintain our own bundler because there aren't other modern bundlers that support both npm and bower. A bundler is a complex piece of work, which means less time spent on work that more directly helps people.
+
+- The javascript ecosystem has moved on, and we haven't. We may not be able to use some opensource projects within components, without complication. Meaning we need to build more aspects of a component from scratch. The same is true for tooling, e.g. the bundler mentioned previously.
+
+- The npm client is now ready to support us. npmv7's automatic install of `peerDependencies` will allow us to maintain a flat origami tree in apps.
```
JakeChampion, line 19: We need to decide whether the `dependencies` field in `package.json` is entirely disallowed, forcing all dependencies for an origami component to be declared in the `peerDependencies` field or whether we want to allow the `dependencies` field but state that dependencies which are themselves origami components **must** be in `peerDependencies`.

The first choice, which disallows `dependencies` field would look like this:
```json
{
	"peerDependencies": {
		"lodash": "^4",
		"@financial-times/o-forms": "^9"
	}
}
```

The second choice, which allows `dependencies` for non-origami component dependencies would look like this:
```json
{
	"dependencies": {
		"lodash": "^4"
	},
	"peerDependencies": {
		"@financial-times/o-forms": "^9"
	}
}
```

@Financial-Times/origami-core, what do you think we should choose here?

JakeChampion: I think the first choice is better because it allows a product which uses origami components to be able to use a different version of a library that an origami component might also depend upon.

For instance they might use an older version of lodash (`^3`) in their project and use origami components which use a newer version of lodash (`^4`), and npm will install perfectly fine.

Origami would still be protected from not having two versions of a single component in the bundle whilst giving the products the control over everything else in their dependency tree.

notlee: Good points. Ideally only one copy of `lodash` would be included but we wouldn't want to force users into the same version we use, and wouldn't want upgrading `lodash` to be a major release. I think you're right, the first choice.

proposals/accepted/0000-npm.md
notlee: https://github.com/Financial-Times/origami-build-service/pull/383

proposals/accepted/0000-npm.md
notlee: > new greenfield projects are using npm more than Bower

Maybe not for customer products since pagekit uses Bower. In more evidence for the extra work/confusion bower causes: PageKit [does some magic to install bower dependencies](https://github.com/Financial-Times/dotcom-page-kit/blob/a25325d2e534895ac386c508af43a50e9ccf4b56/scripts/link-packages-using-bower.js) which can cause some confusion at times, like`bower ls` is no longer useful.

proposals/accepted/0000-npm.md
JakeChampion: sass dependencies **must**** be declared in `peerDependencies` and **must not** be declared in `dependencies` because sass does not handle having two different versions of a dependency.

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,143 @@
...
+
+## work required
+
+- Specification
+	- Create Draft Specification -- https://github.com/Financial-Times/origami-website/pull/273
+		- Require npm 7+
+		- Replace Bower references with npm
+		- Require `peerDependencies` are used, never `dependencies`, to ensure flatness
+		- All packages which export client-side JavaScript **must** set the [browser](https://docs.npmjs.com/files/package.json#browser) field in `package.json` to have the name of the entry point file.
+		- @todo any other `package.json` restrictions to add or remove?
```
JakeChampion, line 43: musts:
- license field exists
- `type` set to `"module"`
- export map which has imports
- browser field which points to entrypoint file
- module field which points to entrypoint file
- bugs field
- repository field

shoulds:
- homepage field which points to registry at the specific version (or latest version) ?

JakeChampion: we should state that the published javascript is compliant ecmascript syntax I.E. not commonjs format and not using any experiment/made-up syntax that requires a compiler to be used, such as css-in-js or importing css/image files into javascipt files.

_our javascript should just work in a modern standards compliant browser_

JakeChampion: > _our javascript should just work in a modern standards compliant browser_

that means that import statements should include the file extension of the module they are importing

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,143 @@
...
+		- Add that demos are not part of the Origami API explicitly, to support future work
+		- Rewrite sass imports of origami components to use the full npm package. E.G. `@financial-times/o-colors` instead of `o-colors`. This will enable a simpler sass configuration for our users.
+		- Add `_index.scss` files to make it possible to import Origami component sass without needing to name a specific file. E.G. let's make `@import('@financial-times/o-colors')` work.
+		- Require dart-sass, to support future work
```
JakeChampion, line 47:
```suggestion
		- Require components' sass is able to be compiled with the primary implementation of sass. Right now that is [Dart Sass](https://sass-lang.com/dart-sass)
```

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,143 @@
...
+## work required
+
+- Specification
+	- Create Draft Specification -- https://github.com/Financial-Times/origami-website/pull/273
+		- Require npm 7+
+		- Replace Bower references with npm
+		- Require `peerDependencies` are used, never `dependencies`, to ensure flatness
+		- All packages which export client-side JavaScript **must** set the [browser](https://docs.npmjs.com/files/package.json#browser) field in `package.json` to have the name of the entry point file.
+		- @todo any other `package.json` restrictions to add or remove?
+		- Add that demos are not part of the Origami API explicitly, to support future work
```
JakeChampion, line 44: What do we mean by this? The `demos` field in origami.json is [already optional](https://origami.ft.com/spec/v1/manifest/#demos)

screenshot showing the part of the spec which states the demos field is optional

proposals/accepted/0000-npm.md
chee: @JakeChampion in your second comment i think you said "first choice" when you meant "second choice"

i think it's only origami components (and, yes, sass deps) that must be in `peerDependencies`. other dependencies should work fine as dependencies.

i don't know how we'll talk about it in the documentation, but that's for later.

JakeChampion: > @JakeChampion in your second comment i think you said "first choice" when you meant "second choice"

🤦 yep, you are right, I meant "second choice"

proposals/accepted/0000-npm.md
chee: i like this a lot

proposals/accepted/0000-npm.md
JakeChampion: The origami components which have non-origami javascript dependencies are:
o-audio: `ftdomdelegate`
o-cookie-message: `superstore-sync`
o-date: `@financial-times/ft-date-format`
o-errors: `raven-js`
o-gallery: `ftdomdelegate`, `ftscroller`
o-overlay: `ftdomdelegate`
o-share: `ftdomdelegate`
o-syntax-highlight: `prismjs`
o-table: `ftdomdelegate`
o-tooltip: `ftdomdelegate`
o-topper: `@financial-times/n-map-content-to-topper`
o-tracking: `ftdomdelegate`
o-typography: `fontfaceobserver`

chee: o-cookie-message, by the way, doesn't actually seem to ever import superstore-sync? https://github.com/Financial-Times/o-cookie-message/search?q=superstore

proposals/accepted/0000-npm.md
JakeChampion: I've added these to the spec-v2 pr 

https://github.com/Financial-Times/origami-website/pull/273/commits/c01f31e6b239a9367ce0d8b031a1f02216be4440

https://github.com/Financial-Times/origami-website/pull/273/commits/b3906685c79b0059d1c18c80971f02257581f243

JakeChampion:
```suggestion
		- `license` field exists
		- [`type`](https://nodejs.org/api/packages.html#packages_type) set to `"module"`
		- [`export` map](https://nodejs.org/api/packages.html#packages_package_entry_points) which has `browser` fields defined. This stops products being able to import files which are private/internal to the component.
		- [`browser` field](https://docs.npmjs.com/files/package.json#browser) which points to entrypoint file
		- `bugs` field
		- `repository` field
		- `homepage` field which points to registry at the specific version (or latest version) ?
		- `types` field which points to a typescript definition file
		- origami component javascript should work directly in a modern standards compliant browser, without requiring a compilation step

```

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,152 @@
...
+
+We'll be bumping the `origamiVersion` from `1` to `"2.0.0"`, so our services and tools know to process the component as an npm component.
+
+## work required
+
+- Specification
+	- Create Draft Specification -- https://github.com/Financial-Times/origami-website/pull/273
+		- Require npm 7+
+		- Replace Bower references with npm
+		- Require `peerDependencies` are used, never `dependencies`, to ensure flatness
```
JakeChampion, line 41:
```suggestion
		- Require `peerDependencies` are used for all dependencies which are origami components or are dependencies for the sass, as these types of dependencies must not have multiple different versions included.
		- Require `dependencies` are used for any production javascript dependencies which are not also Origami components.
```

proposals/accepted/0000-npm.md
JakeChampion: This has now been added to the proposal in commit -- d88bb09b028152eb8e7d912b13dca2de06dff28f

chee: nice :D

proposals/accepted/0000-npm.md
notlee: That Origami JS, Sass, or mustache must not be used within a project

JakeChampion: > That demo JS, Sass, or mustache must not be used within a project

Has that happened in a project/product? 😱

JakeChampion:
We have a [temporary private npm registry](https://origami-npm-registry-prototype.herokuapp.com/) which we are using to host the components whilst we figure out what we want the component spec to be.

We can use this registry to trial the components on real-world Origami products before shipping them to our users. Real-world products such as:
- Origami Registry -- currently uses components via Bower
- Polyfill Service -- currently uses components via npm
- Build Service -- currently uses components via the Build Service

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,153 @@
...
+
+We'll be bumping the `origamiVersion` from `1` to `"2.0.0"`, so our services and tools know to process the component as an npm component.
+
+## work required
+
+- Specification
+	- Create Draft Specification -- https://github.com/Financial-Times/origami-website/pull/273
+		- Require npm 7+
+		- Replace Bower references with npm
+		- Require `peerDependencies` are used for all dependencies which are origami components or are dependencies for the sass, as these types of dependencies must not have multiple different versions included.
```
wheresrhys, line 41: If a dependency declares something as a dependency that an origami component only wants as a peer dependency, could that cause issues? If so, can they be worked around?

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,153 @@
...
+
+We delete `bower.json` from our components and we convert them to npm components, moving the dependencies into the `package.json`.
+
+We'll be bumping the `origamiVersion` from `1` to `"2.0.0"`, so our services and tools know to process the component as an npm component.
+
+## work required
+
+- Specification
+	- Create Draft Specification -- https://github.com/Financial-Times/origami-website/pull/273
+		- Require npm 7+
```
wheresrhys, line 39: Any opinion on ESM vs CJS? Could it also require node 14+ in order to confidently embrace ESM.

Maybe out of scope (or already settled?), but the whole ESM vs CJS dance is a pain when publishing modules, so possibly worth having a policy while the whole relationship with npm is being revisited anyway

JakeChampion: Our spec states component js must work directly in a browser, which rules out CJS completely

proposals/accepted/0000-npm.md
JakeChampion: We are testing this out 👍

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,153 @@
...
+
+## explanation
+
+We stop supporting bower!
+
+We delete `bower.json` from our components and we convert them to npm components, moving the dependencies into the `package.json`.
+
+We'll be bumping the `origamiVersion` from `1` to `"2.0.0"`, so our services and tools know to process the component as an npm component.
+
+## work required
```
wheresrhys, line 35: Just throwing it out there that a codemod tool or similar to reduce engineering effort by other teams might make it easier to sell this to the business

proposals/accepted/0000-npm.md
JakeChampion: The only things origami components will declare as peer-dependencies will be things which are either origami components themselves or are meant to be used in sass.

We do not want to allow those things to be worked around because it will cause hard to diagnose bugs in the end product.

proposals/accepted/0000-npm.md
JakeChampion: The migration effort for products should be minimal, we are testing the migration paths on our own products to ensure that is the case.

The migration for a product using origami via npm has been tested here -- https://github.com/Financial-Times/polyfill-service/pull/2440

The migration for a product using origami via bower has been tested here -- https://github.com/Financial-Times/origami-registry-ui/pull/452

We are currently working out the migration path (if one is needed) for products using origami via the build service

proposals/accepted/0000-npm.md
chee: no! i don't think so, but i think we wanted it to be explicit so it wouldn't be a major change if we decided to refactor how demos work.

but now on revisiting this, i think that amounts to making demos non-normative... which is maybe not ideal? 

it's hard. we want to be able to rethink demos from the ground up without causing a major cascade when the only consumer of demos is the Origami Registry UI. maybe we should make the demo specification separate, and say in Component Spec v2 that it supports Demo Spec v1, and then in a minor Component Spec update we can add support for Demo Spec v1 and v2.

That way we can change how demos work without causing a breaking change to component consumers. 

The version in the Component Spec has 2 jobs, one is a promise to component consumers that the things enumerated upon in the spec won't change without a breaking change. The other promise is that component creators can make things that work with our build system, command line tools, registry etc.

In this case we want to avoid the situation where a change in how components are built causes a ripple of upgrades for users when there have been no changes from their perspective.

JakeChampion: @chee Thank you, that is an excellent answer :D

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,153 @@
...
+		- Require `dependencies` are used for any production javascript dependencies which are not also Origami components.
+		- All packages which export client-side JavaScript **must** set the [browser](https://docs.npmjs.com/files/package.json#browser) field in `package.json` to have the name of the entry point file.
+		- `license` field exists
+		- [`type`](https://nodejs.org/api/packages.html#packages_type) set to `"module"`
+		- [`export` map](https://nodejs.org/api/packages.html#packages_package_entry_points) which has `browser` fields defined. This stops products being able to import files which are private/internal to the component.
+		- [`browser` field](https://docs.npmjs.com/files/package.json#browser) which points to entrypoint file
+		- `bugs` field
+		- `repository` field
+		- `homepage` field which points to registry at the specific version (or latest version) ?
+		- `types` field which points to a typescript definition file
```
JakeChampion, line 51: From my testing, this will only work if we have the `main` field set.

Here is an example where I set the main field in o-tracking/package.json and not in o-cookie-message/package.json

image

image

JakeChampion: @Financial-Times/origami-core hopefully there is a way to get the `types` field to work without having to also set the `main` field. The reason we didn't want to set the `main` field was to ensure users could not require/import the module within a server context such as nodejs

chee: how strange! does it work if the main field is set to an empty javascript file?

JakeChampion: > how strange! does it work if the main field is set to an empty javascript file?
I don't think so :/
image

chee: how strange, isn't that the same behaviour as if there was no `types` field then? like, isn't it just reading the JSDoc from the javascript files?

proposals/accepted/0000-npm.md
wheresrhys: might be worth seeing how complex a PR on one of the ft.com apps looks - far more distributed use of origami imports IIRC

JakeChampion: We will be doing that, PageKit is something we are looking closely at 👍

proposals/accepted/0000-npm.md
JakeChampion: No, it seems without the `types` and `main` fields, vscode/typescript doesn't work correctly.

This is a version where I deleted the package.json.types field from o-tracking:
image

proposals/accepted/0000-npm.md
JakeChampion: @chee, could you add this demos part to the [spec v2 pull-request](https://github.com/Financial-Times/origami-website/pull/273)

chee: yes

proposals/accepted/0000-npm.md
notlee: >Our spec states component js must work directly in a browser, which rules out CJS completely

I assume we're referring to the published bundle? Maybe we should clarify that?

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,153 @@
...
+- Create Origami Component
+	- Update [create-origami-component](https://github.com/Financial-Times/create-origami-component) to create a v2 component
+- Origami Website
+	- Version the specification, so the v1 specification is still available
+	- Publish the v2 specification
+	- Add a `origamiVersion` migration guide
+	- Update all documentation and tutorials to reflect v2. With references to v1, i.e. as asides or links to a document explaining the difference, where needed whilst users migrate
+- Bower Registry
+	- Deprecate the Bower registry.
+- Components
```
JakeChampion, line 106:
```suggestion
	- Deprecate the Bower registry.
- [Origami Readme Linter](https://github.com/Financial-Times/remark-preset-lint-origami-component/)
	- Make the name rule use package.json.name instead of bower.json.name
- Components
```

proposals/accepted/0000-npm.md
JakeChampion: > > Our spec states component js must work directly in a browser, which rules out CJS completely
> 
> I assume we're referring to the published bundle? Maybe we should clarify that?

Interesting point! Because the current spec was built around bower which does not installs directly from a git repository our spec was covering both source code and published code.

The spec should probably only care about the code which is published and not how the source code is authored.

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,155 @@
...
+		- `bugs` field
+		- `repository` field
+		- `homepage` field which points to registry at the specific version (or latest version) ?
+		- `types` field which points to a typescript definition file
+		- origami component javascript should work directly in a modern standards compliant browser, without requiring a compilation step
+
+		- Add that demos are not part of the Origami API explicitly, to support future work
+		- Rewrite sass imports of origami components to use the full npm package. E.G. `@financial-times/o-colors` instead of `o-colors`. This will enable a simpler sass configuration for our users.
+		- Add `_index.scss` files to make it possible to import Origami component sass without needing to name a specific file. E.G. let's make `@import('@financial-times/o-colors')` work.
+		- Require components' sass is able to be compiled with the primary implementation of sass. Right now that is [Dart Sass](https://sass-lang.com/dart-sass)
```
JakeChampion, line 57: Currently the way to use an origami components' sass is like this:
```scss
@import '@financial-times/o-table';
include oTable();
```
The build service creates that file when asked for o-tables's css and then compiles that the sass compiler.

If our components ever change the way they are imported, we will likely need to create a new version of the build service.

I'm mentioning this because at some point we were thinking about changing origami component's to switch from `@import` to `@use`, if we make that change, it'd be great if there was a consistent way to include a component's sass.

something like this is probably all we need:
```scss
@use '@financial-times/o-table'; 
```

chee: the `@import` rule still exists though, right? as long as we continue to provide an `oTable()` mixin even when we switch to `@use` internally, this should continue to work as before?

JakeChampion: > the `@import` rule still exists though, right? as long as we continue to provide an `oTable()` mixin even when we switch to `@use` internally, this should continue to work as before?

Yes, if we keep the current system as well then we won't have to do extra work in the build-service 👍

JakeChampion:
@Financial-Times/origami-core Is there anything which is stopping this from being merged as an accepted proposal?

JakeChampion:
I think this should be merged as we all agree on the proposal and have also gotten reviews from outside the origami team for it 👍 

What do you think @Financial-Times/origami-core ?

chee:
> I think this should be merged as we all agree on the proposal and have also gotten reviews from outside the origami team for it +1
> 
> What do you think @Financial-Times/origami-core ?

okay.

we'll want to rename the file prefix `0000-` to some positive number first.

i am slightly hesitant because i think it's possible we'd like to do even more than this is version 2, but hopefully that can all build on this work and not contradict it. happy to merge

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,157 @@
...
+
+- Our documentation assumes bower but some people are already using npm, which can lead to confusion and wasted time. By becoming npm only Origami documentation becomes more simple.
+
+## explanation
+
+We stop supporting bower!
+
+We delete `bower.json` from our components and we convert them to npm components, moving the dependencies into the `package.json`.
+
+We'll be bumping the `origamiVersion` from `1` to `"2.0.0"`, so our services and tools know to process the component as an npm component.
```
JakeChampion, line 33: I can't remember if we ended up deciding to go with semver for the spec or not.
@notlee, @chee do you happen to remember?

notlee: I thought we decided not to for v2  but thought we might in the near future, but don't remember either. 

FYI this question came up with regards to the Repo Data api:
https://financialtimes.slack.com/archives/CSW6B2VAN/p1607018153027600

notlee: https://github.com/Financial-Times/origami-repo-data/pull/268

chee: @JakeChampion we decided to use "2.0". not semver, just a major and a minor.

chee: https://github.com/Financial-Times/origami/pull/86#issuecomment-712775430

proposals/accepted/0000-npm.md
```diff
@@ -0,0 +1,157 @@
+# Proposal: npm-only origami components
+
+## motivation
+
+There is so much motivation:
```
chee, line 5:
```suggestion
There is _so much_ motivation:
```

proposals/accepted/0000-npm.md
chee, line 33:
```suggestion
We'll be bumping the `origamiVersion` from `1` to `2.0`, so our services and tools know to process the component as an npm component.

The addition of a decimal point will allow us to make and track backwards compatible changes to the spec going forward.
```

PR #87, opened 2020-10-09
fix superscript syntax

apaleslimghost:
(i guess `^` is an org mode thing? it's not supported in GFM)


chee:
lol i think it's a reddit thing, we were using codiMD

PR #90, opened 2020-11-05
Component demos & docs proposals

chee:
**EDIT**: new proposal, you may ignore the other two for now!

here is the new proposal: https://github.com/Financial-Times/origami/blob/component-demos/proposals/accepted/0000-demos-as-isolated-node-modules.md

i will be writing a separate docs proposal next.

---

_outdated pr description:_

Here are two proposals.

1. Component docs as a static site ([rendered markdown](https://github.com/Financial-Times/origami/blob/component-demos/proposals/accepted/0000-component-docs-as-static-site.md))
2. Component docs (convention over configuration) ([rendered markdown](https://github.com/Financial-Times/origami/blob/component-demos/proposals/accepted/0000-component-docs-convention-over-configuration.md))

i'm not sure if i like either of them, now that i've typed them up. i might just be exhausted, but wouldn't it be good if we could have docs like this: https://design-system.service.gov.uk/components/ or this: https://rizzo.lonelyplanet.com?

We can't because Origami, in its current form, exists to solve a technical problem. Not a design problem. It's a component system, not a design system. 

If we had all our components in a single repository, we would replace the registry with documentation. We'd need a separate page where we list our services (unless we put them in same repo too). What if we did that? And started treating Origami as a single project?

Everyone should be committing to one open source repository that Design and Origami are Admins of but that everyone in the company has write access to

We should have holistic documentation about how everything works together. About the wider design decisions that make up Origami, about accessibility

in its present form, with everything in separate repos, every component containing its own documentation etc etc, it is solving a technical problem.

it is a component system, not a design system, it's a way of building components so they can be distributed by our services

right now i can make a component that is lime green with pink text and it just says the word "HAMBURGER" and it flashes on and off `<blink>HAMBURGER, HAMBURGER</blink>`, and i can put an origami.json in the directory, and it will appear in the registry, and i can use the build service.

that doesn't help with design, it doesn't help anyone

### DIFF
--- a/proposals/accepted/0000-component-docs-as-static-site.md
+++ b/proposals/accepted/0000-component-docs-as-static-site.md
@@ -1,0 +1,57 @@
+# Origami component docs as a static site
+
+We remove the demos config from the manifest spec. Documentation, shown in in
+the registry, is any static site present in the `docs` directory after running
+`npm run build`.
+
+## motivation
+
+At the moment our component docs are split between the README.md (which people
+see on the github) and a series of demos defined in the `origami.json` file. The
+demo config is complex, hard to write and parse. It's both too strict, and too
+loose. It contains cruft (like `display_html`) that's built up over the years of
+use, and we can do a better job now.
+
+Neither the demos or the readme are a great way to write more holistically about
+the component, about accessibility and design decisions that went into the component.
+
+If we know docs will be built once `npm run build` is run, then we can remove
+logic around generating the docs from the registry.
+
+## explanation
+
+The spec states only this about docs:
+
+- the readme.md must link to registry/component/docs
+- component documentation, including demos of the component in many states with
+  accompanying prose, must exist as a static site in the `docs` directory after
+  `npm run build` has been issued
+- if your component supports brands, the build command should take into account
+  the `ORIGAMI_BRAND` environment variable and build accordingly
+- a live-reloading dev server must/should run when `npm start` is issued within
+  the component's repository
+
+## work required
+
+- remove the demo config from the v2 manifest spec
+- when migrating cmoponents to v2, create documentation for them
+
+## alternatives
+
+### do nothing
+
+keep our demos as they are. keep our docs spread across readme and demos.
+
+### convention over configuration
+
+A convention-over-configuration system for docs. A more strict system defining
+how docs should be formed. [See here](./0000-component-docs-convention-over-configuration.md)
+
+## notes
+
+- a previous version of this proposal suggested the docs should be built
+  in the published npm module. This is fine, but it makes it more difficult to
+  define how branding should work. Currently in the registry, branding is
+  selected via a query string. Instead of having them prebuilt i'm suggesting
+  that the component is built by the registry and we cache the result. This does
+  open us up to code execution exploits however. This probably needs more work.
--- a/proposals/accepted/0000-component-docs-convention-over-configuration.md
+++ b/proposals/accepted/0000-component-docs-convention-over-configuration.md
@@ -1,0 +1,82 @@
+# Origami component docs (convention over configuration)
+
+We remove the demos configuration from the manifest spec. Documentation is built
+the same way for every component from a strict and specified directory structure
+in a new `docs` directory.
+
+## motivation
+
+At the moment our component docs are split between the README.md (which people
+see on the github) and a series of demos defined in the `origami.json` file. The
+demo config is complex, hard to write and parse. It's both too strict, and too
+loose. It contains cruft (like `display_html`) that's built up over the years of
+use, and we can do a better job now.
+
+Neither the demos or the readme are a great way to write more holistically about
+the component, about accessibility and design decisions that went into the component.
+
+Having a single way for docs to be produced means anyone coming into any Origami
+component who has seen another, knows how to view and make demos. It also means
+less logic required to build the docs.
+
+## explanation
+
+The docs directory looks like this:
+
+```
+o-table/
+└── docs
+	├── basic-table
+	│   ├── index.html
+	│   ├── readme.md
+	│   ├── script.js
+	│   └── style.sass
+	├── monkey-table
+	│   ├── index.html
+	│   ├── readme.md
+	│   ├── script.js
+	│   └── style.sass
+	├── readme.md
+	└── sortable-table
+		├── index.html
+		├── readme.md
+		├── script.js
+		└── style.sass
+```
+
+The main readme.md file contains general documentation about the component (of
+the kind that used to be in the component's main README.md). Laced under that
+are is the content of each of the subdirectory's READMEs, with the demo they are
+documenting displayed in a codepen-like box with an o-tab for each of the
+`index.html`, `script.js`, `style.sass` and output.
+
+It's recursive, so if inside `monkey-table` you have more directories containing
+`index.html`, `script.js`, `style.sass` and `readme.md` then those will be
+weaved in under the `monkey-table` section.
+
+The navigation in the registry for that component can be generated for these
+components, and we can create a tool that generates the static site based on
+this directory structure.
+
+## work required
+
+- remove the demo configuration from the v2 manifest spec
+- add a specification for this demo structure
+- when migrating components to v2, create documentation in this directory structure
+- create a tool for building a static site from this directory structure
+
+## alternatives
+
+### do nothing
+
+keep our demos as they are. keep our docs spread across readme and demos.
+
+### static site
+
+Do not specify this format, but instead allow for any site that is built to the
+`docs` directory to be served. [See here](./0000-component-docs-as-static-site.md)
+
+## notes
+
+- each demo should possibly be an individual node module, and list the
+  dependencies in their own `package.json`
--- a/proposals/accepted/0000-demos-as-isolated-node-modules.md
+++ b/proposals/accepted/0000-demos-as-isolated-node-modules.md
@@ -168,10 +168,12 @@ demos still.
 
 ### other proposals
 
-These abandoned proposals:
+These abandoned proposals when demos and docs where conflated.
 
-https://github.com/Financial-Times/origami/blob/5b52d7ba831ccbfc17a7babb7b83c83bd8694d90/proposals/accepted/0000-component-docs-as-static-site.md
-https://github.com/Financial-Times/origami/blob/5b52d7ba831ccbfc17a7babb7b83c83bd8694d90/proposals/accepted/0000-component-docs-convention-over-configuration.md
+- [https://github.com/Financial-Times/origami/blob/5b52d7ba831ccbfc17a7babb7b83c83bd8694d90/proposals/accepted/0000-component-docs-as-static-site.md](component
+  docs as static site)
+- [https://github.com/Financial-Times/origami/blob/5b52d7ba831ccbfc17a7babb7b83c83bd8694d90/proposals/accepted/0000-component-docs-convention-over-configuration.md](component
+  docs convention over configuration)
 
 ## notes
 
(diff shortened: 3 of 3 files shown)


proposals/accepted/0000-component-docs-as-static-site.md
```diff
@@ -0,0 +1,57 @@
...
+
+## explanation
+
+The spec states only this about docs:
+
+- the readme.md must link to registry/component/docs
+- component documentation, including demos of the component in many states with
+  accompanying prose, must exist as a static site in the `docs` directory after
+  `npm run build` has been issued
+- if your component supports brands, the build command should take into account
```
JakeChampion, line 29: Does this mean that `ORIGAMI_BRAND=interal npm run build` would overwrite the files created by `ORIGAMI_BRAND=whitelabel npm run build` and vice versa?

proposals/accepted/0000-component-docs-convention-over-configuration.md
```diff
@@ -0,0 +1,82 @@
...
+	│   └── style.sass
+	├── readme.md
+	└── sortable-table
+		├── index.html
+		├── readme.md
+		├── script.js
+		└── style.sass
+```
+
+The main readme.md file contains general documentation about the component (of
```
JakeChampion, line 47: How would we work with brand specific docs/demos of our components?
Currently we have demos we are specific to certain brands and so they are not built for the other brands.

A real example is o-message which has very different demos for each brand.

o-message master brand:
image

o-message internal brand:
image


o-message whitelabel brand:
image

proposals/accepted/0000-component-docs-as-static-site.md
chee: yes, but maybe we can come up with something better than that.

to be honest i'm not a big fan of either of these proposals and kind of want to go back to an earlier idea where demos are isolated node modules and list their dependencies in the `package.json` and we build them in the registry and cache those builds.

hopefully we can all have a chat about it and come up with something better than what i've written here

proposals/accepted/0000-component-docs-convention-over-configuration.md
chee: yeah.

so i think that the problem that we keep running into here is that demos serve at least 2 purposes. One is demos for use when working on components: something to preview your changes, to run accessibility tests on etc. (there may be a third purpose, which is the kind of utilities you see on [o-colors](https://registry.origami.ft.com/components/o-colors@5.3.0#demo-contrast-ratio-checker)).

The other is as a kind of user-facing documentation.

While writing these proposals i kept momentarily forgetting about a given purpose.

Now that I can see they are (at least) two separate things, I think I should revisit this whole thing from that perspective.

I think we should make demos super easy to write. The way they are now doesn't have to change much, we probably don't need any long prose about them. I think having them as isolated node modules (in the component's repo) is good. And I also think that a convention over configuration style like the tree outlined in the proposal would be good for them too. but, for sure, you should be able to do something like `npm run create demo "Basic Table"`. and, for sure, you should be able to get a live-reloading demo server up and running really fast. And maybe we can use something off-the-shelf like storybook.

For these, brands should probably work something like they do now. perhaps an `origami: {brands:[]}` node in the package.json, or a demo.json. Or, if we choose to use something like [mdx](https://mdx.mdx.now.sh/) then perhaps we state in the demo's entrypoint which brands it is valid for.

As for the other purpose, user-facing documentation, I think we should separately tackle writing prose about each component in a `docs` directory. Basically we move the content that's currently in the `README.md` and we make what is currently codeblocks and screenshots into live html. And there is where we would write about design decisions and include utilities like the o-colors contrast ratio checker. And there we write about design decisions and accessibility choices. I'm not sure if this should be part of the registry or not, because it seems like it should have its own navigation. 

I'm still trying to form these ideas into something coherent.

chee: Okay, so based on this comment i'm going to rewrite these proposals. In fact, I'm going to delete them.

Instead I'll create one proposal about how demos should be isolated node modules in the component repo.
Because we won't be using `demos` for documentation and utilities, we don't need to support a lot of the stuff that's currently in the demo's manifest specification. 

here's how i see it:

| property        | purpose                               | replacement                           |
|-----------------|---------------------------------------|---------------------------------------|
| name            | output file name                      | n/a, path is path is on disk          |
| title           | title for the demo in the registry    | `<title/>` in template.html           |
| description     | demo description in registry          | `<p/>` in template.html               |
| template        | path to mustache                      | template.html                         |
| sass            | path to sass                          | style.scss                            |
| js              | path to js                            | script.js                             |
| data            | data to pass to the mustache template | n/a, hardcode values in demo          |
| brands          | supported brands                      | probably a field in package.json      |
| documentClasses | classes to set on html root el        | <html> in html file                   |
| dependencies    | dependencies                          | dependencies in package.json          |
| hidden          | hide demo in registry                 | n/a, show all demos                   |
| display_html    | hide demo's html tab in registry      | n/a, utilities should be elsewhere    |

the purpose of demos will be for development and for CI accessibility testing. they also serve as a usage example, and an integration test, because they are a real node module.

documentation we will handle separately. initially with a `docs/` folder in the component repo that will be rendered in the registry and linked to from the readme. Documentation will initially be formed from the current README content and any demos that previously had `display_html` set to true. Documentation will be the initial page for components when viewing them in the registry.

I'm interested in adding authentication to the docs page so designers can make propose edits to the pages, though hopefully they will also feel comfortable making changes directly to the markdown files in the repos.

Eventually docs may be a separate application that talks holistically about how our components fit together, but for now keeping them in the repo seems like the right path.

proposals/accepted/0000-demos-as-isolated-node-modules.md
```diff
@@ -0,0 +1,187 @@
+# Demos as isolated node modules
+
+We remove the demos config from the manifest spec. Demos are isolated node
```
notlee, line 3: Yes!

proposals/accepted/0000-demos-as-isolated-node-modules.md
```diff
@@ -0,0 +1,187 @@
...
+modules that list the containing component as a dependency in the
+`package.json`.
+
+Demos no longer serve as documentation, and are no longer used for utilities
```
notlee, line 7: I think some of those utilities are used and provide value. We should include a proposal to replace them as part of this. Or, if we break them out later, is there a downside to keeping them as demos for now?

proposals/accepted/0000-demos-as-isolated-node-modules.md
```diff
@@ -0,0 +1,187 @@
...
+
+## motivation
+
+Our demos currently serve 3 purposes:
+1. Demos -- for development and CI testing
+2. Utilities -- like the o-colors contrast checker
+3. User facing documentation -- they are displayed prominently in the registry
+
+The demos array in the origami.json is both too strict and too loose. It's
+difficult to parse, and difficult to lint. The `display_html` boolean is a way
```
notlee, line 19: We sometimes use `display_html` to demonstrate Sass features. We hide the html because Build Service users can't copy/paste. Its a way of shouting "no don't override styles look! look! custom themes. Check the readme". I think whilst the registry demo/readme is structured like it is we still need to be able to show these demos and hide the html to [avoid confusion](https://financialtimes.slack.com/archives/C02FU5ARJ/p1580316002079100). 

I'm totally onboard with removing `display_html` though, just uncertain on how to get there 🤔 Is the plan to create a partner documentation proposal? It feels ok to remove `display_html`, and keep the Sass demos which use it, if we have a plan on how we'll frame those demos differently

proposals/accepted/0000-demos-as-isolated-node-modules.md
```diff
@@ -0,0 +1,187 @@
...
+the component specification explaining the conventions.
+
+### 2. Component work
+
+Rewrite the current demos to work in this style. Move utility demos out of the
+current demos directory.
+
+### 3. Tooling
+
+Build a tool that can be run in the component's root that has the following
```
notlee, line 77: Looks good! How will this work with Build Tools? Would Build Tools commands defer to odc? Or as a maintainer would I use both? Kinda seems like an obt replacement minus `test`?

proposals/accepted/0000-demos-as-isolated-node-modules.md
```diff
@@ -0,0 +1,187 @@
...
+
+It starts up a filesystem watcher that builds the code on every change, and
+reloads the user's browser on every build.
+
+#### lint
+
+This lints the demos index.md files to ensure it will work with the rest of the
+tool.
+
+### create
```
notlee, line 123: Nice idea!

proposals/accepted/0000-demos-as-isolated-node-modules.md
```diff
@@ -0,0 +1,187 @@
...
+description and an html code block.
+
+We replace the demo properties in the origami.json as follows:
+
+| property        | purpose                               | replacement                           |
+|-----------------|---------------------------------------|---------------------------------------|
+| name            | output file name                      | n/a, path is path is on disk          |
+| title           | title for the demo in the registry    | top-level heading in index.md         |
+| description     | demo description in registry          | content between h1 and codeblock      |
+| template        | path to mustache                      | the html codeblock in index.md        |
```
notlee, line 51: I'm less certain about this. Many demos use data and share the same moustache template. When they don't it can be time consuming to update demos. Especially when working on new features or breaking changes. I've seen it cause quite a few markup issues which are then copy-pasted around projects. 

Could a separate mustache template work? Would it be rude of it to include a shared partial from outside the individual demo directory? This is sounding rubbish. I'm not sure how to solve this problem. Provide component templates? 🙈

notlee (commented):
I like this thinking a lot. :D There are a couple of bits I'm not sure about.

proposals/accepted/0000-demos-as-isolated-node-modules.md
chee: Oh, yes! Those utilities are great! There will be a joint proposal to go along with this for how documentation will work in the future, and the utilities will be moved there as part of that.

I'd really like for us to have a meeting with design where we discuss a system they'd be comfortable editing before we embark on that docs proposal, and until then I think we can keep them around as demos for sure.

proposals/accepted/0000-demos-as-isolated-node-modules.md
chee: Is it possible that we could write in the demo description something explaining this rather than hiding the html?

Yes, the plan is to have a separate proposal for documentation where we write things about how build service users vs. sass users should approach things.

proposals/accepted/0000-demos-as-isolated-node-modules.md
chee: This could be part of build tools, or a replacement for build tools!

I think you're right. I think that we would be either replacing build tools, or working this into it.

proposals/accepted/0000-demos-as-isolated-node-modules.md
chee: I find personally that when I've been updating demos that have shared templates that it is surprising. I'd expect each demo to relate to a single file. When wanting to change 1 demo and learning that it affects other demos that use the same template I've been surprised too.

This is objective, and we should talk to other people about it, but i think it's less confusing if demos are isolated from one another. @JakeChampion thoughts?

I would hope that with visual regression testing we would see that a breaking change hasn't been updated correctly in all demos.

JakeChampion: I prefer having isolated demos rather than a shared template for the same reasons that @chee has mentioned

JakeChampion:
Chee mentioned in this proposal the [idea of using StoryBook for Origami component demos](https://github.com/Financial-Times/origami/pull/90#discussion_r518720830). [Customer Products are looking at using the StoryBook composition feature.](https://github.com/Financial-Times/next-storybook-composition/pull/1). We should probably be setting aside some time to look into the feasibility of using StoryBook and aligning on this across our groups so that we can have a single location for all component demos (Origami, Customer Products, Internal Products).

chee:
i think we should investigate using storybook for this instead

Issue #26, opened 2020-03-02
Identify Issues/Blockers That Prevent New Origami Component Maintainers


notlee:
The Origami team build and maintain Origami components which are used
by multiple teams, sometimes across multiple brands/groups. But anyone 
can build an Origami component and have it displayed in the registry. E.g. 
the ads teams maintain o-ads, the labs team created o-crossword, the 
comments team maintain o-comments, the (previously known as) conversion 
team maintain o-subs-card.

If more teams built and maintained Origami components with the same tools
and standards it would help design consistency (as we could preview all our
components in the registry) and it would help save developer time (we learnt
how difficult and time consuming it can be to maintain projects with
different standards and tooling during the major cascade).
See related next team issue: [Making n-components Origami-spec compliant](https://github.com/Financial-Times/next/issues/378)

Let's identify issues/blockers people face building and maintaining Origami 
components, so we can come up with incremental solutions to encourage
more people to maintain their own Origami compatible components

notlee:
Some thoughts to investigate:

- Engineers may not know it's possible / encouraged.
- There's no tutorial or complete documentation how (finding the [obt init](https://github.com/Financial-Times/origami-build-tools#init) docs 
is an engineers best bet).
- It requires multiple [Origami specs](https://origami.ft.com/spec/v1/) to be in your head. Which might feel like an 
intimidating / lengthy read. It's easy to make a mistake without Origami team 
reviews.
- Templates. More projects are using JSX, but Origami components should not 
provide any templates, and may only provide mustache templates [according 
to the spec](https://origami.ft.com/spec/v1/markup/#templates). 
- Component composition issues. How do we compose components into larger 
shareable components/patterns? The Origami team don't have to do this often but 
other teams do. One way they approach this is to output the css for components
they depend on and provide a template. Origami components cannot output styles 
for dependencies, all selectors start with the component name, and don't do
templates.
- Ownership. When viewing an Origami component in the registry which is not
maintained by the Origami team there's a note that it's not supported
by the Origami team. But there's no way to see all the components a team own.

wheresrhys:
You've got most/all major ones there.

> It requires multiple Origami specs to be in your head.

Individual specs are also arguably too abstract (structured as normative rules, rather than a tutorial) to be a useful guide. (Although I am just flicking through them now and wow - massive improvement on earlier editions. Not sure when they were updated and split up, but good job!)

Some more ideas

[...]

PR #93, opened 2020-11-27
add proposal for renaming the team

JakeChampion:
Currently name for the team is the same name as the project and this can cause confusion about when talking about Origami.

Having the same name for both could also make people think that they shouldn't contribute to an Origami codebase.

### DIFF
--- a/proposals/accepted/0000-rename-the-team.md
+++ b/proposals/accepted/0000-rename-the-team.md
@@ -1,0 +1,34 @@
+# Rename the Origami team
+
+The name for the team is the same name as the project and this can cause confusion about when talking about either one.
+Having the same name for both could also make people think that they shouldn't contribute to an Origami codebase.
+
+## motivation
+
+We want everyone to feel they can contribute to Origami projects and we want to remove the confusion between the team and the projects when someone mentions Origami.
+
+## explanation
+
+Let's rename the team to something which shows that the team members are the stewards of the Origami project and that anyone can contribute to Origami.
+
+## work required
+
+- Come up with a name which makes it simpler to understand that someone is talking about either the Origami team or the Origami project
+- Update documentation and contact pages to use the new team name
+- Maybe create a new mailing list which uses the new team name rather than `origami.support`
+
+## alternatives
+
+Do nothing and keep the potential confusion when someone talks about Origami (the team) or Origami (the project).
+
+## supporting examples
+
+Vue and the [Vue Core team](https://vuejs.org/v2/guide/team.html)
+NodeJS and the NodeJS Technical Steering Committee?
+
+## counter examples
+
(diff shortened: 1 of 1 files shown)


proposals/accepted/0000-rename-the-team.md
```diff
@@ -0,0 +1,34 @@
+# Rename the Origami team
+
+The name for the team is the same name as the project and this can cause confusion about when talking about either one.
+Having the same name for both could also make people think that they shouldn't contribute to an Origami codebase.
+
+## motivation
+
+We want everyone to feel they can contribute to Origami projects and we want to remove the confusion between the team and the projects when someone mentions Origami.
```
nickcolley, line 8: Is there evidence of this confusion from the community?

In terms of contribution, there may be bigger blockers to contribution than the name of the team.

notlee: - Not that I have to hand. We use Origami interchangeably in documentation in places and have to be careful to clarify team/project.

- To me, this proposal is more about helping with the sense of ownership. But "there may be bigger blockers to contribution than the name of the team" is a [good point](https://github.com/Financial-Times/origami/issues/26).

chee: There are bigger things, but we're also doing those. Doing one thing does not mean we can no longer do other things.

chee: Yes, there are bigger things. But there is also this thing.
The name is part of a _series_ of changes we're making.

The evidence is that when you are talking about Origami and the Origami Team that you have to write "the Origami Team". People use "Origami" to refer to both the project and the team, but the team does more than just work on Origami  and Origami is worked on by more people than just this team.

It's partly about this ambiguity, and it's partly a significant act intended to declare (as part of a series of other actions) that Origami is a project owned by every designer and developer of the FT Group and the team are caretakers.

It's part of a series of changes that we are making along with improving documentation and outreach.

notlee:
I'd love a future where the Origami team includes designers as core members (again?) It would help us meet our [proposed mission/vision statements](https://github.com/Financial-Times/origami-website/issues/292). In such a future could the team be called the Design Operations team? We'd be Design Operations sitting under the Operations & Reliability group.

JakeChampion:
Things we do and what names the may have:
Origami Technical Steering Committee: A group of people who are in charge of updating and maintaining the specification.

Front-end infrastructure: A group of HTTP services which 

Design-Ops

JakeChampion:
Things we do and what names the may have:

- Design-Ops:

- Origami Technical Steering Committee: A group of people who are in charge of updating and maintaining the specification.

- Front-end infrastructure: A group of HTTP services which

Issue #42, opened 2020-04-02
New Component: Tag Input


rowanmanning:
## What

Tag Input is a working title, but lots of people have a similar interface for selecting multiple things, whether they're tags or topics or something else. It's really difficult to describe this component in words, but you'll know what I mean when you see it:

<img width="442" alt="Example of component from the Graphics Hub" src="https://user-images.githubusercontent.com/10405691/64521164-6846d500-d2ef-11e9-9e18-ba3d7cd12485.png">

The whole component is styled to look a bit like a text input or textarea. There's an _actual_ text input at the end (often an [autocomplete](https://github.com/Financial-Times/origami/issues/5)) which is styled to look like part of the background. When the `return` key is pressed after entering text, or an autocomplete option is selected, then a new visual representation of the text or item is added in front of the text input. Each item also has a button which removes it from the list.

## Why

These are used a lot around the FT, there's a definite user need. Often these are custom-built or use a third-party, the experience is inconsistent across FT products and we have no idea if they're accessible or resilient.

I think this is separate from an [autocomplete](https://github.com/Financial-Times/origami/issues/5), because an autocomplete can exist without this tag adding behaviour. By building a separate component we don't enforce tags being autocompleted, but you could add autocomplete functionality to the input if you need it.

## Supporting Examples

These are mostly taken from the separate autocomplete proposal.

### Graphics hub

I think we shouldn't implement the "suggestions" area below, but we _should_ offer a JavaScript API which would allow people to implement this. (e.g. `tagInput.add('thing')`)

<img width="442" alt="Screenshot 2019-05-31 at 11 47 35" src="https://user-images.githubusercontent.com/10405691/64521164-6846d500-d2ef-11e9-9e18-ba3d7cd12485.png">

###  BizOps

<img width="630" alt="Screenshot 2019-05-31 at 11 58 27" src="https://user-images.githubusercontent.com/10405691/64521158-641ab780-d2ef-11e9-8328-7433d140f0f7.png">

### Enterprise tools

We might need the character splitting to be configurable. For emails it makes sense for space to create a new tag, but not necessarily for other things. E.g. topics might be allowed to contain spaces.

<img width="621" alt="Screenshot 2019-05-31 at 11 45 32" src="https://user-images.githubusercontent.com/10405691/64521201-7eed2c00-d2ef-11e9-980b-3c321a11c0f7.png">

### Fotoware

Originally mentioned by @adgad in the autocomplete proposal:
[...]

PR #100, opened 2020-12-07
Autocomplete

origamiserviceuser:
From #5.

see [rendered proposal](https://github.com/Financial-Times/origami/blob/proposals/autocomplete/proposals/accepted/0000-autocomplete.md)

## comments from issue

---

**notlee** _on 2018-11-14T13:02:08_

Another example when creating a health check via biz-ops. Currently uses awesomecomplete.

screen shot 2018-11-14 at 12 59 33

---

---

**GeoffThorpeFT** _on 2019-02-22T11:21:47_

Its worth noting that an autocomplete could be connected to a static or dynamic dataset.
Static datasets (or those datasets which are rarely updated) could be connected and populated at build time (thus removing the need for a permanent connection back to the raw data at runtime). Whereas the dynamic sets would need to retain a connection to data which may only be accessible through authentication. 

---

---

**gvonkoss** _on 2019-04-12T07:35:28_

Here's an interesting UX reference of ['rules' for autocomplete](http://jeremymikkola.com/posts/2019_03_19_rules_for_autocomplete.html).

---

---

**notlee** _on 2019-04-26T10:20:40_

Looks like cloud enablement also uses an autocomplete like interface

Screenshot 2019-04-26 at 11 17 31

---

---

**notlee** _on 2019-04-26T10:21:30_

We just discussed in our open meeting and think there's definitely a strong case for this component, and it something we intend to add 🎉 

---

---

**notlee** _on 2019-09-09T09:50:39_

**Edit by @rowanmanning 2020-04-02:** The autocomplete input at the end of these tag lists are relevant, but the interface for adding multiple items to a list after selection now has a [different proposal](https://github.com/Financial-Times/origami/issues/42)

---

Another auto-complete like component but for creating tags (thanks to Jono for tracking these down a little while ago now):

TagMe:
Screenshot 2019-05-31 at 12 07 42

Graphics hub:
Screenshot 2019-05-31 at 11 47 35

Biz ops:
Screenshot 2019-05-31 at 11 58 27

Enterprise tools:
Screenshot 2019-05-31 at 11 45 32

Loadstar:
image

---

---

**adgad** _on 2020-02-07T14:47:03_

**Edit by @rowanmanning 2020-04-02:** The autocomplete input at the end of these tag lists are relevant, but the interface for adding multiple items to a list after selection now has a [different proposal](https://github.com/Financial-Times/origami/issues/42)

---

We have a use case for something similar in Spark. When search Fotoware for images, our users want to easily search for multiple combined search terms (like "Donald Trump" "Boris Johnson"), without having to use quotes or fancy syntax. In Fotoware, this is done by adding search terms as tag-like things by pressing enter. We've replicated something like that in Spark. In our case, it's not from a known set of terms, so there's no autocomplete required.

Screenshot 2020-02-07 at 14 20 11

---

---

**rowanmanning** _on 2020-04-02T11:17:17_

I've created a [new proposal](https://github.com/Financial-Times/origami/issues/42) for tag inputs. The interface for adding multiple tags _can_ use an autocomplete, but won't be built as _part_ of an autocomplete component.

---

---

**nickcolley** _on 2020-07-13T12:59:01_

I have worked on maintaining this project: https://github.com/alphagov/accessible-autocomplete

Autocomplete is very complicated to get right, my main recommendation would be to try to make sure the results are separate from the interface itself.

So you have an 'engine' like a search engine, which is responsible for taking the input data and outputting useful result data.
An example of this might be in a country picker, determining that 'GB' means Great Britain.

This 'engine' can be separated from the actual interface aspects of the component itself, and I think that way it allows you to reduce the scope of an already very complicated component.

To stretch the metaphor a bit more, it's like having a car without an engine :P

---

---

**JakeChampion** _on 2020-08-04T12:58:37_

`@remindme` to look at this before the next quarter, we need to schedule this bigger work as an OKR.

---


JakeChampion:
There's also the search box on ft.com which is an autocomplete

notlee:
Some helpful references:
- https://adamsilver.io/blog/building-an-accessible-autocomplete-control/
- https://a11ysupport.io/tech/html/datalist_element
- https://github.com/alphagov/accessible-autocomplete

nickcolley:
https://github.com/alphagov/accessible-autocomplete/blob/master/accessibility-criteria.md <- might be useful

JakeChampion:
We're going to do a spike and see if it is possible to use alphagov/accessible-autocomplete and wrap it into an origami component.

notlee:
See https://github.com/Financial-Times/o-autocomplete/pull/2

JakeChampion (approved):

Issue #91, opened 2020-11-11
Provide a way to express that a component is safe to be used with any brand.


wheresrhys:
## What
Provide a way to express that a component is safe to be used with any brand.

## Why

e.g. https://registry.origami.ft.com/components/o-toggle@2.1.2 says 

> o-toggle hasn't been branded yet, it only supports the master brand. 

despite it not containing any of its own styles and being compatible with any brand.

notlee:
Good shout, some kind of n/a option would be good in case we want to support more brands within components directly :) Maybe `all`? We'd need to update the spec, repo data, and the registry to handle that appropriately. In the meantime setting support for all three brands is the way to go

PR #103, opened 2020-12-07
Provide a way to express that a component is safe to be used with any brand.

origamiserviceuser:
From #91.

see [rendered proposal](https://github.com/Financial-Times/origami/blob/proposals/provide-a-way-to-express-that-a-component-is-safe-to-be-used-with-any-brand/proposals/accepted/0000-provide-a-way-to-express-that-a-component-is-safe-to-be-used-with-any-brand.md)

## comments from issue

---

**notlee** _on 2020-11-11T15:33:38_

Good shout, some kind of n/a option would be good in case we want to support more brands within components directly :) Maybe `all`? We'd need to update the spec, repo data, the registry, build tools and the percy action (to run tests against all brands) to handle that appropriately. In the meantime setting support for all three brands is the way to go

---

### DIFF
--- a/proposals/accepted/0000-provide-a-way-to-express-that-a-component-is-safe-to-be-used-with-any-brand.md
+++ b/proposals/accepted/0000-provide-a-way-to-express-that-a-component-is-safe-to-be-used-with-any-brand.md
@@ -1,0 +1,11 @@
+# Provide a way to express that a component is safe to be used with any brand.
+## What
+Provide a way to express that a component is safe to be used with any brand.
+
+## Why
+
+e.g. https://registry.origami.ft.com/components/o-toggle@2.1.2 says 
+
+> o-toggle hasn't been branded yet, it only supports the master brand. 
+
+despite it not containing any of its own styles and being compatible with any brand.


proposals/accepted/0000-provide-a-way-to-express-that-a-component-is-safe-to-be-used-with-any-brand.md
```diff
@@ -0,0 +1,11 @@
+# Provide a way to express that a component is safe to be used with any brand.
+## What
+Provide a way to express that a component is safe to be used with any brand.
```
JakeChampion, line 3: We support this already by adding all the brands to the components `origami.json` `brands` array.

Is this proposal about adding a different way to declare support for all brands?

notlee: Yes. An "any/all" brands option is are more accurate description for some components, especially those without styles. It could be helpful if we create new brands or allow teams to create new brands themselves in the future.

PR #120, opened 2021-03-31
Add proposal to remove origamiCategory from origami.json manifest

JakeChampion:
👋 When working on the proposal to have origami components only on npm, [chee suggested](https://financialtimes.slack.com/archives/CSW6B2VAN/p1602497687031500) that we could remove this field from the origami.json manifest as it not being used by anything. 

👉 [View the rendered proposal here](https://github.com/Financial-Times/origami/blob/origamiCategory/proposals/accepted/0000-delete-origamiCategory-manifest-field.md).

👇 leave comments and any more examples below

### DIFF
--- a/proposals/accepted/0000-delete-origamiCategory-manifest-field.md
+++ b/proposals/accepted/0000-delete-origamiCategory-manifest-field.md
@@ -1,0 +1,38 @@
+# Delete the origami.json manifest field `origamiCategory`
+
+The field is not being used by any tooling or services, let's remove it from the specification.
+
+## motivation
+
+The field is no longer used for anything, let's stop requiring that it exists because it is now a useless field.
+
+## explanation
+
+It used to be used to decide how to display content on the [Origami Registry](https://registry.origami.ft.com/) but that is no longer the case.
+
+
+## work required
+
+From what I can think of, the work required is:
+
+- Update the component specification to remove the field from [`origami.json`](https://github.com/Financial-Times/origami-website/blob/c70e43a0fa06c0629de4e576d31f31885ecce5be/_specification-v1/manifest.md#origamicategory)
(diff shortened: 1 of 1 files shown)


chee (approved):

notlee:
I think categorising components is helpful and should be restored in the registry. They'll be even more helpful in the future when we relax the spec to make it easier to create patterns containing multiple components (that's tricky right now outside of low level components because of the class name requirement). I'd be in favour of removing this until we discuss, commit to review the categories, and restore them in the registry but think maybe it's not worth the extra work and we should keep them around for now?

JakeChampion:
> I think categorising components is helpful and should be restored in the registry. They'll be even more helpful in the future when we relax the spec to make it easier to create patterns containing multiple components (that's tricky right now outside of low level components because of the class name requirement). I'd be in favour of removing this until we discuss, commit to review the categories, and restore them in the registry but think maybe it's not worth the extra work and we should keep them around for now?

I'm also in favour of removing the field, and if later we decide we want a similar field, we can add one :)

notlee:
@JakeChampion Sorry, I think I was not very clear in my indirectness. I'm in favour in principle but in practise I think removing the field and adding it back is pointless work, as I think we will add it back, and therefore should leave it

chee:
we should delete this field.

if we decide to revive it, it is not likely to use the same values.

https://github.com/Financial-Times/origami-website/issues/180

JakeChampion:
I am not sure we would add it back. I also don't think it would be much work to remove the field and then possibly add a similar field in the future.
I like the idea of removing the field because it frees us from the already defined categories and in the future if we want a similar field, we can define all the categories that we want/need rather than being potentially hindered with the current defined categories.

For reference, posted below are our current defined categories:
>*utilities*: Sass and JavaScript utilities that provide no markup, provide no classes and are used to encapsulate shared logic between components
>*primitives*: Base components that provide minimal markup and are used by other components
>*components*: Components built from primitives and utilities, which provide markup for a complete user interface
>*layouts*: Complex components that provide styles for the whole page

notlee:
Okay, I still think my preference would be to keep these categories and deprecate them if we add new categories in the future, but I'm happy to go with the majority here

Issue #4, opened 2018-09-28
Remove the ability to customise classes across components


gvonkoss:
## What

The ability to customise classes has not been consistent across our components. We want to create consistency, so are opening this discussion to figure out whether we should remove the customisation, or add it to all components.

Currently we are leaning toward removing the ability to customise.

## Why

- More readable SCSS and simpler JS.
- Mixins can be used to include features without needing to wrap mixins in classes.
- Encourages CSS reuse including a component once rather than multiple times with different class names.
- Increase reliability and maintainability

rowanbeentje:
What's the ability to customise classes - got a short example or the calls used by the components which do support it?  Is this like the `new Overlay('overlay', { class: 'o-overlay--my-modifier' })` which *adds* an extra class to enable more targeting, or something else?

gvonkoss:
🙈and 🤔  
we were referring to situations like: 
```
@include oMessage($class: my-message) 

//output
.my-message {
    property: value;
}
``` 
I think it would involve some work to break out modifiers and such, but we're trying to get a feel for the general consensus

dan-searle:
I'm in favour of removing this ability. Is there a particular use case it was introduced for?

notlee:
@dan-searle Not that's known to me. It might be useful for using a component to build a new component, but we can still provide mixins which include no classes, [oNormaliseVisuallyHidden](https://github.com/Financial-Times/o-normalise/blob/master/src/scss/_mixins.scss#L3) is a good example .

i-like-robots:
I'm in favour of removing this too because it seems obtuse and is little, if ever, used.

rowanmanning:
Had a quick chat with @wheresrhys on this, wondered whether he'd have some history. We came up with some potential reasons why you _might_ do this:

  - class name shortening
  - vanity
  - retrofitting origami to sites where the dev has limited ability to edit the html

Going to look through some old issues and commits to see if I can find the real reasons

notlee:
>retrofitting origami to sites where the dev has limited ability to edit the html
Good thoughts. That sounded like a compelling reason but actually doesn't _really_ apply here as usually we only allow modifying the block portion of the class, not element or modifier portions.

notlee:
For the reasons above, we're decided to start removing the ability to customise the base class of Origami components. Thanks for your input everyone 👍

notlee:
[...]

PR #124, opened 2021-05-18
Add a proposal to allow components to output styles of other components

notlee:
👉  [view rendered proposal](https://github.com/Financial-Times/origami/blob/allow-other-component-css-output/proposals/accepted/0000-allow-other-component-css-output.md)

### DIFF
--- a/proposals/accepted/0000-allow-other-component-css-output.md
+++ b/proposals/accepted/0000-allow-other-component-css-output.md
@@ -1,0 +1,87 @@
+# Allow A Component To Output The CSS Of Another
+
+Allow one component to output the CSS of another component. This is currently disallowed implicitly in the Origami component specification (v1) as it requires that a component outputs namespaced CSS. E.g. a component `o-foo` only outputs CSS classes which begin with `o-foo`, and therefore couldn't output the styles of a dependency with classes which begin with `o-bar`.
+
+https://origami.ft.com/spec/v1/components/sass/#css-selectors
+
+## motivation
+
+Allowing one component to output the styles of another will make it easier to author, document, and use components which depend on other complex components.
+
+## explanation
+
+This example shows component Sass for a component `o-no` outputting CSS from a dependent component `o-forms`, this is currently not allowed, which means a user has to include `o-forms` CSS in their project themselves to use `o-no`:
+```scss
+// o-no/main.scss
+@mixin oNo() {
+    @include oForms();
+    .o-no {
+        // o-no styles here
+    }
+}
+```
+```html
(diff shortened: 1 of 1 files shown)


notlee:
Oh! So a related proposal:
https://github.com/Financial-Times/origami/issues/4

The Sass of many components used to accept a parameter for the prefix to output custom CSS classes. However there were issues with that which a detailed in the proposal. Off the top of my head it included passing the prefix around Sass mixin/functions to generate selectors (hard to search for maintenance, often missed so only some classes would use the custom prefix); similar the often wasn't a JS counterpart leading to a mix of customised/not customised class names; bundle size increases.

notlee:
E.g. [o-table had a `$class` argument for Sass mixins](https://github.com/Financial-Times/o-forms/pull/207/files#diff-0b65e46a83af62e9ad2a05a31d06b7ebb0aa534cdce4efde7472fbca30212403L166) which sometimes was missed but would also [add o-forms classes with JS, which couldn't be customised](https://github.com/Financial-Times/o-forms/blob/b0ae6d9b18c18474885d4c544bcab12068d387eb/src/js/forms.js#L92).

proposals/accepted/0000-allow-other-component-css-output.md
```diff
@@ -0,0 +1,87 @@
...
+
+https://origami.ft.com/spec/v1/components/sass/#css-selectors
+
+## motivation
+
+Allowing one component to output the styles of another will make it easier to author, document, and use components which depend on other complex components.
+
+## explanation
+
+This example shows component Sass for a component `o-no` outputting CSS from a dependent component `o-forms`, this is currently not allowed, which means a user has to include `o-forms` CSS in their project themselves to use `o-no`:
```
notlee, line 13:
```suggestion
This example shows component Sass for a component `o-no` outputting CSS from a component dependency `o-forms`, this is currently not allowed, which means a user has to include `o-forms` CSS in their project themselves to use `o-no`:
```

notlee:
We effectively accepted this with our work on an updated version of the component specification (v2).
https://github.com/Financial-Times/origami-website/pull/359

> A component **must** only style:
>	- The component’s root element and its children
>	- Elements with classes prefixed with the component’s name, and their
>	  children
>	- Elements with any other data-attribute in the component’s namespace, and
>	  their children

We decided a component outputting the styles of another component is allowed, as the outputting component isn't defining the styles.

However we then deprecated the specification entirely, so I'm going to close this issue as there is no immediate action to take. A component could choose to output styles from another component in the future, provided it's documented, without breaking a specification.

PR #146, opened 2021-09-15
Convert this repo to a component system monorepo

chee:


actions/percy/README.md
```diff
@@ -0,0 +1,51 @@
...
+    paths:
+     - "package.json"
+     - "package-lock.json"
+     - "origami.json"
+     - "main.js"
+     - "main.scss"
+     - "demos/**"
+     - "src/**"
+  push:
+    branches: master
```
JakeChampion, line 31: nit: this should say `main` now, but also, this file is auto-generated so maybe we don't need the sample config here :D

scripts/test.js
```diff
@@ -0,0 +1,42 @@
...
+
+	if (testFilter == "default" || !testFilter) {
+		if (await hasScript(workspace, "lint")) {
+			await $`npm run -w ${workspace} lint`
+		} else {
+			await $`npm exec -w ${workspace} -- origami-build-tools verify`
+		}
+
+		if (await hasScript(workspace, "build")) {
+			// await $`npm run -w ${workspace} build`
```
JakeChampion, line 32: note: commented out code here, this is because the build-script currently is used for heroku review-apps but we dont have those for his repo.

templates/labeler.yml
```diff
@@ -0,0 +1,25 @@
...
+  - components/**/*
+
+action:
+  - actions/**/*
+
+library:
+  - libraries/**/*
+
+hook:
+  - hook/**/*
```
JakeChampion, line 11: I think this should be `hooks/**/*`

templates/labeler.yml
```diff
@@ -0,0 +1,25 @@
...
+
+script:
+  - scripts/**/*
+
+tool:
+  - tools/**/*
+
+template:
+  - template/**/*
```
JakeChampion, line 20: I think this should be `templates/**/*`

PR #198, opened 2021-09-16
improve performance and logic around how to highlight headings in the sidebar

JakeChampion:
This changes the implementation from using a scroll event listener to using a series of intersection observers, which run off the main thread and are built to do this work in a performant manner.

I also took this time as an opportunity to improve the logic used to determine which heading to highlight.

The logic being used now is:
- Figure out what headings are above the intersection target and what ones are below it and place them in two arrays `above` and `below`.
- Figure out the scrolling direction
- If scrolling down the page, pick the last heading which is in the `above` array
- Else pick the first heading in the `below` array

When the page is scrolled to the bottom, then the last heading is highlighted, always. This is different from the previous logic which for certain document would never highlight the last heading.


components/o-layout/origami.json
```diff
@@ -6,6 +6,15 @@
 		"email": "origami.support@ft.com",
 		"slack": "financialtimes/origami-support"
 	},
+	"browserFeatures": {
+		"required": [
+			"Array.from",
+			"Array.prototype.includes",
+			"IntersectionObserver",
```
JakeChampion, line 13: Lee said: If we make the feature only run if IntersectionObserver exists then this would not be a breaking change

components/o-layout/src/js/layout.js
```diff
@@ -110,53 +110,167 @@ class Layout {
...
+					}
 				});
-			});
-		}.bind(this), 300));
+			}
+		}, {
+			threshold: 1, // Trigger only when whole element was visible
+		});
+
+		observerbottom.observe(this.layoutEl.querySelector('.o-layout__footer') || this.layoutEl.querySelector('.o-layout__main').lastElementChild);
```
JakeChampion, line 261: Todo: Extract this into a variable

components/o-layout/src/js/layout.js
```diff
@@ -110,53 +110,167 @@ class Layout {
...
+					if (headingIndexToHighlight === index) {
+						anchor.setAttribute('aria-current', 'location');
+					} else {
+						anchor.setAttribute('aria-current', 'false');
+					}
+				});
+				this.highlightedHeadingIndex = headingIndexToHighlight;
+			}, {
+				rootMargin: `-${
+					window.getComputedStyle(document.querySelector('h2')).fontSize
```
JakeChampion, line 236: Todo: Something better here such as getting the header from within the o-layout__main element instead of the first h2 in the document

components/o-layout/origami.json
JakeChampion: Done in [`120fad2` (#198)](https://github.com/Financial-Times/origami/pull/198/commits/120fad2fa4685f28d1acbf1efa5dd8558e6641d6)

components/o-layout/src/js/layout.js
```diff
@@ -110,53 +110,176 @@ class Layout {
...
+	 * Add aria-current to the correspoding link in the navigation for the header that was clicked
+	 * @private
+	 * @returns {void}
+	 */
+	setupClickHandlersForHeadings() {
+		this.navHeadings.forEach((anchor, index) => {
 			anchor.addEventListener('click', () => {
-				navAnchors.forEach(anchor => anchor.setAttribute('aria-current', false));
-				anchor.setAttribute('aria-current', 'location');
+				for (const sidebarAnchor of this.navAnchors) {
```
notlee, line 146: `sidebarAnchor` is a misleading name here, `headingAnchor`?

notlee (approved):
Looks good. One tiny non-blocking comment :)

chee:
very exciting, very exciting !!

PR #75, opened 2020-08-13
Introduce variable fonts to use more font variants

notlee:
preview: https://github.com/Financial-Times/origami/blob/9458290cd7ef01d81c76f48ac5e03a9213f9f4c4/proposals/accepted/0000-variable-fonts.md

PR #227, opened 2021-09-16
feat!: Output FinancierDisplayWeb Variable Font

JakeChampion:
If any FinancierDisplayWeb font variant is included also output
the equivalent variable FinancierDisplayWeb font once. The variants
of FinancierDisplayWeb specifically requested may be used as
fallback fonts for older browsers, whilst the variable font may be
used to display a greater number of font variants.

`oFontsVariantExists` does not yet account for variable fonts,
therefore they are effectively not usable since `o-typography`,
which checks a font variant exists before outputting typographic
styles, will continue to only allow styles supported by
non-variable fonts. `oFontsVariantExists` should perhaps be
deprecated in favour of `oFontsVariantAllowed`, that includes a
set of variants which may be used (`FinancierDisplayWeb`
supports 500 weights, we'd rather users include a limited set for
design consistency). That's blocked until the design team decide
what variants to allow.

Depends on: https://github.com/Financial-Times/o-fonts-assets/pull/53
Relates to: https://github.com/Financial-Times/origami/pull/75

### DIFF
--- a/components/o-fonts/src/scss/_functions.scss
+++ b/components/o-fonts/src/scss/_functions.scss
@@ -5,16 +5,16 @@
 /// @param {String} $weight [regular] - The font weight.
 /// @param {String} $style [normal] - The font style.
 /// @return {Bool}
-/// @see oFontsDefineCustomFont To register a custom font's avalible variants.
+/// @see oFontsDefineCustomFont To register a custom font's available variants.
 @function oFontsVariantExists($family, $weight, $style) {
 	$weight: if($weight, $weight, 'regular'); // Handles a falsy `$weight` value.
 	$style: if($style, $style, 'normal'); // Handles a falsy `$style` value.
-	$font-properties: map-get($_o-fonts-families, $family);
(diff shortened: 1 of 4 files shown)


JakeChampion:
This is a recreation of https://github.com/Financial-Times/o-fonts/pull/153

JakeChampion:
We can do the o-typography work inside this pull-request :D
https://github.com/Financial-Times/o-typography/pull/292

notlee:
Work on this has stalled. I didn't hear back from Klim/Kevin about a font weight mapping issue that will make it harder to roll out Metric2 to Origami users:



As one directly replaces another I was expecting the weight mappings to be the same, regardless of the name.

Metric 1         Metric 2
600-Semibold =   600-Bold
700-Bold     =   700-Heavy

notlee:
No up date on the above. It's unlikely we'll reuse this code so closing the PR. We may choose to role variable fonts out along with new foundational components with improved multi-brand support.

PR #291, opened 2021-09-20
Add homepage, bugs and support email to the package.json

### LINKED ISSUE #234: Update the `package.json` and `origami.json` to point at this repo
At the moment the manifests point people towards the old repos for support and issues and source code.

We should release a patch to every component with their URLs set to this repo instead

chee:
I've done these as `fix` so it causes a release, but we could do it as `docs` and just wait for the component to get its next release.

i wanted a release so they show up in npmjs with the correct urls :3

you'll wanna hide whitespace changs https://github.com/Financial-Times/origami/pull/291/files?w=1

real yakuza close #234


chee:
why did percy rerun and fail on all these?

chee:
I think it might be because the baseline branch got changed from `master` to `main`, so there are no baseline snapshots.

I've approved all those, because there is no way this PR could change how the components look. So merging this will also get us a new baseline for all our components.

chee:
don't worry about that failed labeler, it's having an unhappy about being asked to add 50+ labels to something that has 50+ labels (even though they're the same labels, it counts that as over 100 )

components/o-audio/origami.json
```diff
@@ -1,7 +1,7 @@
 {
 	"origamiType": "component",
 	"origamiVersion": "2.0",
-	"support": "https://github.com/Financial-Times/o-audio/issues",
+	"support": "https://github.com/Financial-Times/issues/new?labels=o-audio,components",
```
chee, line 4: woops, this should say "/origami/issues" lol!!!

JakeChampion (approved):
The packages in the libraries folder are applying a `components` label, not sure if that was intentional or if it was meant to be `libraries`

This is a great improvement 🙌

components/o-audio/origami.json
JakeChampion: I'm glad you saw that, I totally missed it

chee: i ran `npm bugs` in o-grid to test it and it opened a 404 :see_no_evil:

chee: if we decide we'd rather use `/issues?labels=` instead of `/issues/new?labels=` then we can do that as a chore some other time

chee:
> The packages in the libraries folder are applying a `components` label, not sure if that was intentional or if it was meant to be `libraries`
> 
> This is a great improvement raised_hands

nope! this is a mistake ;-;

Issue #293, opened 2021-09-21
Don't run percy if there have been no user facing changes


chee:
perhaps?? if the commits are all `docs:` and `chore:` then perhaps there should be no percy run

JakeChampion:
This is a great idea, this should save us a good chunk of screenshots :D

JakeChampion:
PR for this is https://github.com/Financial-Times/origami/pull/301

PR #301, opened 2021-09-22
Only run percy on the main branch or when a pull_request has a fix or feat commit and a file has changed within a component

JakeChampion:
This pull-request changes our percy github action and percy workflows.

The main change to the workflows is they will always run, for every component. This change means we can add all the percy workflows to our "branch protections", which will stop a pull-request from being able to merge until all the percy workflows have completed. Another change was also made, which changes the way the workflows run our action, switching from the action syntax of `uses: ./actions/percy/` to `- run: node ./actions/percy/src/index.js`, this let's us use whatever version of NodeJS we want - instead of being stuck on Node12.

The main change made to the github action is to only generate Percy snapshots if a file within the component has changed and a commit exists in the pull-request which is either a `fix` or a `feat`. The action also writes to the output what decision it took and why. E.G. It did not run generate the snapshots because no file in the component was changed.

Closes: https://github.com/Financial-Times/origami/issues/293


.github/workflows/percy-components-n-notification.yml
```diff
@@ -9,39 +13,22 @@ on:
...
       - reopened
       - synchronize
-    paths:
-     - "components/n-notification/package.json"
-     - "components/n-notification/package-lock.json"
-     - "components/n-notification/origami.json"
-     - "components/n-notification/main.js"
-     - "components/n-notification/main.scss"
-     - "components/n-notification/demos/**"
-     - "components/n-notification/src/**"
```
chee, line 19: oh, great! so _we'll_ decide when it runs for a workspace. cool!

.github/workflows/percy-components-n-notification.yml
```diff
@@ -9,39 +13,22 @@ on:
...
     - uses: actions/setup-node@v2
       with:
         cache: 'npm'
         node-version: 16
     - run: npm ci
-    - uses: ./actions/percy/
-      with:
-        # To comment as the Origami Fox mascot GitHub user
-        github-token: ${{secrets.ORIGAMI_FOX_ACCESS_TOKEN}}
+    - run: node ./actions/percy/src/index.js
```
chee, line 31: ? why is this not using it as an action anymore? this is now just a script?

JakeChampion: I wanted to use `zx` but that doesn't work on node12 and actions are stuck on node 12 and I couldn't see a date for when new node versions would be added. If we use it as a script then we decide what node version to use

JakeChampion: I can change it back to action format and not use `zx`, `zx` isn't a deal breaker for me

chee: i'm not against either approach, but i don't think it belongs in `actions` if it isn't being used as an action

what are the benefits of an action over a script in a monorepo?

components/o-big-number/src/scss/_mixins.scss
```diff
@@ -2,10 +2,12 @@
...
 	@include oTypographySans($scale: 8, $weight: 'semibold');
 	margin: 0 0 oSpacingByName('s1');
+	color: royalblue;
 }
 
 /// Styles for the copy under the big number.
 @mixin oBigNumberContent {
 	@include oTypographySans($scale: 1, $line-height: 22px);
 	max-width: oTypographyMaxLineWidth();
+	color: royalblue;
```
JakeChampion, line 12: This is to test the action and workflow changes. I chose o-big-number because it has one demo and won't generate hundreds of screenshots (o-colors generate 300+ screenshots!)

jsconfig.json
```diff
@@ -6,6 +6,7 @@
 		"moduleResolution": "node"
 	},
 	"include": [
+		"actions/**/*",
```
JakeChampion, line 9: This was added due to changing the action to make use of top-level await

.github/workflows/percy-components-n-notification.yml
JakeChampion: As discussed on Slack, we decided that there aren't good benefits to an action compared to a script and have moved this to the scripts folder and remove the actions workspace from the repo

.github/workflows/percy-components-o-colors.yml
```diff
@@ -1,6 +1,10 @@
 name: percy components/o-colors
 concurrency:
-  group: ${{ github.head_ref }}-components/o-colors
+  # We use github.ref instead of github.head_ref because github.head_ref
+  # only exists for pull-requests and this workflow is also running on
+  # pushes to the main branch.
+  # The value of github.ref is `refs/heads/<branch_name>`
+  group: ${{ github.ref }}-components/o-colors
```
chee, line 7: :eyes:

chee (approved):

chee:
@JakeChampion I did some rebasing and force pushing here because of conflicts from the main branch. make sure it's still all like you expect before you merge

chee (approved):

JakeChampion (commented):
All looks good!

PR #303, opened 2021-09-22
chore: revert to checking for changes and printing an error

chee:
we do want this functionality sometime, but this implementation breaks badly after a merge


JakeChampion (approved):
🙏

Issue #235, opened 2021-09-16
Add branch previews


chee:
Something we don't yet have that we had before is branch previews

chee:
I really like that idea where it does a real component prerelease on our private npm registry and we get to see it in action on the private origami registry

PR #305, opened 2021-09-22
chore: explicitly configure the npmjs used when publishing

JakeChampion:
I read https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages and I think this might need to be set in order for the .npmrc file to then be able to run npm publish

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -19,6 +19,7 @@ jobs:
         with:
           cache: 'npm'
           node-version: 16
+          registry-url: 'https://registry.npmjs.org'
         if: ${{ steps.release.outputs.releases_created }}
       - run: npm ci
         if: ${{ steps.release.outputs.releases_created }}


.github/workflows/release-please.yml
```diff
@@ -19,6 +19,7 @@ jobs:
         with:
           cache: 'npm'
           node-version: 16
+          registry-url: 'https://registry.npmjs.org'
```
JakeChampion, line 22: When we come to support branch previews #235 -- we can have a separate publish workflow which sets the npmjs registry to our staging one

JakeChampion:
I'm hoping this will make these workflows runs start to work -- https://github.com/Financial-Times/origami/runs/3677830865?check_suite_focus=true

JakeChampion:
Yes, I really do think this is what we needed to set:
https://github.com/actions/setup-node/blob/041bafb67276a76a9cc88cd8a4e99165e9eb287d/action.yml#L15-L16

chee (approved):
so strange!

PR #306, opened 2021-09-23
chore: remove unused .github folders within components which were left over from the migration to a monorepo

JakeChampion:
Question: should the root .gitignore contain entries to ignore `.github` folders in each workspace?


chee:
> Question: should the root .gitignore contain entries to ignore `.github` folders in each workspace?

i don't think we need to do this, because they'll stop happening when we stop importing projects, lol. and if we're using `git subtree` to import them, they'll already be committed (the gitignore will not take effect because it won't exist when those old commits are brought in)

chee (approved):

PR #307, opened 2021-09-23
fix all the readmes

chee:
ho! warnings


chee:
hark the herald angel sing no readme linting errors

JakeChampion (approved):

PR #317, opened 2021-09-27
feat: add script which logs out what releases will be made for which packages

JakeChampion:
The script checks all the commits between the base commit ref of a pull request and the head commit ref of a pull request. If the script is not running on a pull request then the base ref is `main` and the head ref is `HEAD` -- this allows the script to work for local development as well.

Once the commits are retrieved, the script then checks each commit to see if it is a 'publishable commit' -- I.E. a `fix` or `feat`.
If a commit is 'publishable' then the script looks at the files which changed in the commit and figures out if any of the files are within a 'publishable package' -- I.E. has an entry in `release-please-config.json`.packages.

The script also works out what version bump will occur for a given package.

An example of it in used:
If commits were made to `components/o-layout` which were `fix`, `fix!`, `chore`, `fix` - then the version bump will be `major` due to the breaking fix.


chee:
This is great!

It seems like it could be useful to pull out the part of this that finds packages that have had changes since the base branch (perhaps returning an object like `{[package_name]: [patch_minor_or_major]}`), and make it reusable. Then to make this script call that library function, and print out the list.

chee:
> This is great!
> 
> It seems like it could be useful to pull out the part of this that finds packages that have had changes since the base branch (perhaps returning an object like `{[package_name]: [nothing_or_patch_or_minor_or_major]}`), and make it reusable. Then to make this script call that library function, and print out the list.

could we then also use this same library function in the percy script?

JakeChampion:
> This is great!
> 
> It seems like it could be useful to pull out the part of this that finds packages that have had changes since the base branch (perhaps returning an object like `{[package_name]: [nothing_or_patch_or_minor_or_major]}`), and make it reusable. Then to make this script call that library function, and print out the list.

I've given this a go 👍

package.json
```diff
@@ -54,9 +54,9 @@
 	},
 	"dependencies": {
 		"@actions/core": "^1.2.6",
-    "@actions/exec": "^1.0.4",
-    "@actions/github": "^4.0.0",
-    "@actions/io": "^1.0.2",
+		"@actions/exec": "^1.0.4",
+		"@actions/github": "^4.0.0",
+		"@actions/io": "^1.0.2",
```
chee, line 59: what happened here

scripts/check-next-version.js
```diff
@@ -0,0 +1,32 @@
...
+    } else {
+        core.notice('No commits since last release are a `fix` or `feat` - which means no user facing changes have been made. This means no releases will be made.')
+    }
+} catch (error) {
+	core.setFailed(error.message)
+}
+
+function getBaseRef() {
+    let baseRef = 'main'
+    if (context?.payload?.pull_request?.base?.ref) {
```
chee, line 21: ?.nice?.

scripts/lib/getChangedPackages.js
```diff
@@ -0,0 +1,141 @@
...
+            userFacingChangedPackages.set(name, bump)
+        }
+    }
+    return userFacingChangedPackages
+}
+
+async function getPublishablePackages() {
+    let releasePleaseFile
+    try {
+        releasePleaseFile = await fs.readFile(path.join(__dirname, '../../release-please-config.json'), "utf-8")
```
chee, line 72: the rest of the scripts assume they are being run from the root of the repo, which is something i've personally been fine with. do you think we should update them to do this instead? perhaps we should make a `dirname()` utility function if so

JakeChampion: I'll change it to assume running from root 👍

scripts/check-next-version.js
JakeChampion: I'm still not used to this syntax

package.json
JakeChampion: In the pull-request where we changed percy from an action to a script, the dependencies were added to the root package.json manually and with space based indentation.

In this pull-request I ran `npm i splitonce` and npm fixed the indentation

scripts/lib/getChangedPackages.js
```diff
@@ -0,0 +1,141 @@
...
+ * 
+ * @param {string} baseRef The commit sha to use as the base
+ * @param {string} headRef The commit sha to use as the head
+ * @returns {Promise<Map<string, string|null>>} A promise which resolves to a map of package names to version bump type or null if no version bump required E.G. 'components/o-quote' => 'major'
+ */
+export async function getChangedPackages(baseRef, headRef) {
+    const publishablePackages = await getPublishablePackages()
+    // Map of package names to version bump type
+    // E.G. 'components/o-quote' => 'major'
+    const changedPackages = new Map
```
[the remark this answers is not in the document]
JakeChampion: No particular reason, I can switch to an object

scripts/check-next-version.js
```diff
@@ -0,0 +1,32 @@
...
+import {getChangedPackages, filterUserFacing} from "./lib/getChangedPackages.js"
+import {context} from "@actions/github"
+
+try {
+    let baseRef = getBaseRef()
+    let headRef = getHeadRef()
+    const packagesToPublish = filterUserFacing(await getChangedPackages(baseRef, headRef))
+    if (packagesToPublish.size) {
+        core.notice('A commit is a `fix` or `feat` commit, a release will be made.')
+        console.log({packagesToPublish})
```
JakeChampion, line 11: In some future work, we could made a workflow which updates a comment on a pull-request to let you know what packages will be released

chee:
what's going on with the `package-lock.json`? it seems to have removed all the workspaces

chee (approved):

PR #333, opened 2021-09-30
Add initital storybook config for o-buttons

chee:
image

## storybook
wow, i worked late but this is some of the most fun i've had. recommend checking this branch out and running `npm run storybook` and enjoying it.

storybook's docs plugin understands if you define a typescript interface for the props, and it automatically turns those into controls for the component. wild. so cool.

there isn't any mdx in this PR, but I did try that out locally and it works well. We can write prose about buttons and show examples.

## templates

We're going to make some templates, but we're not going to think about how Origami's javascript works with something like React.

We're not going to use any framework at all in the jsx.

Our templates are only for server-side rendering.


chee:
default to `secondary` rather than `primary` if no `type` is selected

chee:
ignore the templates in .npmignre

chee:
If you used this PR locally before `storybook` was moved to a subdirectory, you may have to run storybook once with the argument `--no-manager-cache`, so it doesn't get confused by the old cache

chee:
Update origami-build-tools to work when `src/tsx` is ignored, or move the tsx out of `src`

apps/storybook/config/preview.js
```diff
@@ -0,0 +1,9 @@
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
+  controls: {
+    matchers: {
+      color: /(background|color)$/i,
+      date: /Date$/,
+    },
+  },
+}
```
chee, line 9:
```suggestion
}

```

tools/origami-build-tools/test/integration/verify/verify.test.js
```diff
@@ -26,7 +26,7 @@ describe('obt verify', function () {
 			.then(() => process.chdir(process.cwd()));
 	});
 
-	describe('npmignore', function () {
+	describe.only('npmignore', function () {
```
chee, line 29:
```suggestion
	describe('npmignore', function () {
```

apps/storybook/readme.md
```diff
@@ -0,0 +1,11 @@
+# Origami storybook
+
+This app sources its stories from the `stories` directory inside each
+component.
+
+Stories that refer to a number of components, or are about wider guidance, can
```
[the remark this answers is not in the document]
notlee: o-table with a filter would be good to display in a "patterns" section of some kind rather than directly under o-table: demo shows o-table functionality with o-table + o-forms styles
https://registry.origami.ft.com/components/o-table@9.0.3#demo-filter-with-a-text-input

tools/origami-build-tools/lib/tasks/verify-dot-npmignore.js
```diff
@@ -20,7 +20,7 @@ async function dotNpmignore(config) {
 
 		let lineNumber = 0;
 		for await (const line of reader) {
-			const match = line.trim().match(/(main\.scss|main\.js|src\/?|demos\/?$|demos\/(?!local))/);
+			const match = line.trim().match(/(main\.scss|main\.js|src(\/js|\/scss|\/?$)|demos\/?$|demos\/(?!local))/);
```
notlee, line 23: 1 repo to rule them all

notlee (approved):
Approve with 1 caveat: do we really want to delete `.github/workflows/sync-repo-labels.yml`? Don't we still want to sync common labels like current/next across all our repos?

Tremendously excellent work on this :D

apps/storybook/readme.md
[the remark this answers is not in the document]
chee: nice, that will be fun to do. we can make that an MDX page so we can talk about what is from o-forms and what is from o-table

chee:
> do we really want to delete `.github/workflows/sync-repo-labels.yml`? Don't we still want to sync common labels like current/next across all our repos?

In the original discussion about this repo we decided not to have github label sync here, and let this repo manage its own labels

but then it was brought in by accident when the other actions were being added

PR #334, opened 2021-10-01
chore: Add content to fixture between h1 and h2

chee:
This will hopefully mean that the headless browser never selects the h2


chee:
@Financial-Times/origami-core this seems to be working!

PR #336, opened 2021-10-04
Add website

chee:
this repo's `origami.ft.com` branch is served as https://financial-times.github.io/origami
the storybook will be served at https://financial-times.github.io/origami/storybook

some issues remain:

- the urls are wrong (styles break), but will be right if we serve this at a root rather than a subdir (it's at `financial-times.github.io/origami`). we should get it a nice url to live at until we switch over from the website repo. i don't know how DNS works at the FT, can we get a `origami-next.ft.com` or something pointed at github?
-


.github/workflows/build-websites.yml
```diff
@@ -0,0 +1,28 @@
+name: build the websites
+on:
+  pull_request:
+    branches: [ main ]
+  push:
+    branches: [ main ]
+
```
chee, line 7:
```suggestion
  push:
    branches: [ main ]

```

scripts/build-website.bash
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+rm -r origami.ft.com/
+cd apps/website/ && bundler && make build
+mv _site ../../origami.ft.com
```
notlee, line 4: 👀

notlee (approved):
Hello what is this!
https://github.com/Financial-Times/origami-website/commit/172f02b0195011ce0d946ba8ea9cb4cd2b646915

Exciting times!

PR #339, opened 2021-10-05
add a small amount of documentation to the README

chee:
I think we should also add a readme to all of the top-level directories
so we can give people a little more info when they browse in.

for instance: document the scripts in scripts/readme.md


notlee (approved):
✨

PR #340, opened 2021-10-05
add a specific time for the demo date that prints the whole date

chee:
So that percy doesn't show a different date every time

I'll change this to a `docs: ` commit and do a force push after percy has run. This doesn't need a release.

### DIFF
--- a/components/o-date/demos/src/declarative.js
+++ b/components/o-date/demos/src/declarative.js
@@ -9,4 +9,8 @@ for (let i = 0; i < dates.length; i++) {
 	dates[i].setAttribute('datetime', today.toISOString());
 }
 
+// this one is being set to a specific time because otherwise
+// the demo changes on every build
+dates[dates.length - 1].setAttribute("datetime", "1912-04-15T05:18Z");
+
 ODate.init();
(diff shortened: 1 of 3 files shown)


chee:
this works great :)

components/o-date/demos/src/declarative.js
```diff
@@ -1,12 +1,16 @@
...
-const today = new Date();
+const sixHoursAgo = new Date();
+sixHoursAgo.setHours(now.getHours() - 6);
 const dates = document.querySelectorAll('time:not([datetime])');
-today.setHours(now.getHours() - 6);
 
 for (let i = 0; i < dates.length; i++) {
-	dates[i].setAttribute('datetime', today.toISOString());
+	// the last 2 are being set to a specific time because otherwise the demo
+	// changes on every build and upsets percy the hedgehog
```
notlee, line 10:
```suggestion
	// changes on every build and upsets percy the hedgehog (visual regression tests)
```

just in case :D

components/o-date/demos/src/declarative.js
```diff
@@ -1,12 +1,16 @@
...
+	const date =
+		i < dates.length - 2 ? sixHoursAgo.toISOString() : '1912-04-15T05:18Z';
```
notlee, line 12: Maybe we could check the value of `data-o-date-format` instead of using the dom position, hopefully that will break less when we add new date examples.

notlee (approved):
even better :D

PR #346, opened 2021-10-06
[ADSDEV-899] - updated width and height to 2px to better serve users with voice activation software

gitGremlin:
Originally for [this ticket](https://financialtimes.atlassian.net/browse/ADSDEV-899), there were several ways to potentially resolve this.

This is possibly the best as all as the `oNormaliseVisuallyHidden` component is already invisible by a `clip: rect(0,0,0)`, so increasing its footprint to 2px should not affect anything visually... but should help with accessibility!

Let me know if there are any questions.

### DIFF
--- a/components/o-normalise/src/scss/_mixins.scss
+++ b/components/o-normalise/src/scss/_mixins.scss
@@ -8,8 +8,8 @@
 	border: 0;
 	overflow: hidden;
 	padding: 0;
-	width: 1px;
-	height: 1px;
+	width: 2px;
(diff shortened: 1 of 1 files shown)


chee:
@notlee you know, i think that height/width combo originally happened because VoiceOver won't read out something that has a height of `0`. i think it's a min height/width rather than a max, and 2px should work just as well

notlee:
Hey, we found the same article!

notlee (approved):

PR #355, opened 2021-10-08
build every demo as part of the website build

chee:


notlee:
hey, you mentioned you were about to abandon this work to use storybook as the review app, should this PR be closed?

PR #379, opened 2021-10-29
replace pa11y test inside obt with a custom accessibility testing script

JakeChampion:
The script should be called from within the component directory that is going to be tested.

This is roughly the same code that obt used to run pa11y but it runs on all demos, not just the demo named "pa11y".
I have also updated some of the syntax used and commented out the all the pa11y rules which were ignored. The reason the ignore rules are commented out is so that we can see what errors appear with the default pa11y ruleset and make a decision upon what we should ignore.


tools/origami-build-tools/lib/tasks/test.js
```diff
@@ -1,7 +1,6 @@
 'use strict';
 
 module.exports = function (cfg) {
-	const pa11y = require('./pa11y');
 	const karma = require('./karma');
```
chee, line 4: :D

chee (approved):
This looks great to me.

tools/a11y/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "version": "1.0.0",
+  "description": "Run accessibility audit on the demos of a component",
+  "main": "index.js",
+  "bin": "index.js",
+  "type": "module",
+  "scripts": {
+    "lint": "exit 0",
+    "test": "exit 0"
+  },
+  "keywords": [],
```
JakeChampion, line 12: this is missing dependencies for pa11y...

PR #385, opened 2021-10-29
Remove math.div warnings when compiling components

JakeChampion:
This PR introduces a temporary helper library named `@financial-times/math` to avoid newer dart-sass versions printing warnings when compiling Origami.

The library contains a `div` function which is defined twice, once in `index.scss` and again in `index.import.scss`

When compiling using dart-sass, the `index.import.scss` file will be used, which implements the `div` function with the new `math.div` if supported and with `/` if not supported.

When compiling using lib-sass or ruby-sass, the `index.scss` file will be used, which implements the `div` function using the `/` operator (which is deprecated in dart-sass)

As well as a `div` function, the library also re-exports all the functions from `mathsass` but re-implements the ones which used the `/` operator, such as `pow` and `sqrt`.

I've also added an internal fork of `sass-mq` which makes use of `@financial-times/math` to remove the last remaining sass warnings which happen


JakeChampion:
@Financial-Times/origami-core This is now ready for a review :-D

libraries/sass-mq/README.md
```diff
@@ -0,0 +1,348 @@
+<div align="center">
+	<img width="500" height="500" src="https://avatars3.githubusercontent.com/u/9341289?v=3&s=500" alt="Awesome">
+</div>
+
+# Media Queries with superpowers [![Build Status](https://api.travis-ci.org/sass-mq/sass-mq.svg?branch=master)](https://travis-ci.org/sass-mq/sass-mq)
```
notlee, line 5: would be great to add a note here as to why we're forking it, as in the math library :)

libraries/math/index.import.scss
```diff
@@ -0,0 +1,42 @@
...
+		@return math.pow($base, $exponent);
+	} @else {
+		@return pow($base, $exponent);
+	}
+}
+
+@function div($number1, $number2) {
+	// math.div was added in version 1.33.0
+	// import-only files were added in version 1.23.0
+	// meta.module-functions was added in version 1.23.0
```
notlee, line 34: 💪

notlee (approved):
very clever way to remove these very frustrating warnings :D

libraries/sass-mq/README.md
JakeChampion: done in [`71a92c0` (#385)](https://github.com/Financial-Times/origami/pull/385/commits/71a92c089112fbddf5055315ab82f145435444f5)

Issue #328, opened 2021-09-28
Verify component demos have valid html


notlee:
The build service throws an error rather than return invalid html from the html endpoint. However we do not
have the same check when working on components locally. For example o-header@9.0.2 has invalid demo
markup, which was not caught by tests.

![Screenshot 2021-09-28 at 11 26 50](https://user-images.githubusercontent.com/10405691/135070368-12a6eb70-1700-4e89-91a6-c87384df2be8.png)

chee:
yes!

PR #388, opened 2021-11-01
Validate component demo html

notlee:
I'm using [html-validate](https://www.npmjs.com/package/html-validate) with the `html-validate:standard` preset.
However I've overridden the preset so `buttons` don't require a `type` and `th` elements don't require a `scope`.
I don't think either of these are actually required and the browser default is fine.
>Enables rules related to validating according to the WHATWG HTML standard (Living Standard).
>Use this preset if you want validation similar to the Nu Html Checker and similar tools.

Here's where the magic happens:
https://github.com/Financial-Times/origami/compare/validate-component-demo-html?expand=1#diff-51a745927fcfcc4176f0a09906a4f9bdcf155b460c7960bc25d06349d83cfe70R1

See commits for more details.

Closes: https://github.com/Financial-Times/origami/issues/328

### DIFF
--- a/components/o-colors/demos/src/color-mixer/color-mixer.js
+++ b/components/o-colors/demos/src/color-mixer/color-mixer.js
@@ -17,7 +17,6 @@ document.addEventListener('DOMContentLoaded', function() {
 	mixColors(mixer.value, base.value);
 
 	//set visible hex value and sass function to default values of mixColors
-	console.log(form['range']);
 	form['range'][5].checked = true;
 	const defaultHex = getComputedStyle(document.documentElement).getPropertyValue(`--o-colors-${mixer.value}-50`);
 	fillCodeSnippets(defaultHex, mixer.value, base.value, 50);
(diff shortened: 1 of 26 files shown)


components/o-colors/demos/src/color-mixer/color-mixer.js
```diff
@@ -17,7 +17,8 @@ document.addEventListener('DOMContentLoaded', function() {
 	mixColors(mixer.value, base.value);
 
 	//set visible hex value and sass function to default values of mixColors
-	forms[1]['range'][5].checked = true;
+	console.log(form['range']);
```
JakeChampion, line 20: rogue console.log found in the wild

components/o-table/demos/src/base.mustache
```diff
@@ -41,8 +41,8 @@
 				<thead>
 					{{#multipleHeaderRows}}
 					<tr>
-						<th scope="col" role="columnheader" scope="col" role="columnheader" colspan="4">Description</th>
-						<th scope="col" role="columnheader" scope="col" role="columnheader" colspan="4">Price</th>
+						<th scope="col" role="columnheader" colspan="4">Description</th>
```
JakeChampion, line 44: nice :D

JakeChampion (approved):

notlee:
I don’t think this is actually run in CI right? Same with https://github.com/Financial-Times/origami/pull/379? Do we need to update the github actions?

JakeChampion:
> I don’t think this is actually run in CI right? Same with #379? Do we need to update the github actions?

I've made the change in `scripts/regenerate-ci-config-files.js` to run this in ci 👍

PR #389, opened 2021-11-02
backstage: remove all the build, start and test scripts which were added when we had separate repos and heroku review apps

JakeChampion:


JakeChampion:
closing in favour of https://github.com/Financial-Times/origami/pull/390

PR #397, opened 2021-11-03
backstage: extract true-sass test runner from obt and into it's own tool

JakeChampion:
One less thing to use obt for :D


scripts/component/test.bash
```diff
@@ -2,4 +2,5 @@
 set -e
 origami-build-tools test
 a11y
-validate-component-demo-html
\ No newline at end of file
+validate-component-demo-html
+test-sass
```
chee, line 6:
```suggestion
test-sass

```

tools/test-sass/index.js
```diff
@@ -0,0 +1,13 @@
...
+
+import {globbySync as glob} from "globby"
+const sassTestFiles = glob('./test/**/index.test.scss', {onlyFiles: true})
+if (sassTestFiles.length) {
+  try {
+    await $`mocha ../../tools/test-sass/sass-test-suite.js`
+  } catch {
+    process.exitCode = 1;
+  }
+}
```
chee, line 13:
```suggestion
}

```

tools/test-sass/index.js
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
+
+import {$} from "zx"
+
+import {globbySync as glob} from "globby"
+const sassTestFiles = glob('./test/**/index.test.scss', {onlyFiles: true})
+if (sassTestFiles.length) {
+  try {
+    await $`mocha ../../tools/test-sass/sass-test-suite.js`
```
chee, line 9: i don't know why, but this particular `../../` seems confusing to me. probably fine though.

chee (approved):

tools/test-sass/index.js
JakeChampion: I think we could change this to using __dirname instead

JakeChampion: @chee I changed it to use __dirname and it works -- [`1854e61` (#397)](https://github.com/Financial-Times/origami/pull/397/commits/1854e61959e419e4ed2429dc8bb5581a26063dc0)

tools/test-sass/index.js
```diff
@@ -1,12 +1,16 @@
 #!/usr/bin/env node
 
-import {$} from "zx"
+import {$} from 'zx'
+import {globbySync as glob} from 'globby'
+import { dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
```
chee, line 6: nice :)))

PR #409, opened 2021-11-08
move readme linting to the components/lint.bash

chee:


JakeChampion (commented):
Look at that diff!
1 line added and 90 removed -- functionality has stayed the same :D

JakeChampion (approved):

PR #411, opened 2021-11-08
fix: Remove Build Service v2 use from origami website

notlee:
Fixes component demos embedded with the "demo" partial template e.g. `{% include demo.html component="o-table" demo="row-stripes" brand="core" height="242px" %}`


notlee:
**Ready for review but do not merge.** Found 2 issues working on this:
1. The Build Service v3 api docs are inaccurate for the `fonts` endpoint, I'll update that 👍 
2. Looks like these updated Build Service urls are returning an error saying the requested version can't be found in the npm registry, but they do exist. @JakeChampion I think this is an issue with the Build Service over an expected cache delay? 

https://www.ft.com/__origami/service/build/v3/bundles/css?components=o-normalise@^3.0.3,o-typography@^7.0.3,o-layout@^5.0.6,o-header-services@^5.1.0,o-footer-services@^4.0.3,o-syntax-highlight@^4.0.1,o-message@^5.1.0,o-table@^9.0.3,o-autoinit@^3.0.1&brand=internal&system_code=origami

JakeChampion:
>     2. Looks like these updated Build Service urls are returning an error saying the requested version can't be found in the npm registry, but they do exist. @JakeChampion I think this is an issue with the Build Service over an expected cache delay?
> 

Yes, I forgot that the work to improve the build-services caching logic wasn't completed, I'm looking into it now

JakeChampion (approved):

PR #412, opened 2021-11-08
Update README.md

ka-tsu-ra:
I didn't understand that bit of the readme, so I added the words I needed in order to understand it


notlee (approved):
thank you! :)

Issue #407, opened 2021-11-08
Add new tool to verify the sass of components is valid


JakeChampion:
This is part of a bigger piece of work on deconstructing origami-build-tools into smaller isolated tools -- #402

I think the verify sass task could be replaced with a tool which runs stylelint within the components folder, using the components `.stylelintrc.cjs` configuration

The code within origami-build-tools for the existing verify sass task is located here -- https://github.com/Financial-Times/origami/blob/595aca7158ce80f271c445211c9b666a8f582617/tools/origami-build-tools/lib/tasks/verify-sass.js

JakeChampion:
PR for this work is -> https://github.com/Financial-Times/origami/pull/422

PR #422, opened 2021-11-10
replace obt verify sass task with stylelint cli

JakeChampion:
Resolves https://github.com/Financial-Times/origami/issues/407

It's great that lint.bash exists - we can use the stylelint cli directly and run it whenever a component has a .stylelintrc.cjs file


scripts/component/test.bash
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 set -e
+npx remark-cli -f -q README.md --no-stdout
+if test -f ".stylelintrc.cjs"; then
+    npx stylelint "**/*.scss" --config=".stylelintrc.cjs"
```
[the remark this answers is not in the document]
JakeChampion: I put it in the wrong file 🙈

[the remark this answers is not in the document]
JakeChampion: @emortong This is now fixed, thanks for spotting that!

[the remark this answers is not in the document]
emortong: have you pushed the file? can't find it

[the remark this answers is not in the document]
emortong: I see it now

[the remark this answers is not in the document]
JakeChampion: It should be there now 🤞 
https://github.com/Financial-Times/origami/pull/422/files#diff-ec1ed95d5fa926281722db4c69442ed2eb829f8e61cf2be5334f7d01ac0221a7

scripts/component/lint.bash
```diff
@@ -1,4 +1,7 @@
 #!/bin/bash
 set -e
 npx remark-cli -fq README.md
+if test -f ".stylelintrc.cjs"; then
+    npx stylelint "**/*.scss" --config=".stylelintrc.cjs"
```
emortong, line 5: **question**: will "**/*.scss" read through the correct files? I am doing as chee suggested using `$workspace/**/*.{js,tsx}`

JakeChampion: `**/*.scss` will read the correct files as the working directory when lint.bash runs will be the component whose lint task is running

I didn't know about `$workspace` - do you know where that is being set?

JakeChampion: I've just pushed a commit which should hopefully show only o-table failing and no other components

emortong: oh it was being set right before it as `workspace=$(pwd)` sorry I was not clear.. 

I was asking because I was getting an error since my implementation was trying to stylelint on this file https://github.com/Financial-Times/origami/blob/main/apps/website/assets/main.scss#L1 and it will fail, but I think I might have been doing something wrong

on your implementation, where is `.stylelintrc.cjs` supposed to be? root level or on each component?

JakeChampion: It is on each component, I went with this approach as that way a component can opt-in to running our stylelint rules

emortong: ok, I talked to Chee, I was running the script wrong, that's why I was getting a bunch of errors, I'll push up our implementation of the `verify-javascript` so that you can see how we are using the `$workspace` variable

emortong: @JakeChampion https://github.com/Financial-Times/origami/pull/423

emortong:
I think these should also be removed https://github.com/Financial-Times/origami/blob/main/tools/origami-build-tools/test/unit/tasks/verify.test.js#L16

emortong (approved):
🎉

Issue #404, opened 2021-11-08
Add new tool to verify the javascript of components is valid


JakeChampion:
This is part of a bigger piece of work on deconstructing origami-build-tools into smaller isolated tools -- #402

I think the verify javascript task within origami-build-tools could be replaced with a tool which runs eslint within the components folder, using the components `.eslintrc.cjs` configuration

The code within origami-build-tools for the existing verify javascript task is located here -- https://github.com/Financial-Times/origami/blob/595aca7158ce80f271c445211c9b666a8f582617/tools/origami-build-tools/lib/tasks/verify-javascript.js

PR #423, opened 2021-11-11
Replace obt verify-javascript task with eslint cli

emortong:
Resolves https://github.com/Financial-Times/origami/issues/404

using eslint cli directly in lint.bash whenever a component has a .eslintrc.cjs file


emortong:
@chee should I remove all of the component's `.eslintignore` file now that we are using `--ignore-pattern` ?

scripts/component/lint.bash
```diff
@@ -1,4 +1,10 @@
 #!/bin/bash
 set -e
 npx remark-cli -fq README.md
+workspace=$(pwd)
+echo $workspace
+cd ../..
+if test -f "$workspace/.eslintrc.cjs"; then
+    npx eslint -c $workspace/.eslintrc.cjs --ignore-pattern "demos/local/** " --ignore-pattern "bookmarklet/" "$workspace/**/*.js" --no-error-on-unmatched-pattern
+fi
 origami-build-tools verify
```
chee, line 10:
```suggestion
npx remark-cli -fq README.md
origami-build-tools verify
workspace=$(pwd)
echo $workspace
cd ../..
if test -f "$workspace/.eslintrc.cjs"; then
    npx eslint -c $workspace/.eslintrc.cjs --ignore-pattern "demos/local/** " --ignore-pattern "bookmarklet/" "$workspace/**/*.js" --no-error-on-unmatched-pattern
fi
```

oh, obt verify still needs to be run from the component's directory

scripts/component/lint.bash
```diff
@@ -1,4 +1,10 @@
 #!/bin/bash
 set -e
 npx remark-cli -fq README.md
+workspace=$(pwd)
+echo $workspace
+cd ../..
+if test -f "$workspace/.eslintrc.cjs"; then
+    npx eslint -c $workspace/.eslintrc.cjs --ignore-pattern "demos/local/** " --ignore-pattern "bookmarklet/" "$workspace/**/*.js" --no-error-on-unmatched-pattern
+fi
```
JakeChampion, line 9: we will need to cd back into the workspace before running origami-build-tools verify, other it will run across the entire repo, either that or we can run origami-build-tools verify before the other commands

JakeChampion: chee just explained to me why we have this code instead of `npx eslint` within the workspace -- it is to get annotations working 🎉 
Could we add a comment which explains that? I will definitely need reminding about this when I come to look at this file in the future

emortong: I'll modify the script to run origami-build-tools verify before and add a comment

JakeChampion (approved):

JakeChampion (approved):

PR #424, opened 2021-11-11
re-import o-subs-card v6

chee:
o-subs-card was originally imported in an in-between state

### DIFF
--- a/components/o-subs-card/MIGRATION.md
+++ b/components/o-subs-card/MIGRATION.md
@@ -1,5 +1,23 @@
 # Migration Guide
 
+## Migrating from v5 to v6
+
+The "more/less" toggle has been updated to use a `button` element instead of a `div` to improve keyboard accessibility.
+
+To upgrade, replace the "read more" `div` tag with a `button` tag in your markup.
+```diff
+- <div class='o-subs-card__read-more'>Read more</div>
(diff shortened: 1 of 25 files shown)


components/o-subs-card/MIGRATION.md
```diff
@@ -83,13 +101,21 @@ The markup has been rearranged, and some classes have been removed.
 			</div>
 -			<div class="o-subs-card__copy-pitch">Access to FT.com on your desktop, mobile and tablet</div>
 				<div class="o-subs-card__expander">
+<<<<<<< HEAD
 -		 		<div class='o-subs-card__read-more'>Read more</div>
+=======
+-		 		<div class='o-subs-card__read-more'></div>
+>>>>>>> de60d1048d796fe4bd53ccb4c8a7b39dccaece06
```
JakeChampion, line 108: I think this is meant to be the below but I'm not sure, I got this from looking at the code examples in the readme
```suggestion
-		 		<div class='o-subs-card__read-more'>Read more</div>
+		 		<button class='o-subs-card__read-more'></button>
```

chee: well... this is from the old migration guide from v1 to v2 so i think it was a mistake in the original pull request that this changed at all

JakeChampion (approved):

PR #436, opened 2021-11-12
fix: make obt demo build the demos for every brand supported by the component

JakeChampion:
I thought I'd fix this bug so that when we come to extract building demos from obt, we know that it should build all the brands when no flag is supplied


tools/origami-build-tools/lib/tasks/demo.js
```diff
@@ -13,8 +14,18 @@ module.exports = function (cfg) {
...
-			return buildDemo(config);
+		task: async () => {
+			if (config.brand) {
+				return buildDemo(config);
+			} else {
+				const brands = await getModuleBrands(config.cwd)
+				const demos = []
+				for (const brand of brands) {
+					const brandedConfig = Object.assign({brand}, config);
+					demos.push(buildDemo(brandedConfig));
```
notlee, line 25: doesn't a build for one brand override the build for another?

JakeChampion: Good spot, you're right - that is now fixed in [`b27d6a1` (#436)](https://github.com/Financial-Times/origami/pull/436/commits/b27d6a12352127a215bfa02eb6b8e2430cb5f24f)

JakeChampion:
@emortong - As you are working on extracting this logic I thought it would be good to make you aware of this  bug in the current demo build script, which this PR is attempting to fix

emortong:
@JakeChampion thanks for letting me know, this will cause some conflicts, but once this fix is approved I can resolve them

tools/origami-build-tools/lib/tasks/demo-build.js
```diff
@@ -129,20 +131,19 @@ function buildDemoHtml(buildConfig) {
...
 			// core will replace master brand, this is temporary whilst origami build service is updated to support both.
 			// we're updating the o-brand component and build tools first as origami build service uses branded components
 			// for integration tests.
 			const brand = buildConfig.brand === 'core' ? 'master' : buildConfig.brand;
 			data.oDemoTitle = moduleName + ': ' + buildConfig.demo.name + ' demo';
 			data.oDemoDocumentClasses = buildConfig.demo.documentClasses || buildConfig.demo.bodyClasses;
 
-			data.oDemoComponentStylePath = sassPath ?
-				path.basename(sassPath).replace('.scss', '.css') :
+			console.log(buildConfig.demo.sass, buildConfig.demo.sassDestination)
```
emortong, line 141: is this `console.log` supposed to be here or is it leftover from debugging?

JakeChampion: Definitely debugging :D

JakeChampion:
```suggestion
```

chee (approved):
I think we can simplify the percy script now too

https://github.com/Financial-Times/origami/blob/main/scripts/percy.js#L59-L86

JakeChampion:
> I think we can simplify the percy script now too
> 
> [`main`/scripts/percy.js#L59-L86](https://github.com/Financial-Times/origami/blob/main/scripts/percy.js?rgh-link-date=2021-11-16T23%3A47%3A57Z#L59-L86)

Nice! :D

PR #437, opened 2021-11-12
backstage: add content to the layout sections and decrease viewport

JakeChampion:
The fixture is now closer to what would be written in a real-world
project, projects are not only a set of headings, they also have
paragraphs.

This is staying as a draft until I prove that the tests can pass 1000 times in a row


JakeChampion:
It passed 1000 times in a row :D

notlee (approved):

PR #438, opened 2021-11-12
backstage: upgrade to npm 8

chee:


chee:
@chee for monday

just remembered we wanna redo this with spaces because release-please uses spaces when it outputs the package.json after an update

also the libraries will need updated

chee:
can't see a reason for just one of the components to be failing with EBADENGINE

chee:
@chee split this into two

PR #440, opened 2021-11-15
Allow configuring the path to the manage cookies page

chee:
specialist titles used to need it to be on /preferences/cookies, but will be
updating their pages to /preferences/manage-cookies. let's let them
set it.

this code could be less algorithmic. it appears flexible, but is not. it's
tied hard to the server and routing. i feel like maybe the component
should live alongside the server lib it's associated with, rather than
in origami.

### DIFF
--- a/components/o-cookie-message/README.md
+++ b/components/o-cookie-message/README.md
@@ -113,6 +113,16 @@ const cookieMessage = new oCookieMessage(null, { theme: 'alternative' });
 ```
 
 
+### Custom Cookie Management path
+
+To use a different cookie management path, use the `manageCookiesPath` option:
+
+```js
+import oCookieMessage from '@financial-times/o-cookie-message';
+const cookieMessage = new oCookieMessage(null, { manageCookiesPath: 'preferences/manage-cookies' });
+```
+
+
 ### Firing an oDomContentLoaded event
 
 ```js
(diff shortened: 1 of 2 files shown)


JakeChampion:
>  i feel like maybe the component
> should live alongside the server lib it's associated with, rather than
> in origami.

Yes, I agree with that, do you know who owns the server library?

chee:
> > i feel like maybe the component
> > should live alongside the server lib it's associated with, rather than
> > in origami.
> 
> Yes, I agree with that, do you know who owns the server library?

no idea. i don't know if there even is one. but it seems to me like the consent server should be a node library that provides front-end templates.

JakeChampion (approved):

notlee:
Would it be possible to add docs to the readme?

Agree this component should be less tied to implementation 👍 as we chatted about we could create a component to provide a UI without assumption around how consent is managed

notlee:
Is this still something we need? I notice this PR removed the option:
https://github.com/Financial-Times/o-cookie-message/pull/155/files

JakeChampion:
@notlee I've asked in the specialist-dev slack channel for their input on this --> https://financialtimes.slack.com/archives/C7D7KQEN9/p1641916238016600

notlee:
thanks @JakeChampion, we should add a url option to the storybook if so
https://github.com/Financial-Times/origami/pull/539

notlee:
this has been open for a couple of months and there's no reply in the slack message, shall we close this PR? (and update o-cookie-message to remove business logic in the future)

JakeChampion:
I think we could make this release, as it was a blocker for the team at the time - we should also note down that we want to deprecate this sometime in the future

notlee:
okay, it would be good to update the readme with an example before releasing though

PR #441, opened 2021-11-15
clean up all the linting errors!!!

### LINKED ISSUE #442: Do not run percy because of a feat or fix unless that feat or fix touches the target component's files
https://github.com/Financial-Times/origami/blob/main/scripts/percy.js#L20-L37

we need to filter the commits to only the ones that touched files in the workspace before checking if a `fix` or `feat` commit has happened, otherwise any `fix` or `feat` commit anywhere will cause :hedgehog:  to run even if the `fix` or `feat` did not touch this component.

chee:
~i've reached components/o-forms and am stopping for the evening. will carry on mañana~

while waiting for the rice to coo~k~l i did some more. i will carry on from o-~share~subs-card tomorrow.

this also closes #442 for bants


scripts/component/lint.bash
```diff
@@ -8,7 +8,7 @@ verify-origami-json
 verify-package-json
 
 if test -f ".eslintrc.cjs"; then
-	args=(-c ".eslintrc.cjs" --no-error-on-unmatched-pattern)
+	args=(-c ".eslintrc.cjs" --no-error-on-unmatched-pattern --fix)
```
JakeChampion, line 11: Question: do we want these --fix flags?

JakeChampion (approved):

PR #450, opened 2021-11-17
fix: stop building specific demos for each brand in percy

JakeChampion:
this is all handled within obt demo now instead, which is great


chee (commented):
much nicer

chee (approved):
much nicer

PR #451, opened 2021-11-19
feat: add tool to help teams avoid installing versions of origami which are only available on npm and not also on bower

JakeChampion:
This is a first attempt, nothing here is decided, we can change any and all of it 👍 

Below is a example of it running on next-article (after I installed o-table into next-article):

```
❯ ../origami/tools/origami-bower-helper/index.js
This project ("ft-next-article") is installing versions of Origami dependencies which are not available on Bower.
This may cause build failures for this project or any projects which depend on "ft-next-article" and have not yet upgraded to use the latest version of Origami.

The last version of "@financial-times/o-table" which is on both npm and bower is "8.1.5".
Downgrade "@financial-times/o-table" to at least "8.1.5" in order to avoid build failures for your users.

Run these commands to downgrade:
npm install --save-exact --save "@financial-times/o-table@8.1.5"
```


tools/origami-bower-helper/README.md
```diff
@@ -0,0 +1,25 @@
+# Origami Bower Helper
```
chee, line 1: let's workshop the name, make it more specific. maybe something like "origami-version-constrainer" "origami-bower-safe-version-supervisor" lol

tools/origami-bower-helper/lastVersionsAvailableOnBower.js
```diff
@@ -0,0 +1,62 @@
...
+    "@financial-times/o-loading":"4.0.4",
+    "@financial-times/o-message":"4.2.6",
+    "@financial-times/o-meter":"2.0.6",
+    "@financial-times/o-normalise":"3.0.1",
+    "@financial-times/o-overlay":"3.1.0",
+    "@financial-times/o-quote":"4.1.6",
+    "@financial-times/o-share":"7.6.0",
+    "@financial-times/o-spacing":"2.1.0",
+    "@financial-times/o-stepped-progress":"2.0.7",
+    "@financial-times/o-subs-card":"6.0.0-bower",
```
chee, line 48: these `-bower` suffixes are not available on npm, so will need to be dropped when saving to the package.json

tools/origami-bower-helper/index.js
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env node
+
+import process from 'node:process'
+import {readPackageAsync} from "read-pkg"
+import semverExtra from "semver-extra"
+import origamiComponents from "./lastVersionsAvailableOnBower.js"
```
chee, line 6: i think in most of our other code we use kebab case in file names

tools/origami-bower-helper/index.js
```diff
@@ -0,0 +1,73 @@
...
+import majors from "major-versions"
+import isBounded from "semver-bounded"
+import dedentTabs from "dedent-tabs"
+const dedent = dedentTabs.default
+
+function checkDepdencies(dependencies) {
```
chee, line 12: i think this function should be called `checkDependencies`

tools/origami-bower-helper/index.js
```diff
@@ -0,0 +1,73 @@
...
+		for (const [name, lastVersionAvailableOnBower] of [...dependenciesToDowngrade, ...devDependenciesToDowngrade, ...peerDependenciesToDowngrade]) {
+			console.error(dedent`The last version of "${name}" which is on both npm and bower is "${lastVersionAvailableOnBower}".	
+			Downgrade "${name}" to at least "${lastVersionAvailableOnBower}" in order to avoid build failures for your users.\n`)
+		}
+		
+		console.error(`Run these commands to downgrade:`)
+		
+		for (const [name, lastVersionAvailableOnBower] of dependenciesToDowngrade) {
+			console.error(`npm install --save-exact --save "${name}@${lastVersionAvailableOnBower}"`)
+		}
```
chee, line 56: maybe we should add a note to this output that `--save-peer` should be used if the project they are working on is a library, even if they had it in `dependencies` before

tools/origami-bower-helper/lastVersionsAvailableOnBower.js
```diff
@@ -0,0 +1,62 @@
...
+    "@financial-times/o-teaser-collection":"3.0.8",
+    "@financial-times/o-toggle":"2.1.3",
+    "@financial-times/o-tooltip":"4.1.0",
+    "@financial-times/o-topper":"3.1.7",
+    "@financial-times/o-tracking":"4.1.2",
+    "@financial-times/o-typography":"6.4.6",
+    "@financial-times/o-utils":"1.1.7",
+    "@financial-times/o-viewport":"4.0.5",
+    "@financial-times/o-visual-effects":"3.0.6",
+}
```
chee, line 62:
```suggestion
}

```

tools/origami-bower-helper/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "license": "MIT",
+  "dependencies": {
+    "dedent-tabs": "^0.10.1",
+    "major-versions": "^2.2.1",
+    "read-pkg": "^6.0.0",
+    "semver": "^7.3.5",
+    "semver-bounded": "^2.0.3",
+    "semver-extra": "^3.0.0"
+  },
+  "private": false
```
chee, line 21: !

chee (commented):
some notes

JakeChampion:
@chee thanks for the review - I've gone through and applied all the suggestions now 👍

chee (approved):

PR #457, opened 2021-11-22
Make percy snapshot from the demos/local folder which is where the demos live

JakeChampion:
Previously Percy would move all the demos from `demos/local` to `demos/percy` and rename the demos to include the brand in their filename.

In a [previous patch](https://github.com/Financial-Times/origami/pull/450), we stopped doing the moving and renaming of demos but we did not update the percy command to run from `demos/local`


chee (approved):
ah, very cool ++

PR #459, opened 2021-11-22
Add a script to check the workspace has private set explicitly

### LINKED ISSUE #452: private: false
TODO: explain this

chee:
this seems better than it being possible to publish something by mistake by not
setting `private: true` amirite?

closes #452


JakeChampion:
> this seems better than it being possible to publish something by mistake by not setting `private: true` amirite?

You are right ☺️

.github/workflows/check-privacy.yml
```diff
@@ -0,0 +1,22 @@
...
+    - uses: volta-cli/action@v1
+    - uses: actions/cache@v2
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+    - run: npm ci
+    - run: npm run regenerate
+    - run: ./scripts/check-for-changes.bash
```
JakeChampion, line 22: I think this is meant to call ./scripts/check-privacy.js

chee: lol wtf. did i not p[ush ?

scripts/check-privacy.js
```diff
@@ -0,0 +1,28 @@
...
+let message = `
+make sure private is true for packages that should be published, and false for
+those that shouldn't.
+`
+
+for (let cwd of await workspaces.paths()) {
+	let pkg = await readPackage({cwd})
+	if (typeof pkg.private != "boolean") {
+		log(`${cwd}/package.json#private is not a boolean`)
+		code = 130
```
JakeChampion, line 20: Nice

scripts/check-privacy.js
```diff
@@ -0,0 +1,28 @@
...
+import {readPackage} from "read-pkg"
+
+let {exit} = process
+let {log} = console
+
+/** exit code */
+let code = 0
+
+let message = `
+make sure private is true for packages that should be published, and false for
```
JakeChampion, line 12: I think this sentence might be the wrong way around

chee: oh you're right

chee, line 12:
```suggestion
make sure private is false for packages that should be published, and true for
```

JakeChampion (approved):

PR #464, opened 2021-11-23
make o-no work with the * range and exact versions

JakeChampion:
previously o-no could only work with standard semver ranges and not the npm specific `*` range or with exact-versions

this patch brings support for both those features and also corrects the latest version for o-normalise (we still had an npm-only tag on the old o-normalise repository which has since been deleted)


tools/origami-bower-safe-version-supervisor/index.js
```diff
@@ -1,32 +1,45 @@
...
+						if (semverExtra.gt(versionRange, lastVersionAvailableOnBower)) {
+							dependenciesToDowngrade.push([name, lastVersionAvailableOnBower, versionRange])
+						}
+					} else {
+						const versions = [lastVersionAvailableOnBower, ...majors(versionRange)].map(version => {
+							return String(semverExtra.coerce(version, {
+								loose: true
+							}))
+						})
+						console.log({versions},majors(versionRange))
```
notlee, line 37:
```suggestion
```
I think this console log is there accidentally?

notlee (approved):
One potentially rogue console log there, but otherwise looks good 👍

tools/origami-bower-safe-version-supervisor/index.js
JakeChampion: Yep!

PR #468, opened 2021-11-30
fix!: Drop IE9 support in o-stepped-progress

chee:
We currently have two hacks, one for IE9 and one that overrides the hack for
IE10 and IE11. We don't support IE9 so instead of doing two things we can do
none, which is the number of things I prefer to do.

I'm releasing this as a breaking change, but really I don't think it is because
I don't think we ever told anyone we support IE9 and this component was made
after we no longer supported IE9. What ya gonna do, though? call the cops?

live laugh love

stay humble

🧱

### DIFF
--- a/components/o-stepped-progress/MIGRATION.md
+++ b/components/o-stepped-progress/MIGRATION.md
@@ -1,5 +1,9 @@
 # Migration
 
+## Migrating from v3 to v4
+
+Support for IE9 has been dropped. No other changes.
+
 ## Migrating from v2 to v3
 
 Support for Bower and version 2 of the Origami Build Service have been removed.
--- a/components/o-stepped-progress/README.md
+++ b/components/o-stepped-progress/README.md
@@ -1,4 +1,5 @@
 # o-stepped-progress
+
 Track progress through a multi-step process, such as a form.
 
 - [Usage](#usage)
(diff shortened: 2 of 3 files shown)


notlee:
> I don't think it is because I don't think we ever told anyone we support IE9 and this component was made after we no longer supported IE9

Plus in the past we've dropped IE8/9/10 support without releasing a breaking change, e.g o-grid. Maybe technically a no-no but I'd be in favour of continuing that trend and making this a patch release too, for the smoothest upgrade path for Origami users. 

Maybe our promise going forward should be "we support what the ft supports" rather than a specific browser, unless our components are ever generic enough to be helpful to wider open source users...

chee:
> > I don't think it is because I don't think we ever told anyone we support IE9 and this component was made after we no longer supported IE9
> 
> Plus in the past we've dropped IE8/9/10 support without releasing a breaking change, e.g o-grid. Maybe technically a no-no but I'd be in favour of continuing that trend and making this a patch release too, for the smoothest upgrade path for Origami users.

semver shouldn't have knowledge of how the end users are using the component. it's semantic, it has meaning. the previous version supported a browser that this version doesn't, so that's a breaking change. unless our documentation indicated it only ever supported browsers above that. i'm not clear if our documentation ever did, so i'm erring on the side of caution.

it shouldn't be too much of a hassle to upgrade to a new major version of a leaf component like this.

> 
> Maybe our promise going forward should be "we support what the ft supports" rather than a specific browser, unless our components are ever generic enough to be helpful to wider open source users...

I don't think it's a good idea for our support to be able to change independently of the code. I think we should document what browsers a component supports at time of release and do major versions when we drop support for those browsers. and, we should optimize our components so upgrading major versions is less of a hassle.

notlee:
>I don't think it is because I don't think we ever told anyone we support IE9 and this component was made after we no longer supported IE9

I think you're right on that and would be confident releasing this as a patch, but if you'd prefer to be cautious could you add a migration guide to go along with the major release?

chee:
> > I don't think it is because I don't think we ever told anyone we support IE9 and this component was made after we no longer supported IE9
> 
> I think you're right on that and would be confident releasing this as a patch, but if you'd prefer to be cautious could you add a migration guide to go along with the major release?

:+1: done

notlee (approved):

PR #475, opened 2021-12-08
feat: support internal brand on o-autocomplete

maxcbc:
# feat: support internal brand on o-autocomplete

## What
o-autocomplete doesn't currently support the internal brand, but its only branded bit is a single colour reference which is available in the internal brand. This adds the same settings for internal as for core.

## Why
I want to use this in a product using the internal brand


JakeChampion:
@maxcbc do you know if a designer has approved this design for internal products?

JakeChampion:
I don't think the teal color is in the internal palette -- https://github.com/Financial-Times/origami/blob/49bf07f2eeca8c1806d64c2d603ade9af7f7d30c/components/o-autocomplete/main.scss#L133-L136

image

notlee:
Teal is, we can add support relatively easily :) https://registry.origami.ft.com/components/o-colors@6.4.0/?brand=internal#demo-primary-palette

maxcbc:
@JakeChampion nope on the designer check, this is entirely my own initiative.

JakeChampion:
We would want a designer to sign off on this before accepting it into Origami

maxcbc:
@JakeChampion no worries, I'll convert to draft for now and run it by our designer.

JakeChampion:
Actually, I realise this requires no actual design changes, let's add internal support - we will discuss with the design team about adding this to the UI Kit and getting guidelines written up

JakeChampion (approved):

PR #477, opened 2021-12-09
change the accessibility tests to use axe directly and not via pa11y

JakeChampion:
This changes the a11y test suite to use axe directly with the axe/playwright system instead of using pa11y which was causing us some issues which we could not find a solution for. Pa11y began to return lots of false positives which were not detected when running the same test using several other accessibility testing tools, including using the exact same rules with Axe directly that Pa11y as passing to axe.


JakeChampion:
The remaining components which have failing accessibility tests are:

# o-tabs
The issue is the tabs have nested interactive elements, the `role=tab` should be on the anchor elements. To fix this would be a breaking change and we could take this opportunity to rewrite the component following the example version in "inclusive-components" which is progressively enhanced and accessible. This is a demo the the "inclusive-components" tab component https://codepen.io/heydon/full/veeaEa 

The command I use to test this component: `npm run build -w 'components/o-tabs' && npm t -w 'components/o-tabs'`

# o-colors
Some of the demos use o-tabs. We will need to fix the above o-tabs accessibility issue and then make the o-colors demos update to the version of o-tabs with the accessibility fix.

The command I use to test this component: `npm run build -w 'components/o-colors' && npm t -w 'components/o-colors'`

# o-header-services 
The demo which has issues is the B2C demo, the issue is one of insufficient colour-contrast on the navigation links. I spoke with Lee about this demo and Lee thinks there are no products which use the B2C theme and that we could possibly remove it. Lee posted in the design-system-guild https://financialtimes.slack.com/archives/C01481FKWA2/p1639149529087300

The command I use to test this component: `npm run build -w 'components/o-header-services' && npm t -w 'components/o-header-services'`

# o-cookie-message 

I think this is a false positive from Axe, for some reason it is reporting that the icon used in the external link has insufficient colour-contrast but the icon is the same colour as the text and the text has good colour-contrast

The command I use to test this component: `npm run build -w 'components/o-cookie-message' && npm t -w 'components/o-cookie-message'`

# o-buttons

Same false positive as o-cookie-message, the icon in the icon button is reported as having insufficient colour-contrast but the icon is the same colour as the text and the text has good colour-contrast

The command I use to test this component: `npm run build -w 'components/o-buttons' && npm t -w 'components/o-buttons'`

# o-tooltip

Another possible false positive. This time the text in the tooltip is reported as insufficient colour contrast but the text is black and the background is white.

The command I use to test this component: `npm run build -w 'components/o-tooltip' && npm t -w 'components/o-tooltip'`

# o-normalise

I can't remember, I'll rerun it now.

The command I use to test this component: `npm run build -w 'components/o-normalise' && npm t -w 'components/o-normalise'`

tools/a11y/index.js
```diff
@@ -39,6 +40,11 @@ for (const file of builtDemoHtmlFiles) {
 	const page = await context.newPage();
 	await page.goto(pathToFileURL(file).toString());
 
+	// We sleep for 1-second to allow for short animations to finish running
+	// This is useful specifically for o-tooltip's demos which animate the opacity
+	// of the tooltip and during the animation - the contrast of the text is insufficient.
+	await setTimeout(1000);
+
```
chee, line 47: is there a way we can do this just in the tests that need it? like, do it in o-tooltip somehow? probably not, it would be nice if it could be local to the tests but i guess it can't with our testing setup

JakeChampion:
o-tooltip was failing accessibility tests on the demos which open the tooltip, when the tooltip becomes visible it does a transition on the opacity, during the transition the text has insufficient contrast - this is a race condition because after the transition the text has excellent contrast. This race condition was never happened in the previous accessibility tests which only tested the pa11y demo as the  pa11y demo never opens the tooltip...

tools/a11y/index.js
JakeChampion: I agree that would be better - maybe we could do something like have an optional "setup" function which is defined inside the components' own folder?

I was also wondering if there is a browser api in playwright which can wait until animations are completed

scripts/percy.js
```diff
@@ -9,6 +9,7 @@ const workspace = "./" + process.env.WORKSPACE
 const isPullRequest = context.payload.pull_request
 
 async function shouldPercyRun() {
+	throw new Error('no percy right now');
```
JakeChampion, line 12: When ready to run percy, we need to remove this line

components/o-tabs/demos/src/base-style.mustache
```diff
@@ -1,8 +1,8 @@
-<ul data-o-component="o-tabs" class="o-tabs" role="tablist" data-o-tabs-update-url>
-	<li role="tab"><a href="#tabContent1">Tab 1</a></li>
-	<li role="tab"><a href="#tabContent2">Tab 2</a></li>
-	<li role="tab"><a href="#tabContent3">Tab 3</a></li>
-</ul>
+<div data-o-component="o-tabs" class="o-tabs" role="tablist" data-o-tabs-update-url>
```
JakeChampion, line 1: This needs a migration guide

components/o-colors/package.json
```diff
@@ -29,6 +29,9 @@
     "@financial-times/o-brand": "^4.0.1",
     "@financial-times/math": "^1.0.0"
   },
+  "devDependencies": {
+    "@financial-times/o-tabs": "^6.2.0"
```
JakeChampion, line 33: If we change the origami-build-service demo's endpoint to install all dependencies of the component instead of only the production dependencies - then we will be able to defined demo dependencies in the package.json.devDependencies field instead of origami.json!

This has the added benefit that the demos could have their dependencies be linked within the monorepo, meaning that we could for example add a feature to o-tabs and update all the component demos which use o-tabs to use the new feature. That is currently not possible to do right now as we are not linking the dependencies for demo.

JakeChampion:
All the tests are passing 🥳

components/o-colors/package.json
JakeChampion: I made a pull-request to build-service which adds this feature -- https://github.com/Financial-Times/origami-build-service/pull/558

components/o-tabs/demos/src/base-style.mustache
JakeChampion: I should have added this before turning percy on... we've just gone over our percy screenshot limit for december 😔

components/o-colors/package.json
JakeChampion: The build-service change is now being deployed to production

JakeChampion:
rebased and all passing once again 🥳

chee:
woo, okay

.release-please-manifest.json
```diff
@@ -1 +1 @@
...labels":"6.2.0","components/o-layout":"5.2.0","components/o-lazy-load":"3.1.0","components/o-loading":"5.2.0","components/o-message":"5.2.0","components/o-meter":"3.1.0","components/o-normalise":"3.2.0","components/o-overlay":"4.2.1","components/o-quote":"5.2.0","components/o-share":"8.2.0","components/o-spacing":"3.2.0","components/o-stepped-progress":"4.0.0","components/o-subs-card":"6.2.0","components/o-syntax-highlight":"4.2.0","components/o-table":"9.2.0","components/o-tabs":"6.2.0","components/o-teaser":"6.2.0","components/o-teaser-collection":"4.2.0","components/o-toggle":"3.2.0","components/o-tooltip":"5.2.0","components/o-topper":"5.2.0","components/o-typography":"7.2.0","components/o-video":"7.2.0","components/o-viewport":"5.1.0","components/o-visual-effects":"4.1.0","libraries/math":"1.1.0","libraries/o-autoinit":"3.1.0","libraries/o-brand":"4.2.0","libraries/o-errors":"5.2.0","libraries/o-tracking":"4.2.0","libraries/o-utils":"2.1.0","libraries/sass-mq":"5.2.0","presets/eslint-config-origami-component":"2.2.0","presets/remark-preset-lint-origami-component":"2.0.0","presets/stylelint-config-origami-component":"1.0.4","tools/origami-bower-safe-version-supervisor":"1.1.2"}
```
chee, line 1: why do we have a release-please change here?

components/o-buttons/demos/src/grouped.mustache
```diff
@@ -1,5 +1,5 @@
 <div class="o-buttons-group">
-	<button class="o-buttons o-buttons--secondary" aria-selected="true">John</button>
+	<button class="o-buttons o-buttons--secondary" aria-pressed="true">John</button>
```
chee, line 2: embarrassed for us

.release-please-manifest.json
JakeChampion: that'd be from when this pull-request was targeting main and I rebased main onto this branch

chee (approved):
let's go

components/o-buttons/demos/src/grouped.mustache
JakeChampion: lol

PR #519, opened 2022-01-10
o-banner: remove unused line-separator in demo

JakeChampion:
vscode complains about it -- should we replace it with a word-break-opportunity element? `<wbr>`

vscode says this:
>The file 'custom.mustache' contains one or more unusual line terminator characters, like Line Separator (LS) or Paragraph Separator (PS).
>It is recommended to remove them from the file. This can be configured via `editor.unusualLineTerminators`.


notlee (approved):
I don't think the word break is important for the demo, so let's not 👍

Issue #370, opened 2021-10-27
show sass and js in a panel in the storybook


chee:
an addon that shows the syntax highlighted source of a file provided via parameter

PR #523, opened 2022-01-10
 feat: Add Stories for o-banner demos

notlee:
## Destroy Method

Adds a `destroy` method to o-banner. This is used in the o-banner
Story to clean up the close button when re-rendered, which o-banner
dynamically creates on `init`. This is required to support the
`suppressCloseButton` Story control.

## Hidden Controls

It appears banner headings (headingLong/headingShort) should only
be used with the small/compact layout variants. Not with the default
banner style which is full bleed. However it is possible to use
Story controls to create that view.

ok: default layout story
Screenshot 2022-01-10 at 15 00 07

bad: theme story, with layout changed to the default layout via a control
(the default layout shouldn't be used with a heading)
Screenshot 2022-01-10 at 15 00 36

`@storybook/addon-knobs` (deprecated) allowed dynamic knobs,
so we could hide the heading controls if the default layout
was selected. That's not possible with `@storybook/addon-controls`:
storybookjs/storybook#11984

I think that's probably a good thing. Support for
dynamic controls was worked on but not merge. It's
a poor experience when controls shift around.
storybookjs/storybook#13890

For now this commit hides the layout control on layout
demos, and hides the heading controls from the default
layout demo, to avoid showing the discouraged
heading + layout combination. However it is still
possible to select the base layout with heading on the
theme specific demos, so that the small/compact layout
can also be selected which is allowed.

**This could be resolved by exporting 2 templates**, one
for each kind of banner / usecase. This could make
components easier to reason with and maintain. For
now this commit sticks to one banner template as
coming up with a name without history / useage
guidelines is difficult, and we don't know that
users aren't already using headings with the base
layout - though we never intended it as far as I
can tell.

## No Custom Theme Demo

The custom theme demo has not been migrated to a Story
yet.
https://registry.origami.ft.com/components/o-banner@4.2.0#demo-custom-theme

I'm not sure there is value, and maybe harm, in showing
made up customised styles alongside those with brand/design
approval. Plus it's not clear how to re-create the style
without understanding Sass and inspecting demo code.
We probably want to have stories for customised components
at a later date, with improved guidelines around them
and demo Sass/JS pane. See related issue:
#370

## No Custom Call To Action Demo

https://registry.origami.ft.com/components/o-banner@4.2.0#demo-custom-call-to-action-layout

The description of the demo starts "Although not recommended for
design consistency..." Let's not recomend it with a demo.


components/o-banner/src/js/banner.js
```diff
@@ -270,10 +270,20 @@ class Banner {
 		}, {});
 	}
 
+	/**
+	 * Undo the init method
+	 */
+	destroy() {
```
JakeChampion, line 276: suggestion: we should create an issue to add tests for this new public method

components/o-banner/src/tsx/banner.tsx
```diff
@@ -0,0 +1,105 @@
...
+					)}
+					<div className="o-banner__actions">
+						{primaryAction && !isBannerSubmitButton(primaryAction) && (
+							<div className="o-banner__action">
+								<a href="{primaryAction.url}" className="o-banner__button">{primaryAction.copy}</a>
+							</div>
+						)}
+						{primaryAction && isBannerSubmitButton(primaryAction) && (
+							<form className="o-banner__action" method="{primaryAction.method}" encType="{primaryAction.encoding}" action="{primaryAction.action}">
+								<input className="o-banner__button" type="submit" value="{primaryAction.copy}" />
```
JakeChampion, line 92: question: should this input have an associated label?

components/o-banner/stories/banner.stories.tsx
```diff
@@ -0,0 +1,143 @@
...
+	decorators: [withDesign, withHtml],
+	args: {
+	},
+	parameters: {
+		design: {
+			type: 'figma',
+			url: 'https://www.figma.com/file/MyHQ1qdwYyek5IBdhEEaND/FT-UI-Library?node-id=0%3A1489',
+		},
+		guidelines: {
+			notion: '37cdc7ac2cac4d60a4c9f451c47a4647',
```
JakeChampion, line 20: question: how did you find this ID?

JakeChampion (commented):
I've reviewed the changes and left a suggestion and some questions

I deployed this to a review app but it looks to not work due to lots of responses returning 404s -- https://origami-webs-stories-o--wn51ap.herokuapp.com/storybook/

notlee:
>I deployed this to a review app but it looks to not work due to lots of responses returning 404s
🤔 

...I deleted it, sorry

components/o-banner/src/tsx/banner.tsx
notlee: Is it weird to have a form with only a submit button? Maybe. Seems logical using VoiceOver though

JakeChampion (approved):

PR #560, opened 2022-01-12
add polyfill.io urls to local demos

JakeChampion:
this enables us to test demos locally in older browsers such as ie11

this is relying on the origami.json.browserFeatures array which I know we plan to remove. When we do remove it, we will need to find a replacement solution for running local demos in older browsers

this code is a rewrite of what was in origami-build-tools (https://github.com/Financial-Times/origami-build-tools/blob/e7fe4046aca39bd4823f109fcd45cb4b33cd7ce2/lib/helpers/construct-polyfill-url.js) to make it work in a monorepo


notlee (approved):
Looks good! We should consider going back and adding tests for our obt replacement scripts in the near future

Issue #557, opened 2022-01-11
o-share: improve label for links - ID: DAC_Share_Horizontal_01


JakeChampion:
# Issue
This is from page 185 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-share@8.0.0&demo=horizontal&system_code=origami-repo-data&brand=master>

The links in the share element are non-descriptive and do not inform the user of their purpose. Screen reader users may think the links are to visit social media sites as opposed to sharing to the sites.
The links also open in a new window but fail to inform users of this behaviour. This issue may particularly affect screen reader users who may become disorientated by the change in behaviour.

# Screen reader comments:

> “The social media links each open in a new window, but I was not informed of this behaviour. This behaviour may confuse some Screen Reader users as it is not immediately apparent that users will need to use the Alt + TAB combination to return to the previous window. This is because the social media page opens in a separate instance of the web browser, making the Alt + left arrow’ combination to go back to the previous page redundant.
> It would benefit users if this information was presented in the link text. Alternatively, if the link was opened in the same window, they could then use the Alt + left arrow’ combination to go back to the previous page.”
> “Testing the social media links using both VoiceOver and TalkBack, I found that they opened in a new tab. Unfortunately, this was not apparent from the link text.
> Informing users that the link opens in a new tab would help to avoid confusion and ensure that users do not become disorientated by the appearance of the new tab.”

# Additional instances:
Additional instances of this issue exist on other pages throughout the website; wherever this issue occurs, they too will need to be resolved.
URL: https://www.ft.com/__origami/service/build/v3/demo?component=o-share%408.0.0&demo=small&system_code=origami-repo-data&brand=master
Journey 52.
URL: https://www.ft.com/__origami/service/build/v3/demo?component=o-share%408.0.0&demo=inverse&system_code=origami-repo-data&brand=master
Journey 53.

# Proposed Solution:
Add more description to the links which describes that the links are to share to social media.

# Change on request:
Add visually-hidden text to the links to inform screen reader users that they open in a new window.
Example:
```html
<li class="o-share__action">
[...]

PR #565, opened 2022-01-12
fix: use the o-share js render method for demos

JakeChampion:
using the render method ensures that the demo markup is up-to-date with how sharing is meant to be implemented

this resolves https://github.com/Financial-Times/origami/issues/557

I've also changed the demos to be different templates instead of using one template and several data files as I found it simpler to understand - I'm happy to revert it back to one template if that is what we prefer

screenshot of a demo of o-share which shows that the labels for the links now have accurate descriptions such as "Share on Twitter (opens in a new window)"


components/o-share/demos/src/custom.mustache
```diff
@@ -0,0 +1,12 @@
+<div data-o-component="o-share" class="o-share">
+	<ul>
+		<li class="o-share__action o-share__action--labelled">
+			<!-- demo only: forms with a submit button may be used for custom share features -->
```
chee, line 4: what does "demo only: " mean here?

JakeChampion: not really sure, I kept it from the original template - I'll see if the original commit has more context

components/o-share/demos/src/horizontal.mustache
```diff
@@ -0,0 +1,10 @@
+<div
+	data-o-component="o-share"
+	class="o-share"
+	data-o-share-links="twitter facebook linkedin whatsapp pinterest"
+	data-o-share-url="http://on.ft.com/1mUdgA2"
+	data-o-share-titleExtra="FT.com | Pharmaceuticals"
+	data-o-share-summary="US drugs group vows to maintain big British presence"
+	data-o-share-relatedTwitterAccounts="ftcompanies"
+>
```
chee, line 9: style nitpick: i think we put the `>` on the same line as the last attr in most of our other html and jsx templates?

components/o-share/origami.json
```diff
@@ -31,47 +31,33 @@
...
-			"data": "demos/src/data-inverse.json",
+			"template": "demos/src/inverse.mustache",
 			"documentClasses": "demo demo--inverse",
 			"description": "inverse share icons"
 		},
 		{
 			"title": "Vertical",
 			"name": "vertical",
-			"template": "demos/src/main.mustache",
-			"data": "demos/src/data-vertical.json",
```
chee, line 57: prefer these without the data +1

components/o-share/demos/src/custom.mustache
JakeChampion: I couldn't find more context in the original commit - shall we remove the comment?

JakeChampion: I suppose it is trying to say that any html can go into the section?

notlee: Just that it's for demo purposes, Origami doesn't support the share action seen on ft.com, which brings up a share overlay, and is still an ft.com specific implementation. 

This demo was added as part of our work to help remove css overrides overrides:
https://github.com/Financial-Times/next-article/pull/3796/files

Issue #550, opened 2022-01-11
o-header: update demo link text to not use abbreviations - ID: DAC_Link_Purpose_Link_Only_02


JakeChampion:
# Issue
This is from page 168 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-header-services@5.0.1&demo=drop-down-navigation&system_code=origami-repo-data&brand=master>

The links text ‘DOCS’ and ‘SPECS’ is non-descriptive, and some users may not understand the abbreviations used and the purpose of the links.

# Proposed Solution:
The ‘DOCS’ and ‘SPECS’ links should use the full meaning of the abbreviation to allow all users to identify the purpose of the links for example ‘Documentation’ and ‘Specification’.

chee:
i guess we should also update this on our website ^_^

PR #566, opened 2022-01-12
fix: expand all uses of spec to specification and docs to documentation

JakeChampion:
during an external accessibility audit it was mentioned that the abbreviation 'spec' and 'docs' may not be well known and it would be best to use the full words instead.

I've updated all uses of `spec` and `docs` - including in url paths that we own

resolves https://github.com/Financial-Times/origami/issues/550


apps/website/pages/manifests.md
```diff
@@ -1,13 +1,14 @@
...
 description: Origami manifest files are used to describe and configure Origami projects.
-permalink: /docs/manifests/
+permalink: /documentation/manifests/
 layout: collection
 collection_id: manifests
 collection_title: Manifests
 
 # Redirect from legacy URLs
 redirect_from:
+  - /docs/manifests/
```
notlee, line 11: 👍

components/o-colors/package.json
```diff
@@ -37,7 +37,7 @@
     "@financial-times/o-normalise": "^3.2.0",
     "@financial-times/o-overlay": "^4.2.1",
     "@financial-times/o-syntax-highlight": "^4.2.0",
-    "@financial-times/o-tabs": "^6.2.0"
+    "@financial-times/o-tabs": "^7.0.1"
```
notlee, line 40: 👀

notlee: Do demos markup need updating to migrate to o-tabs@v7?

JakeChampion: Yep, the demos already use the v7 markup but they didn't pull in the right version of o-tabs. this fixes that :-)

release notes for o-tabs v7 -- https://github.com/Financial-Times/origami/releases/tag/o-tabs-v7.0.0

notlee (approved):

Issue #549, opened 2022-01-11
o-header: make esc key close menus which were opened by hover or focus - ID: DAC_Content_On_Hover_01


JakeChampion:
# Issue
This is from page 167 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-header@9.0.1&demo=mega-menu&system_code=origami-repo-data&brand=master>

When users hover or focus the links which expand to reveal content below, the menu cannot be closed without moving the mouse (using the Esc key on the keyboard). This issue may affect low vision users who use the mouse to track focus on the page.

# Proposed Solution:
Ensure that users are able to close the expanding menu without having to use the mouse.

PR #569, opened 2022-01-14
o-header: add the ability to close any open mega-menus by pressing the escape key

JakeChampion:
this is required to meet WCAG 1.4.13 Content on Hover or Focus (Level AA)

this resolves https://github.com/Financial-Times/origami/issues/549

### DIFF
--- a/components/o-header/src/js/mega.js
+++ b/components/o-header/src/js/mega.js
@@ -24,8 +24,11 @@ function addEvents (parent, menu) {
 	});
 
 
-	const handleKeydown = (e) => {
-		if (e.keyCode === 27) {
+	const handleKeydown = (event) => {
+		const key = event.key || event.keyCode;
+
+		// Internet Explorer 11 incorrectly maps the escape key to `Esc` instead of `Escape`
+		if (key === 'Escape' || key === 'Esc' || key === 27) {
 			if (isOpen(menu)) {
 				hide(menu);
 			}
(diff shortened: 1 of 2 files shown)


JakeChampion:
This can be tested by the following steps:
- `npm run build -w 'components/o-header' && npm start -w 'components/o-header'`
- visit http://localhost:5000/core-mega-menu
- hover the second link with the mouse and wait for the menu to appear
- press the escape key and the menu should close

components/o-header/src/js/mega.js
```diff
@@ -23,6 +23,17 @@ function addEvents (parent, menu) {
 		}, INTENT_ENTER);
 	});
 
+
+	const handleKeydown = (e) => {
+		if (e.keyCode === 27) {
```
notlee, line 28: Non-blocking suggestion: Looks like `.keyCode` is deprecated. Let's use [.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) with a temporary `.keyCode` fallback for IE?

notlee (approved):

components/o-header/src/js/mega.js
JakeChampion: Thanks - I've implemented this suggestion in [`b08fcae` (#569)](https://github.com/Financial-Times/origami/pull/569/commits/b08fcaee5c8ffcaa874de6a78be541de23ccc73c)

notlee (approved):

PR #571, opened 2022-01-18
o-tabs: implement all the required features for a tab component which implements the wai-aria tabs design pattern

### LINKED ISSUE #559: o-tabs: focus is taken off the website when activating a tab - ID: DAC_Name_Role_Value_Focus_Order_01
# Issue
This is from page 197 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-tabs@6.0.0&demo=buttontabs&system_code=origami-repo-data&brand=master>

The tabs have been structured incorrectly with links being placed inside list items which have the role of tab. This issue may affect how screen reader users identify the tabs in the panel.
When users press enter to select a tab, the focus is taken of the page into the browser. This issue may affect keyboard alone and screen reader users. Keyboard alone users would have to tab back through the browser to get back to the tabs, whilst screen reader users may become disorientated and be unable to distinguish where the focus has been placed.

# Screen Reader user comments:

> “Although I could focus on and activate each of the tabs with the Screen Reader, using ‘F’ to jump to each of the tabs followed by ‘Enter’, I was not informed which tab was selected. This was confusing and also time consuming and may also confuse other Screen Reader users because extra steps are needed to verify which tab has been opened.
> Using NVDA I found that I could identify which tab had been chosen despite not being informed that a selection had been made.
> Using TalkBack and VoiceOver, I found that I could determine which tab had been selected when swiping in context.
> Ensuring that the Screen Reader informs the user which tab is selected at the point of activation should resolve confusion and make navigation swifter.
> Please note that this issue was partially present with NVDA, TalkBack and VoiceOver.”

> After I have activated the tab, my focus goes back into the main address bar, and I have to tab back into the page to select a different one. if this were a real content page users would not expect that to happen, they would expect their focus to continue where it was from otherwise, they would have to start from the top of the page again.

# Additional instances:
Additional instances of this issue exist on other pages throughout the website; wherever this issue occurs, they too will need to be resolved.
- URL: https://www.ft.com/__origami/service/build/v3/demo?component=o- tabs%406.0.0&demo=buttontabs-inverse&system_code=origami-repo-data&brand=master
   
- URL: https://www.ft.com/__origami/service/build/v3/demo?component=o- tabs%406.0.0&demo=buttontabs-big&system_code=origami-repo-data&brand=master

# Proposed Solution:
Change the links inside the list to buttons, this will enable screen readers to navigate and identify the tabs correctly.
Please refer to [ARIA Tab Panel Accessibility: A11y Support Series](https://www.deque.com/blog/a11y-support-series-part-1-aria-tab-panel-accessibility/), [Tabs - GOV.UK Design System](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html) and [Example of Tabs with Manual Activation - W3.org](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-2/tabs.html)

JakeChampion:
This is an o-tabs implementation of [WAI-ARIA tab-panels](https://www.w3.org/TR/wai-aria-practices/#tabpanel)

Resolves #559 

The changes in this component include:

- added keyboard support for moving between tabs using arrow keys
- when navigating the tab-list or selecting a tab, keep focus within the tab-list and do not move focus to the shown tab-panel
- add aria-controls relationships between tabs and their associated tab-panel
- added support to activate a tab by pressing the space-bar

### DIFF
--- a/components/o-tabs/MIGRATION.md
+++ b/components/o-tabs/MIGRATION.md
@@ -4,6 +4,8 @@
 
 `disableFocus` has been removed as it is now the default behaviour and there is no way to change the behavior. This was changed because during an accessibility audit it was explained that focus should stay within the tab-list and not move away when a tab is activated.
 
+The method `getTabPanelEls` has now been marked as "private" as it is only meant to be used internally within o-tabs. If you need to use this method, please raise an issue on GitHub or Slack.
+
 ## Migrating from v6 to v7
 
 The html for o-tabs was changed to bring support for assistive technologies.
--- a/components/o-tabs/MIGRATION.md
+++ b/components/o-tabs/MIGRATION.md
@@ -6,6 +6,8 @@
 
 The method `getTabPanelEls` has now been marked as "private" as it is only meant to be used internally within o-tabs. If you need to use this method, please raise an issue on GitHub or Slack.
 
+The method `keyPressHandler` has now been renamed to `keyUpHandler`.
+
 ## Migrating from v6 to v7
 
 The html for o-tabs was changed to bring support for assistive technologies.
--- a/components/o-tabs/MIGRATION.md
+++ b/components/o-tabs/MIGRATION.md
@@ -4,7 +4,7 @@
 
 `disableFocus` has been removed as it is now the default behaviour and there is no way to change the behavior. This was changed because during an accessibility audit it was explained that focus should stay within the tab-list and not move away when a tab is activated.
 
-The method `getTabPanelEls` has now been marked as "private" as it is only meant to be used internally within o-tabs. If you need to use this method, please raise an issue on GitHub or Slack.
+The method `getTabPanelEls` has now been renamed to `_getTabPanelEls` and marked as "private" as it is only meant to be used internally within o-tabs. If you need to use this method, please raise an issue on GitHub or Slack.
 
 The method `keyPressHandler` has now been renamed to `keyUpHandler`.
 
--- a/components/o-tabs/MIGRATION.md
+++ b/components/o-tabs/MIGRATION.md
@@ -10,11 +10,11 @@ The component no longer uses `ul` and `li` elements and instead uses `div` and `
 -<ul data-o-component="o-tabs" class="o-tabs" role="tablist" data-o-tabs-update-url>
 +<div data-o-component="o-tabs" class="o-tabs" role="tablist" data-o-tabs-update-url>
 -	<li role="tab"><a href="#tabContent1">Tab 1</a></li>
-+	<a role="tab" aria-controls="#tabContent1" href="#tabContent1">Tab 1</a>
++	<a role="tab" aria-controls="tabContent1" href="#tabContent1">Tab 1</a>
 -	<li role="tab"><a href="#tabContent2">Tab 2</a></li>
-+	<a role="tab" aria-controls="#tabContent2" href="#tabContent2">Tab 2</a>
++	<a role="tab" aria-controls="tabContent2" href="#tabContent2">Tab 2</a>
 -	<li role="tab"><a href="#tabContent3">Tab 3</a></li>
-+	<a role="tab" aria-controls="#tabContent3" href="#tabContent3">Tab 3</a>
++	<a role="tab" aria-controls="tabContent3" href="#tabContent3">Tab 3</a>
 -</ul>
 +</div>
 ```
--- a/components/o-tabs/MIGRATION.md
+++ b/components/o-tabs/MIGRATION.md
@@ -10,11 +10,11 @@ The component no longer uses `ul` and `li` elements and instead uses `div` and `
 -<ul data-o-component="o-tabs" class="o-tabs" role="tablist" data-o-tabs-update-url>
 +<div data-o-component="o-tabs" class="o-tabs" role="tablist" data-o-tabs-update-url>
 -	<li role="tab"><a href="#tabContent1">Tab 1</a></li>
-+	<a role="tab" href="#tabContent1">Tab 1</a>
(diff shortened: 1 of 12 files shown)


components/o-tabs/README.md
```diff
@@ -25,11 +26,11 @@ The _tabpanel_ elements must have a `o-tabs__tabpanel` class added to them.
...
-<ul data-o-component="o-tabs" class="o-tabs" role="tablist">
-	<li role="tab"><a href="#tabContent1">Tab 1</a></li>
-	<li role="tab"><a href="#tabContent2">Tab 2</a></li>
-	<li role="tab"><a href="#tabContent3">Tab 3</a></li>
-</ul>
+<div data-o-component="o-tabs" class="o-tabs" role="tablist">
+	<a role="tab" aria-controls="tabContent1" href="#tabContent1" >Tab 1</a>
+	<a role="tab" aria-controls="tabContent2" href="#tabContent2">Tab 2</a>
+	<a role="tab" aria-controls="tabContent3" href="#tabContent3">Tab 3</a>
+</div>
```
JakeChampion, line 33: The markup hasn't changed in this pull-request, this was some out-of-date documentation that I chose to update at the same time

components/o-tabs/README.md
```diff
@@ -202,12 +198,25 @@ This table outlines some of the possible button themes you can request in the [`
...
+Space or Enter | Activates the tabpanel which is associated with the focused tab if it was not already activated.
+
+
 ## Migration Guide
 
 | State | Major Version | Last Minor Release | Migration guide |
 | :---: | :---: | :---: | :---: |
-| ✨ active | 6 | N/A | [migrate to v6](MIGRATION.md#migrating-from-v5-to-v6) |
-| ✨ active | 5 | 5.0 | [migrate to v5](MIGRATION.md#migrating-from-v4-to-v5) |
+| ✨ active | 6 | N/A | [migrate to v7](MIGRATION.md#migrating-from-v6-to-v7) |
```
JakeChampion, line 217:
```suggestion
| ✨ active | 7 | N/A | [migrate to v7](MIGRATION.md#migrating-from-v6-to-v7) |
```

components/o-tabs/src/js/Tabs.js
```diff
@@ -16,7 +16,7 @@ class Tabs {
 		this.boundClickHandler = this.clickHandler.bind(this);
 		this.rootEl.addEventListener('click', this.boundClickHandler, false);
 		this.boundKeyPressHandler = this.keyPressHandler.bind(this);
-		this.rootEl.addEventListener('keypress', this.boundKeyPressHandler, false);
+		this.rootEl.addEventListener('keyup', this.boundKeyPressHandler, false);
```
JakeChampion, line 19: This needed to change because when pressing left or right arrow keys, the active tab should only change once

components/o-tabs/MIGRATION.md
```diff
@@ -10,11 +10,11 @@ The component no longer uses `ul` and `li` elements and instead uses `div` and `
 -<ul data-o-component="o-tabs" class="o-tabs" role="tablist" data-o-tabs-update-url>
```
chee, line 10: Should we mention in the migration guide that `disablefocus` was removed?

components/o-tabs/src/js/Tabs.js
chee: should we rename this function from `KeyPress`?

components/o-tabs/src/js/Tabs.js
```diff
@@ -59,17 +59,16 @@ class Tabs {
...
 			if (targetEl) {
-				tab.setAttribute('aria-controls', tabTargetId);
 				tab.setAttribute('tabindex', '0');
-
 				const label = tab;
 				const labelId = tabTargetId + '-label';
-				label.setAttribute('tabindex', '-1');
 				label.id = labelId;
 				targetEl.setAttribute('aria-labelledby', labelId);
 				targetEl.setAttribute('role', 'tabpanel');
```
notlee, line 67: I would expect a public function named `getTabPanelEls` to return table panel elements, but this is doing more and is setting tabs up for the first time – called once in the constructor – and would have unexpected consequences if called a second time. For example by a dependent project. 

Since this is a breaking change would now be a good time to refactor and make sure the public api is sensible? Or we might consider making method private to allow us to refactor more easily at a later date

JakeChampion: I marked it as internal and then we can refactor it later 👍

components/o-tabs/MIGRATION.md
JakeChampion: Definitely, I've now added a migration guide for the new version 👍

components/o-tabs/src/js/Tabs.js
JakeChampion: I've renamed this now 👍

components/o-tabs/src/js/Tabs.js
notlee: thanks 👍 worth prepending with an `_` too? I think that helps make it more obvious when it has been used in a project despite being considered private. Don't forget to ass to the migration guide too! :)

notlee (approved):

components/o-tabs/src/js/Tabs.js
JakeChampion: Good point, I've done that in [`41b3eea` (#571)](https://github.com/Financial-Times/origami/pull/571/commits/41b3eeaa360664a7e44ca06a5ae1e2c4e2d85473)

notlee (approved):

Issue #501, opened 2022-01-06
o-banner: Improve label for the close button - ID: DAC_Headings_And_Labels_01


JakeChampion:
# Issue
This is from page 21 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-banner%404.0.0&demo=banner&system_code=origami-repo-data&brand=master>

The ‘Close’ button may be ambiguous for users who rely on audio feedback such as screen reader users. Users may not be aware what will be closed when the button is selected.

# Screen reader user comments:  

> I found a ‘Close’ button which when viewed out of context either with the JAWS Forms dialogue (Insert + F5) or the ‘B’ keystroke.
> When viewed in this way, the ‘Close’ button does not describe to what it refers.
> Including clear and descriptive button labels should help to avoid confusion for Screen Reader users when navigating in and out of context.

 
# Proposed Solution:
Add more description to the aria-label which will enable screen reader users to identify the purpose of the button.

Example:
```html
 <button class=”o-banner__close” aria-label=”Close home page banner” title=”Close”></button>
```

JakeChampion:
This issue is also present on <https://www.ft.com/__origami/service/build/v3/demo?component=o- banner%404.0.0&demo=small&system_code=origami-repo-data&brand=master> and found on page 28 of the audit report

JakeChampion:
This issue is also present on <https://www.ft.com/__origami/service/build/v3/demo?component=o-banner%404.0.0&demo=marketing&system_code=origami-repo-data&brand=master> and found on page 31 of the audit report

JakeChampion:
This issue is also present on <https://www.ft.com/__origami/service/build/v3/demo?component=o-banner%404.0.0&demo=product&system_code=origami-repo-data&brand=master> and found on page 33 of the audit report

PR #578, opened 2022-01-21
o-banner: improve the label description and association of the close button with the banner

JakeChampion:
I've updated the default close button label to say "close banner" rather than only "close", ideally products which user o-banner will be passing in a custom label for the close button which describes the banner in some more detail. E.G. "Close the limited time offer banner."

I also added an controller association between the close button and the banner (if the banner had an ID), this should improve the experience for user even more as they should hopefully be able to see what element the close button will close.
This association was done via `aria-controls`, which is a way to associate which element(s) are controlled by the current element.
Some screen-readers, such as Jaws will read the aria-controls and annouce it to the user
I.E. Jaws will announce "Use the Jaws key Alt and m to move to the controlled element"
This is especially useful when the user has found the control via the "JAWS Forms dialogue" which will find the control usually out of it's context within the page.

This should resolve https://github.com/Financial-Times/origami/issues/501


components/o-banner/src/js/banner.js
```diff
@@ -154,6 +154,7 @@ class Banner {
 	buildBannerElement (bannerElement) {
 		bannerElement = bannerElement || document.createElement('div');
 		bannerElement.innerHTML = '';
+		bannerElement.setAttribute('role', 'banner');
```
JakeChampion, line 157: This change might be a bad idea because the [specification](https://www.w3.org/TR/wai-aria-1.2/#banner) states that their should only be one element with the banner role on a page

notlee:
hey, haven't had time to look at this properly but whilst it's in my head: we might need to make some changes to the Storybook Story too

JakeChampion:
I've also updated the tests to use chai assert instead of proclaim. By moving to chai assert we can also use `assert.dom.equal` which is a really nice way to compare whether the DOM contains what we expect it to contain

components/o-banner/src/js/banner.js
notlee: >Site-oriented content typically includes things such as the logo or identity of the site sponsor, and a site-specific search tool. 

Reading the spec it sounds to me like the banner role is intended to identify the site header.

[MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/banner_role) makes me think that more so.

>While it is best to use the header element and ensure it is not a descendant of any subsection of the page, sometimes you don't have access to the underlying HTML. If this is the case, you can add the role of banner to the main header of the page with JavaScript. Identifying the page's banner in this way will help improve the site's accessibility.

Sounds to me like role banner would be more suited to o-header. I see on ft.com that Chrome is giving the header a role of banner but it looks like Firefox is giving it a landmark role.

Would a [role="landmark"](https://www.w3.org/TR/wai-aria-1.2/#landmark) be more appropriate for o-banner?

components/o-banner/test/banner.test.js
```diff
@@ -425,12 +424,12 @@ describe('Banner', () => {
...
 
 			it('returns an HTML element', () => {
-				proclaim.isInstanceOf(returnValue, HTMLElement);
+				assert.instanceOf(returnValue, HTMLElement);
 			});
 
 			it('constructs the element HTML based on the given options', () => {
-				proclaim.strictEqual(returnValue.outerHTML.replace(/[\t\n]+/g, ''), `
-					<div class="o-banner">
+				assert.dom.equal(returnValue, `
```
notlee, line 431: nice!

components/o-banner/demos/src/banner.mustache
```diff
@@ -1,5 +1,5 @@
 
-<div class="o-banner" data-o-component="o-banner">
+<div class="o-banner" data-o-component="o-banner" id="banner">
```
notlee, line 2: Could we make this id more origami or demo specific to defend against copy-paste mistakes? An element with id of "banner" is generic enough it could already exist on a users page.

components/o-banner/src/js/banner.js
JakeChampion: > Sounds to me like role banner would be more suited to o-header. I see on ft.com that Chrome is giving the header a role of banner but it looks like Firefox is giving it a landmark role.
One of those browsers must not be following the description from the [W3 HTML Sectioning Elements: ARIA Landsmarks Example](https://www.w3.org/TR/wai-aria-practices/examples/landmarks/HTML5.html)
A `header` element has a default role of `banner` when in context of the `body` element. The `header` element is not a `banner` landmark when it is a descendant of the following HTML sectioning elements:                                            - `article`
- `aside`
- `main`
- `nav`
- `section`

I'll ask the auditors if they think a role of `banner` is appropriate here 👍

JakeChampion: > Would a [role="landmark"](https://www.w3.org/TR/wai-aria-1.2/#landmark) be more appropriate for o-banner?

Yes, this is the role to use ☺️

components/o-banner/demos/src/banner.mustache
JakeChampion: done in [`b9d4673` (#578)](https://github.com/Financial-Times/origami/pull/578/commits/b9d46734c31de5146db91384315674995cc1e647)

components/o-banner/src/js/banner.js
JakeChampion: done in [`2d539db` (#578)](https://github.com/Financial-Times/origami/pull/578/commits/2d539db713e584517c944b540c4604b2f11d6fba)

notlee (approved):

notlee:
🎉

PR #581, opened 2022-01-24
fix: o-grid, retain existing max grid width when adding larger layout

notlee:
`oGridAddLayout` used to error when adding a new larger layout.
Since that has been fixed we discovered it also sets the max
grid width to the new larger layout. A user doesn't necessarily
want this, for example when the new breakpoint is used to
target background imagery for large screens.

### DIFF
--- a/components/o-grid/src/scss/_mixins.scss
+++ b/components/o-grid/src/scss/_mixins.scss
@@ -56,8 +56,6 @@
 	@if $gutter-width {
 		$o-grid-gutters: $temp-gutters !global;
 	}
-
-	$_o-grid-max-width: map-get($o-grid-layouts, nth($_o-grid-layout-names, -1)) !global;
 }
 
 /// Apply styles at a given layout size (breakpoint) as the grid. It should be


notlee:
(we can add some other api to update the max width when/if needed in the future)

components/o-grid/src/scss/_mixins.scss
```diff
@@ -56,8 +56,6 @@
 	@if $gutter-width {
 		$o-grid-gutters: $temp-gutters !global;
 	}
-
-	$_o-grid-max-width: map-get($o-grid-layouts, nth($_o-grid-layout-names, -1)) !global;
```
JakeChampion, line 60: Do we need to do something to this line as well? https://github.com/Financial-Times/origami/blob/o-grid-add-layout-max-width/components/o-grid/src/scss/_mixins.scss#L168

notlee: no, we still want the max width to be set using the default value of `$_o-grid-max-width`

JakeChampion (approved):

PR #586, opened 2022-01-26
fix: add the region role to the demos to help show which role is applicable to a banner

JakeChampion:


notlee (approved):

components/o-banner/demos/src/compact.mustache
```diff
@@ -1,11 +1,11 @@
 
-<div class="o-banner o-banner--compact" data-o-component="o-banner" id="o-banner-demo">
+<div role="region" class="o-banner o-banner--compact" data-o-component="o-banner" id="o-banner-demo">
 	<div class="o-banner__outer">
 		<div class="o-banner__inner" data-o-banner-inner="">
 			<div class="o-banner__content">
-				<header class="o-banner__heading">
+				<div class="o-banner__heading">
```
JakeChampion, line 6: Now that we are applying a landmark role (`role="region"`), we can no longer use elements such as `header` within the component because header elements have a landmark role and it is advised to not nest landmark roles -- https://dequeuniversity.com/rules/axe/4.3/landmark-banner-is-top-level

PR #593, opened 2022-01-27
move component stories into a subdir and rename to component name

chee:
i think this is probably more clear, and then later we can add
a nice Guidance directory, maybe a Design directory and an Accessibility directory?

whatever the future holds.

before:

looks like this now:


JakeChampion (approved):

JakeChampion:
Really nice idea 👍

PR #596, opened 2022-01-28
backstage: Update component `.npmignore`

notlee:
- Ignore all JSX templates (for now) and Stories
- Ignore .prettierrc.toml, .stylelintignore, .eslintignore
- Add a trailing slash to directories we want to ignore,
to be explicit about ignoring a directory and not a file.
- Don't ignore anything that isn't there or hasn't been there
in a long time. Except `.env` because that could be bad.
- Don't explicitly ignore anything npm@7 will ignore anyway.
Note npm@8 docs say `.npmignore` and `.gitignore` are ignored
by default but the npm@7 docs don't mention them so let's play
it safe:
https://docs.npmjs.com/cli/v7/using-npm/developers#keeping-files-out-of-your-package


notlee:
Maybe we should reverse this about be explicit about what _should_ be published. We should double check each component though and maybe consider it a breaking and who has time for that right now. This is really just to ignore the jsx :)

chee (approved):

PR #599, opened 2022-01-31
backstage: fix website layout

notlee:


apps/website/_config.yml
```diff
@@ -28,7 +28,7 @@ defaults:
   - scope:
       path: ""
     values:
-      layout: o-layout-documentation
+      layout: o-layout-docs
```
JakeChampion, line 31: oops 🙈  thanks for fixing this

PR #613, opened 2022-02-07
o-banner: include a label on the banner demos to make the banners appear in the navigation tree

JakeChampion:
The navigation tree can be used as a way to navigate to sections of a page without having to tab through all the content

These banners will be added to the navigation tree because they are assgined a role of 'region' and now include a label


components/o-banner/demos/src/compact.mustache
```diff
@@ -1,8 +1,8 @@
 
-<div role="region" class="o-banner o-banner--compact" data-o-component="o-banner" id="o-banner-demo">
+<div role="region" class="o-banner o-banner--compact" data-o-component="o-banner" id="o-banner-demo" aria-labelledby="o-banner-demo-title">
 	<div class="o-banner__outer">
 		<div class="o-banner__inner" data-o-banner-inner="">
-			<div class="o-banner__content">
+			<div class="o-banner__content" id="o-banner-demo-title">
 				<div class="o-banner__heading">
 					<h1>FT Compact</h1>
```
notlee, line 7: Would the heading be more appropriate for a label? The content could be quite long

notlee:
@JakeChampion could you also make these improvements to the Story template? There might be other things we forgot to port including the [role](https://github.com/Financial-Times/origami/blob/894f62a7799302d43dcafcd325b57041101d5d87/components/o-banner/src/tsx/banner.tsx#L65)  and the [close button label](https://github.com/Financial-Times/origami/blob/894f62a7799302d43dcafcd325b57041101d5d87/components/o-banner/stories/banner.stories.tsx#L94) 😬

JakeChampion:
I didn't mean to close this...

components/o-banner/demos/src/compact.mustache
JakeChampion: I've opted to go with a custom aria-label instead so that we can supply different text for the navigation tree description

JakeChampion:
@notlee I've updated the stories now 👍

components/o-banner/demos/src/compact.mustache
notlee: Nice, should we make it required in the jsx template or is there a case where we wouldn't want it to show in the navigation tree?

PR #615, opened 2022-02-07
Stories o buttons pagination

notlee:


notlee:
Storybook is showing controls from ButtonProps for some reason. I don't think it likes [Pick](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) 😢 Many ideas? We might have to redefine this interface without `Pick`. Pick is nice because we can match type/size/theme options defined in `ButtonProps` without manually keeping them in sync

```
export type ButtonPaginationProps = Pick<ButtonProps, "type" | "size" | "theme"> & {
  previousPager: ButtonPaginationPager,
  pages: ButtonPaginationItem[]
  nextPager: ButtonPaginationPager,
};
```
 
Screenshot 2022-02-07 at 16 05 29

chee:
Hmm! strange, Pick should work: https://github.com/styleguidist/react-docgen-typescript/blob/a33dfbbebf7d9ca997aca7d110fe865fd9be7a08/src/__tests__/data/ColumnWithPick.tsx

notlee:
Oh https://github.com/Financial-Times/origami/pull/615/files#diff-42311ce2571509da9dc44637442ac35b81754502a54413e8d7e917113374322fL11

notlee:
I take back that Oh - nope not sure

notlee:
Sorted.

components/o-buttons/src/tsx/pagination.tsx
```diff
@@ -0,0 +1,115 @@
...
+      pageAttributes['aria-current'] = 'page';
+    }
+    if(page.current && !page.href) {
+      pageAttributes['aria-selected'] = true;
+    }
+    if(page.onClick) {
+      pageAttributes['onClick'] = page.onClick;
+    }
+
+    return <PageTag key={nanoid()} href={page.href} label={page.number.toString()} attributes={pageAttributes} theme={theme} size={size} type={type}></PageTag>
```
JakeChampion, line 98: @notlee The key will be a different value on each render, a value which is unique per page and stable between renders is the `page` object. Can we try using that instead?

```suggestion
    return <PageTag key={page} href={page.href} label={page.number.toString()} attributes={pageAttributes} theme={theme} size={size} type={type}></PageTag>
```

components/o-buttons/src/tsx/pagination.tsx
```diff
@@ -0,0 +1,115 @@
...
+  }));
+
+    return (
+        <div className="o-buttons-pagination">
+          <PreviousTag attributes={firstPageIsSelected ? {disabled: true} : {onClick: previousPager.onClick}} href={previousPager.href} label={previousPager.label} icon={'arrow-left'} iconOnly={true} theme={theme} size={size} type={type}></PreviousTag>
+          {pageElementsInGroups.flatMap((pageElementGroup, pageGroupIndex) => {
+						const elementGroup = [];
+						if(pageGroupIndex > 0) {
+							elementGroup.push(<Ellipsis key={nanoid()}></Ellipsis>);
```
JakeChampion, line 107: Same here, I think we can use pageElementGroup which is stable between renders instead of `nanoid()` which I think generates a new value each render -- 
```suggestion
							elementGroup.push(<Ellipsis key={pageElementGroup}></Ellipsis>);
```

components/o-buttons/src/tsx/pagination.tsx
notlee: 👍 I've gone with `page.number` which should be unique and stable

components/o-buttons/stories/button-group.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+	args: {},
+	parameters: {
+		design: {},
+		guidelines: {},
+		html: {},
+	},
+};
+
+const ButtonGroupStory = args => <ButtonGroup>
+	{args.buttons.map(buttonProps => <Button {...buttonProps} key={nanoid()}/>)}
```
JakeChampion, line 21: Are we able to remove this usage of nanoid? I believe the key value will be different on every render which is not what is expected, the key is meant to be consistent between renders

notlee: That's right. I'm struggling for an appropriate key here though, label+type would probably do but isn't unique really

JakeChampion: The buttonProps object is unique with regards to each button and is also consistent between renders, let's use that?

```suggestion
	{args.buttons.map(buttonProps => <Button {...buttonProps} key={buttonProps}/>)}
```

notlee: Ooops I didn't push the last commit. Using the props is what I've done but as a string. It's not guaranteed to be unique because it's possible to add a button with the same props using the stories controls. It'll do for our purpose, but doesn't seem ideal

JakeChampion: The `buttonProps` object looks to be the ideal as it is different for each button and is consistent between renders. Even if each button has the same label and type, their object would be different

JakeChampion (approved):

PR #619, opened 2022-02-07
add o-typography Stories

notlee:


components/o-typography/stories/typography-list.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+		guidelines: {
+			notion: 'f82c8bd1ef1f4f649a61d17d5d49750c',
+		},
+		html: {},
+	},
+};
+
+const items = ['Apples', 'Bananas', 'Oranges', 'Pears', 'Tangelos'];
+const Story = args => <Body>
+    <List {...args} children={args.items.map(item => <ListItem key={nanoid()}>{item}</ListItem>)} />
```
chee, line 26: doesn't calling nanoid() like this mean there is a different key every time it rerenders?

JakeChampion: I think so!

Could we use the array index as the key?

chee: `key={item}` should be fine here

notlee: Oops, yes! I've removed that

JakeChampion (approved):

Issue #554, opened 2022-01-11
remove use of either aria-label or title on elements which have both - ID: DAC_Close_Button_01


JakeChampion:
# Issue
This is from page 180 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-message@5.0.0&demo=alert-inner-success&system_code=origami-repo-data&brand=master>

The close button may be ambiguous for users who rely on audio feedback such as screen reader users. Users may not be aware what will be closed when the button is selected.
Additionally, the close button has been given both an aria-label and a title. This means that the close label is read out twice to screen reader users. Please note: This issue is present in other areas of the service.

Additional instances of this issue exist on other pages throughout the website; wherever this issue occurs, they too will need to be resolved.

# Proposed Solution:
Add more description to the aria-label which will enable screen reader users to identify the purpose of the button. Remove either the aria-label or the title attribute. This will enable the label to be read out just once for screen reader users.

## Example:
 ```html
<button class=”o-banner__close” aria-label=”Close dialog box”></button>
```

JakeChampion:
fixd via https://github.com/Financial-Times/origami/pull/621

PR #621, opened 2022-02-08
o-banner: do not repeat the description for the banner close button

JakeChampion:
If an element has both a title attribute and an aria-label attribute, then both will be read out by screen-readers

This change removes the aria-label attribute and keeps the title attribute.

This resolves issue https://github.com/Financial-Times/origami/issues/554

### DIFF
--- a/components/o-banner/src/js/banner.js
+++ b/components/o-banner/src/js/banner.js
@@ -227,7 +227,6 @@ class Banner {
 		const closeButton = document.createElement('button');
 		closeButton.className = classNames.close;
 		closeButton.setAttribute('aria-label', this.options.closeButtonLabel);
-		closeButton.setAttribute('title', this.options.closeButtonLabel);
 		if (this.bannerElement.id) {
 			closeButton.setAttribute('aria-controls', this.bannerElement.id);
 		}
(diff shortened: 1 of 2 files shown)


notlee:
Are you confident `title` is a reliable choice over `aria-label`? Support by screen readers seems better than I expected but I noticed this:

>Results with JAWS 2020 / JAWS 2019 [...] The only failure noted is with Microsoft Edge: JAWS simply fails to read the title attribute for all the four links.

https://www.deque.com/blog/text-links-practices-screen-readers/

Removing `title` for `aria-label` might be a more reliable way to provide essential information? It would be a shame to lose the `title` popup on mouse over but that is the experience keyboard and mobile users get, who can't benefit from the popup – we could talk to design about preferring visible close labels.

chee (approved):
good

notlee (changes requested):
☝️ see above query

JakeChampion:
> Are you confident `title` is a reliable choice over `aria-label`? Support by screen readers seems better than I expected but I noticed this:
> 
> > Results with JAWS 2020 / JAWS 2019 [...] The only failure noted is with Microsoft Edge: JAWS simply fails to read the title attribute for all the four links.
> 
> [deque.com/blog/text-links-practices-screen-readers](https://www.deque.com/blog/text-links-practices-screen-readers/)
> 
> Removing `title` for `aria-label` might be a more reliable way to provide essential information? It would be a shame to lose the `title` popup on mouse over but that is the experience keyboard and mobile users get, who can't benefit from the popup – we could talk to design about preferring visible close labels.

I'll confirm with a manual test in Edge with Jaws

notlee (approved):

Issue #516, opened 2022-01-07
o-footer-services: Improve link descriptions for use with screen readers - ID: DAC_Information_Displayed_Using_CSS_03


JakeChampion:
# Issue
This is from page 73 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-footer-services@4.0.2&demo=dark-footer&system_code=origami-repo-data&brand=master>

The ‘like this one’ link contains a visual cue for sighted users to inform them that the link opens in a new window, but for screen reader users there is no auditory cue to inform users of this behaviour.

# Proposed Solution:
Add visually hidden text to the link to inform screen reader users that the link opens in a new window.
Example:
 <a class="o-footer-services__content--external" href="/somewhere">like this one</a>
 <a class="o-footer-services__content--external" href="/somewhere">like this one<span class="visually-hidden">(opens in a new window)</span></a>

notlee:
Done. We decided to remove this ambiguous icon. See: https://github.com/Financial-Times/origami/pull/636

PR #630, opened 2022-02-09
o-footer-services: include text for screen-readers to explain external links open in a new window

JakeChampion:
Currently we have a visual indication that the link will open in a new window but we do not supply any text to indicate this to users using screen-readers

This pull-request adds a textual description via an "after" pseudo-element which is visually hidden but available in the accessibility tree for screen-readers to read out.

An alternative solution to this would be to include the text description in a "span" element which has a class to visually hide the content. I went with the "after" pseudo-element approach as it involves zero markup changes and will apply to all external links which make use of the visual indicator via the "o-footer-services__content--external" class

This resolves https://github.com/Financial-Times/origami/issues/516

### DIFF
--- a/components/o-footer-services/src/scss/_mixins.scss
+++ b/components/o-footer-services/src/scss/_mixins.scss
@@ -142,6 +142,16 @@
 				$include-base-styles: false
 			);
 		}
+		.o-footer-services__content--external::after {
+			content: "(opens in a new window)";
+			clip: rect(0 0 0 0);
+			clip-path: inset(50%);
+			height: 1px;
+			overflow: hidden;
+			position: absolute;
+			white-space: nowrap;
+			width: 1px;
+		}
 	}
 
 	@if $brand-strip {


components/o-footer-services/src/scss/_mixins.scss
```diff
@@ -142,6 +142,16 @@
...
+		.o-footer-services__content--external::after {
+			content: "(opens in a new window)";
+			clip: rect(0 0 0 0);
+			clip-path: inset(50%);
+			height: 1px;
+			overflow: hidden;
+			position: absolute;
+			white-space: nowrap;
+			width: 1px;
+		  }
```
JakeChampion, line 154:
```suggestion
	}
```

components/o-footer-services/src/scss/_mixins.scss
```diff
@@ -142,6 +142,16 @@
 				$include-base-styles: false
 			);
 		}
+		.o-footer-services__content--external::after {
+			content: "(opens in a new window)";
```
notlee, line 146: Have you considered using `oTypographyLink` which includes the "opens in new window" pseudo element when the `external` argument is `true`?

https://registry.origami.ft.com/components/o-typography@7.3.0/sassdoc?brand=core#mixin-otypographylink

notlee: Oh wait! that is being used 🤔

JakeChampion: That's used on line 141 but does not include the pseudo-element, maybe it's buggy?

notlee: It only applies when `&[target="_blank"]`

https://github.com/Financial-Times/origami/blob/7697c7844aae401ea2f463d17e5be58764f0f2d1/components/o-typography/src/scss/use-cases/_general.scss#L189

JakeChampion: Hmm, should the oTypographyLink mixin only add the icon if the target=_blank?

notlee: No because that icon indicates an external link, like a link from ft.com to google.com, not that the link opens in a new window

JakeChampion: > No because that icon indicates an external link, like a link from ft.com to google.com, not that the link opens in a new window

When our audit was done, the icon was interpreted as meaning it will open in a new window/tab. It seems the icon has an ambiguous interpretation 😔

JakeChampion: I've asked DAC for input on this --> https://financialtimes.slack.com/archives/C02NW8G3VPD/p1644420773095199

notlee: I guess often we'd choose to open in a new tab for an external link. But I imagine if we only apply the icon given `[target="_blank"]` we'll probably strip the icon unintentionally from many ft.com outbound links

notlee: Here's a chat we had a few years ago about new tab / external links:
https://financialtimes.slack.com/archives/C8U9J5YJ2/p1586866165077900

Maybe a question for design to research patterns out there, I can definitely see why it would be confused. For now maybe clarifying with DAC that it doesn't open in a new tab would be sufficient?

JakeChampion:
This is already done in o-typography for links which open in a new tab as mentioned by Lee in the review -- https://github.com/Financial-Times/origami/blob/7697c7844aae401ea2f463d17e5be58764f0f2d1/components/o-typography/src/scss/use-cases/_general.scss#L189

Issue #548, opened 2022-01-11
o-header: add missing focus state to the international edition link - ID: DAC_Focus_Visible_02


JakeChampion:
# Issue
This is from page 166 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-header@9.0.1&demo=header&system_code=origami-repo-data&brand=master>

When the side navigation menu is activated, the ‘Switch to International Edition’ link has no visible focus indicator. This issue may affect low-vision and keyboard alone users who may find difficulty in distinguishing when focus has been placed on the link

# Proposed Solution:
Ensure all focusable elements have a visible focus highlighting mechanism

PR #631, opened 2022-02-09
fix: show the entire focus ring for links within the drawer

JakeChampion:
previously only the top line of the focus ring was being shown

resolves https://github.com/Financial-Times/origami/issues/548

Without this change applied:
image

With this change applied:
image

### DIFF
--- a/components/o-header/src/scss/features/_drawer.scss
+++ b/components/o-header/src/scss/features/_drawer.scss
@@ -246,8 +246,11 @@
 		}
 	}
 
+	// These values exist to show the focus ring
+	// We need to keep them at-least equal to the outline width in o-normalise
 	.o-header__drawer-menu-item {
-		margin-top: 1px;
+		margin-top: 2px;
+		margin-bottom: 2px;
 	}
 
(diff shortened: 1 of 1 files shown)


components/o-header/src/scss/features/_drawer.scss
```diff
@@ -246,9 +246,6 @@
 		}
 	}
 
-	.o-header__drawer-menu-item {
-		margin-top: 1px;
```
notlee, line 250: I think this still _might_ be needed assuming the 1px space between items was also intended to visually separate the dropdown buttons

notlee: before
before

after
after

JakeChampion: I reverted that change back 👍

notlee: Doesn't look like the drawer is captured in the UI Kit so hard to know the intent

notlee (approved):

PR #637, opened 2022-02-11
breaking: provide text which explains a link "opens in a new window or tab" in a real element and not a pseudo-element

JakeChampion:
This is a breaking change in o-typography and is why this is a draft pull-request:

This pull-request also removes the pseudo-element we were using to supply "opens in a new window or tab" and instead websites will need to provide this text in a real element and not a pseudo-element. This was advised by the auditing company (DAC) as they said not all screen-readers support reading pseudo-elements such as versions of older versions of JAWS, and also Talkback reads out the psuedo-element twice.

A community page which records whether pseudo-elements work with different screen-reader/browser combinations is [a11ysupport.io/tech/css/generated_content](https://a11ysupport.io/tech/css/generated_content)

I've tested JAWS 2021 and 2019 and they read out the psuedo-element, I don't have access to older versions of JAWS. I've also tested Narrator on Edge and that reads out the pseudo-element (I've made a pull-request on a11ysupport to update their test and table) - I've not yet tested TalkBack.


JakeChampion:
closing this as we are going with the pseudo-element approach unless someone can show us which screen reader and browser combination does not support it

PR #639, opened 2022-02-11
add an ft concept button

### LINKED ISSUE #583: Add myFT follow button to Origami
## Background

The [accounts team are working on a new Newsletters page](https://financialtimes.slack.com/archives/C02FU5ARJ/p1643025681036000) and require the style of the myFT follow button (aka lozenge, but let's not name after shape which might change and isn't even very diamond like).

These styles are [defined in n-myft-ui](https://github.com/Financial-Times/n-myft-ui/blob/bf2145fcfd211001dfab6a3a271dd72808d062bd/mixins/lozenge/main.scss#L52) and duplicated in [x-follow-button](https://github.com/Financial-Times/x-dash/blob/dc08ea8b028a6737b307b65e52f73c3b9be79b6d/components/x-follow-button/src/styles/mixins/lozenge/main.scss#L92). The accounts team feel they can use neither because:

- n-myft-ui comes with inconsistencies and technical debt (I guess that's why x-follow-button duplicates the styles)
- x-follow-button includes a template with logic which doesn't suit the accounts team's specific need. And it's not geared to allow other projects to depend on Sass mixins, as selectors may change without warning (markup changes are not breaking changes for x-dash).

Creating a myFT follow button in Origami with jsx template will allow teams to use the myFT follow button UI consistently without depending on n-myft-ui or being tied to certain business logic - e.g. the button could be used for different purposes to follow a concept, or sign up to a news letter, or potentially work with different backends, etc. It also means the myFT follow button will show in the Origami registry, versioned, with usual documentation / tools / workflow.

The Origami template should not include business logic. x-dash is a great project to continue to wrap Origami components with shared Customer Products business logic.

## Work Required

- Add a myFT follow button component to Origami, including a public jsx template (without business logic)
- Work with the platforms team to update `x-follow-button` to use the Origami jsx template

Questions: 
- is this part of o-buttons?
- Do all component jsx templates become public?

## Examples of the myFT follow button

myft page:
myft page

article page:
article page

stream page:
stream page

ft app:
ft app

## References

- Aside: See recent chat between accounts and content discovery team about myft architecture https://financialtimes.slack.com/archives/C3LRB6JCE/p1642764059153900

chee:
maybe this should be called toggle button, i guess it depends on the guidelines that are yet to be writ

closes #583


.github/workflows/percy-components-myft-follow-button.yml
```diff
@@ -0,0 +1,41 @@
...
+    - uses: actions/cache@v2
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+    - run: npm ci
+    - run: node ./scripts/percy.js
+      env:
+        PERCY_TOKEN: ${{ secrets.PERCY_TOKEN_COMPONENTS_MYFT_FOLLOW_BUTTON }}
```
JakeChampion, line 40: Note: we will need to make the percy project for this and add the secret to our github secrets. Extra note - we should probably document this in a section about creating a new component

chee: 100% true ✅ true ✅

chee: can we fail the build if `PERCY_TOKEN_COMPONENTS_MYFT_FOLLOW_BUTTON` is not defined? and explain in the error message why it will have failed? and then hope and pray that somebody would read the error message and believe it?

components/myft-concept-button/.gitignore
```diff
@@ -0,0 +1,9 @@
+.DS_Store
+.env
+.idea/
+/.sass-cache/
+/bower_components/
+/build/
+/demos/local
+/node_modules/
+npm-debug.log
```
notlee, line 9:
```suggestion
.env
.gitignore
.npmignore
.eslintrc.cjs
.remarkrc.cjs
.stylelintrc.cjs
.prettierrc.toml
.eslintignore
.stylelintignore
stories/
src/tsx/
.github/
test/
demos/local/

```

☝️ Brings inline with other components. We need to update this in the create new component workflow at some point. https://github.com/Financial-Times/origami/pull/596

Also means the tsx/stories won't be published. But maybe we want to make this the first component with published tsx?

components/myft-concept-button/main.scss
```diff
@@ -0,0 +1,228 @@
+@import '@financial-times/o-typography/main';
+
+$myft-lozenge-themes: (
```
notlee, line 3:
```suggestion
/// @access private
$_myft-lozenge-themes: (
```

Could we make this private with a leading `_`? Same for any variable/function/mixin that we don't expect the a user to call directly? It'll make it easier to refactor. For example easier to switch to design tokens without a breaking, in the case of the theme variable.

components/myft-concept-button/main.scss
```diff
@@ -0,0 +1,228 @@
...
+		disabled: rgba(oColorsByName('white'), 0.5),
+		focus-outline: oColorsByName('white'),
+	),
+);
+
+@function getThemeColor($key) {
+	@return map-get($theme-map, $key);
+}
+
+@mixin withTheme($theme) {
```
notlee, line 42: We should prepend public mixins/functions/variables with the component name as with the components, to avoid conflicts (until we switch to dart sass modules).

components/myft-concept-button/main.scss
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+}
+
+@mixin _base() {
+	@include oTypographySans($weight: 'semibold', $scale: -1);
+	border-radius: 100px;
+	box-sizing: content-box;
+	display: block;
+	font-size: 14px;
```
notlee, line 193: `$scale: -1` is 14px so this line is redundant?

components/myft-concept-button/main.scss
```diff
@@ -0,0 +1,228 @@
...
+	margin: 6px 4px 6px 2px;
```
notlee, line 194: It would be good opportunity to introduce `o-spacing` here. That will allow relative units to be output too, which the apps team may be able to start using

https://registry.origami.ft.com/components/o-spacing@3.2.1/sassdoc?brand=core#variable-o-spacing-relative-units

components/myft-concept-button/src/js/myft-concept-button.js
```diff
@@ -0,0 +1,87 @@
+export default class MyftConceptButton {
+	// TODO make a @typedef for options
```
notlee, line 2: Could we make the TODO an issue instead? It's easier to track and not forget

JakeChampion (approved):
👍 This looks good

components/myft-concept-button/src/js/myft-concept-button.js
```diff
@@ -0,0 +1,87 @@
...
+
+	get isFollowed() {
+		return this.button.ariaPressed;
+	}
+
+	/**
+	 * Toggle the button between followed and not.
+	 * Updates the UI and adds text to the a11y box.
+	 *
+	 * @param {boolean=} [setting] If this is true, we are going
```
notlee, line 26: We are going...to...? 😅

components/myft-concept-button/src/tsx/follow-button.tsx
```diff
@@ -0,0 +1,58 @@
...
+			className={`myft-concept-button myft-concept-button--follow ${
+				theme ? `myft-concept-button--${theme}` : ''
+			}`}
+			data-o-component="myft-concept-button"
+			data-myft-concept-button-concept={conceptName}
+			method="get"
+			{...extraFormProps}>
+			{csrf || null}
+			{
+				// TODO o-normalise-visually-hidden
```
notlee, line 40: What's this TODO for and is it something blocking release? If not it would be great to get into an issue instead

components/myft-concept-button/main.scss
notlee: Aside: The name doesn't begin with `o-`! Are we dropping `o-` as a convention? 

[drop the the](https://youtu.be/k5fJmkv02is?t=170)

notlee (commented):
An exciting addition!

components/myft-concept-button/.gitignore
chee: yeah, we will want to publish the tsx on this one for sure

components/myft-concept-button/main.scss
chee: well that is something i wondered about, and we should all discuss.

i somehow don't think of this as an Origami component because it is so specific to a Financial Times product.

Should it live inside `o-buttons`? in the design UI kit it is separate from `o-buttons` but of course that reflects the UI Library's history as a reflection of what we have (rather than it future of being the SOURCE of what we have).

if we made it an `o-` component it would be tempting to name it something more generic like `o-rounded-button-that-may-or-may-not-have-a-follow-state` but more succinct than that.

i felt like... it was a good idea to be honest and explicit that this is the `myft-concept-button` (which offers a follow-button and a concept-button template (the second of which i've just realized needs added to the TODO list)).

components/myft-concept-button/main.scss
chee: yes we should do that.

components/myft-concept-button/main.scss
chee: lol that sounds likely.

components/myft-concept-button/.gitignore
notlee: 👍 I've updated the suggestion to remove tsx. It still adds stories to the exclude and a couple of missing dot files

components/myft-concept-button/main.scss
chee: yep for sure. this margin is actually a real pity tbh, not sure it should have that. also, it doubles up because it's applied to both the form and the button. me from four years ago seems to have made some real psychedelic choices.

components/myft-concept-button/src/js/myft-concept-button.js
chee: oh this TODO is TODO before this PR is merged 😊

components/myft-concept-button/src/tsx/follow-button.tsx
chee: yeah this is TODO before release

components/myft-concept-button/stories/follow-button.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+		extraFormProps: {
+			onSubmit(event) {
+				event.preventDefault();
+			},
+		},
+	},
+	parameters: {
+		design: {
+			type: 'figma',
+			url: 'TODO point this at the myft button design',
```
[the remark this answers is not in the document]
chee: design tab showing 404
 😊

components/myft-concept-button/main.scss
notlee: - Definitely happy to drop the `o-`. I don't see the component name as a distinction between an Origami component or not an Origami component or more or less generic. If it's in the mono repo it's an Origami component and it should be used however the design guidelines / readme says! The only reasons I see for having the `o-` on any components are:
    - Chance to avoid conflicts with global CSS / Sass.
    - Communicates to engineers looking at some HTML that a component comes from Origami.
- It's probably more clear outside of o-buttons, since o-buttons is quite complex and branded, in my opinion.

notlee: Whatever the name is let's prepend its CSS/Sass with it to avoid conflicts

components/myft-concept-button/.gitignore
chee: wait. why do we want to ignore these in the gitignore??

```
.gitignore
.npmignore
.eslintrc.cjs
.remarkrc.cjs
.stylelintrc.cjs
.prettierrc.toml
.eslintignore
.stylelintignore
```

we use those files

components/myft-concept-button/README.md
```diff
@@ -0,0 +1,116 @@
...
+	<div class="myft-concept-button__announcement" aria-live="assertive"></div>
+	<button aria-pressed="false" class="myft-concept-button__button">
+		Sign up
+	</button>
+</div>
+```
+
+### MyFT follow button
+
+This one has a little plus or a tick icon. You'd be advised to include the `data-myft-concept-button-aria-live-pressed-text` and `data-myft-concept-button-aria-live-unpressed-text` options so screenreader users are informed that a change has occurred.
```
JakeChampion, line 42: 💯

components/myft-concept-button/README.md
```diff
@@ -0,0 +1,116 @@
...
+export interface ConceptButtonProps {
+	// button text
+	label: string;
+	// text to announce to screenreaders
+	ariaLiveText?: string;
+	// descriptive label for button
+	ariaLabel?: string;
+	// button theme
+	theme?: 'standard' | 'inverse' | 'opinion' | 'monochrome';
+	// button type (the follow type has an icon)
```
JakeChampion, line 109: these comments inside the interface are also great

components/myft-concept-button/src/js/myft-concept-button.js
```diff
@@ -0,0 +1,104 @@
...
+	 * Set the followed state
+	 * Updates the UI and adds text to the a11y box.
+	 *
+	 * @param {boolean} [state] the state to set
+	 */
+	set isPressed(state) {
+		this.button.ariaPressed = state ? 'true' : 'false';
+
+		if (
+			this.options.ariaLivePressedText &&
```
JakeChampion, line 32: question: for these three if statements - should we throw an error if only one of the options from each pair is supplied? I.E. If `this.options.ariaLivePressedText` is supplied but `this.options.ariaLiveUnpressedText` has not been supplied

JakeChampion (approved):

components/myft-concept-button/src/js/myft-concept-button.js
chee: yes this should probably error

chee: also if they add the `ariaLivePressedText` but `this.live` doesn't exist maybe we should make it and inject it

components/myft-concept-button/.gitignore
notlee: Oh no 🙈 this should go in `.npmignore`

components/ft-concept-button/main.scss
```diff
@@ -0,0 +1,239 @@
...
+		outline: unset;
+	}
+
+	// Styles given :focus-visible support. Extra selectors needed to match
+	// previous `:focus` selector specificity.
+	body:not(.js-focus-visible) &:focus-visible,
+	html:not(.js-focus-visible) &:focus-visible,
+	&:focus-visible {
+		outline: 2px solid $focus-color;
+	}
```
notlee, line 156: I think this outline is [handled by o-normalise](https://github.com/Financial-Times/origami/blob/a4d4b745da06235623b03293f1ac1de1ca65116d/components/o-normalise/src/scss/_mixins.scss#L110) and all we need to do here is change the colour?
```suggestion
	&:focus,
	&.focus-visible {
		outline-color: $focus-color;
	}
```

notlee (approved):
Couple of non-blocking suggestions (.npmignore, focus state) but otherwise approved! Nice work on this 🎉

JakeChampion (approved):

.github/workflows/percy-components-myft-follow-button.yml
JakeChampion: I raised a pull-request which checks for the existence of the percy token now ☺️ 
https://github.com/Financial-Times/origami/pull/654

PR #641, opened 2022-02-14
use `--o-colors-page-background` as default page background colour

chee:
## changes

- line 10 of withOrigamiBackground.js adds `|| "page-background"`
	-  so we default to whatever is the right background for the current theme
-  line 14 of withOrigamiBackground.js`.trim()`s the origamiBackgroundHex before using it
	- because for some reason the returned hex on it had a space in it when i was testing locally!?!?!
- prettier was run on the code

### DIFF
--- a/apps/storybook/addons/background/src/preset/preview.js
+++ b/apps/storybook/addons/background/src/preset/preview.js
@@ -1,18 +1,23 @@
-import { withOrigamiBackground } from '../withOrigamiBackground';
+import {withOrigamiBackground} from "../withOrigamiBackground"
 
-export const decorators = [withOrigamiBackground];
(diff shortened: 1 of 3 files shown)


apps/storybook/addons/background/src/preset/preview.js
```diff
@@ -1,18 +1,23 @@
...
+		name: "Origami Background",
+		description: "Background colour to show components on.",
+		toolbar: {
+			icon: "photo",
+			items: [
+				{
+					value: "",
+					title: "recommended",
+				},
+				"paper",
```
notlee, line 16: Nice 👍 let's add white as an option!
```suggestion
				"paper",
				"white",
```

notlee (approved):

PR #642, opened 2022-02-14
add ghost button type

chee:


chee:
https://origami-webs-buttons-gh-uxzd2v.herokuapp.com/storybook/?path=/story/components-o-buttons--ghost-button

i think it should be here when it's ready

JakeChampion (approved):

PR #644, opened 2022-02-14
design tokens: o-colors

notlee:
This draft PR moves colour pallet and usecase definitions for each brand from Sass to design tokens. Then generates CSS, Sass, and JSON from that. It's a bit gnarly. It might be preferable to deprecate o-colors and other low level components like o-spacing for a simpler bucket of variables. This draft PR is not ready for review but is ready for discussion.

Very much hitting this with a hammer to see what happens. See `@todo` code comments for extra details.

____

- Defines o-colors colour palette and usecases for each brand using design tokens.
- On `npm run -w components/o-buttons build` Style dictionary writes Sass, CSS, and JSON for o-colors in `components/o-buttons/src/dist-tokens/[brand]` (bad name, bad place, yes).
- Generates `components/o-buttons/src/dist-tokens/[brand]/backward-compatible-tokens.scss` with tokens for all brands, prefixed with brand and component name, which is imported in  `components/o-colors/src/scss/_brand.scss` (also generated) and the correct token map is [assigned to a global Sass variable](https://github.com/Financial-Times/origami/pull/644/files#diff-915b1cc35291f84b95896576781ee8c5eef26ab6ffae2f194071118bab9fc29eR17) depending on the currently set brand (for backward compatibility with the existing `$o-brand` Sass variable).
- Maps old api to new tokens.

Pretty sure we could look at deprecating most of our low level components for a set of variables. Consider:

```
Mixins

    oColors - keep, switch to namespaced module and rename to `style` in keeping with official Sass example
    oColorsSetColor - deprecate, use CSS Custom Properties or Sass variables
    oColorsSetUseCase - deprecate, use CSS Custom Properties or Sass variables

Functions

    oColorsByName - deprecate, use CSS Custom Properties or Sass variables
    oColorsByUsecase - deprecate, use CSS Custom Properties or Sass variables
    oColorsGetTone - deprecate, use CSS Custom Properties or Sass variables
    oColorsGetToneDetails - deprecate, used to generate button colours, write design guidelines and define each button colour explicitly
    oColorsMix - deprecate, Sass can do this with hex colours itself (however we mix with the brands default background colour by default, helpful or not?)
    oColorsGetTextColor - move to or require to a library?
    oColorsColorBrightness - deprecate, used by oColorsGetTextColor only?
    oColorsColorLuminance - deprecate, looks like Sass can do this https://sass-lang.com/documentation/modules/color#whiteness
    oColorsGetContrastRatio - deprecate? looks like it's only used sparingly in Origami components. encourage contrast checking outside of sporadic Sass enforcement?
    oColorsGetPaletteDetails -  deprecate, used internally only, replace with sass map

Variables

    o-colors-is-silent - deprecate, use primary mixin
```


components/o-colors/src/scss/_brand.scss
```diff
@@ -0,0 +1,28 @@
+
+			@import '@financial-times/o-brand/main';
```
notlee, line 2: this is generated and shouldn't be commited (and shouldn't be in this dir)

libraries/o-brand/scss/_deprecated.scss
```diff
@@ -13,3 +13,14 @@
...
 	@return $brand;
 }
+
+/// Get current brand in use.
+/// This function was deprecated when the "master" brand was renamed "core".
+/// It will continue to return "master" if "core" is the set brand for backward compatibility.
+/// @deprecated - Use `oBrandIs($brand)` to check if a given brand is the current brand.
+/// @access public
+///
+@function oBrandGetCurrentBrandNew() {
```
notlee, line 23: todo rename this, `oBrandGetCurrentBrand` (without `New`) is deprecated and returns `master` instead of `core` for backward compatibility reasons

tools/origami-style-dictionary/README.md
```diff
@@ -0,0 +1,181 @@
+# Basic Style Dictionary
+
```
notlee, line 2: todo: lies all lies

notlee:
We have decided to roll out new, design token driven components alongside existing components.

PR #660, opened 2022-03-18
backstage: update origami via bower migration guide

notlee:
Customer Products are first moving to npm without bumping
component version numbers, i.e. moving to the beta npm releases.
This is so they can release packages with both bower and npm
support to help the rollout. This has caused confusion because
it means they need to add `node_modules/@financial-times` to their
compiler's `includePath` configuration, it is not an option anymore.

The guide has been updated the cover that case. It has also been
updated to require Sass imports include the `@financial-times`
namespace, rather than rely on the compiler's `includePath`
configuration. This is because a project which does that forces
all its depndents to also.

### DIFF
--- a/apps/website/_posts/2021-07-01-origami-on-npm-and-how-to-migrate.md
+++ b/apps/website/_posts/2021-07-01-origami-on-npm-and-how-to-migrate.md
@@ -61,7 +61,7 @@ Your migration to npm Origami components is finished! 🎉
 ### Using Origami via Bower
 
 <aside>
-This guide used to give a choice between updating Sass include paths ( <code>@import  <code>o-icons/main';</code> becomes  <code>@import '@financial-times/o-icons/main';`</code>) or adding <code>node_modules/@financial-times</code> to your compiler's <code>includePath<code> configuration. It has since been updated to require the former and discourage the latter outside a specific circumstance.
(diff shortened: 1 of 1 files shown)


apps/website/_posts/2021-07-01-origami-on-npm-and-how-to-migrate.md
```diff
@@ -59,22 +60,26 @@ Your migration to npm Origami components is finished! 🎉
 
 ### Using Origami via Bower
 
+<aside>
+This guide used to give a choice between updating Sass include paths ( <code>@import  <code>o-icons/main';</code> becomes  <code>@import '@financial-times/o-icons/main';`</code>) or adding <code>node_modules/@financial-times</code> to your compiler's <code>includePath<code> configuration. It has since been updated to require the former and discourage the latter outside a specific circumstance.
```
chee, line 64:
```suggestion
This guide used to give a choice between updating Sass include paths ( <code>@import  <code>o-icons/main';</code> becomes  <code>@import '@financial-times/o-icons/main';</code>) or adding <code>node_modules/@financial-times</code> to your compiler's <code>includePath<code> configuration. It has since been updated to require the former and discourage the latter outside a specific circumstance.
```

i think this backtick is extra

chee (approved):

PR #662, opened 2022-03-21
fix: bring support for firefox to ft-concept-button

JakeChampion:
ariaPressed and ariaLabel are not yet supported in Firefox - we can use setAttribute instead

https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaPressed#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaLabel#browser_compatibility

### DIFF
--- a/components/ft-concept-button/src/js/ft-concept-button.js
+++ b/components/ft-concept-button/src/js/ft-concept-button.js
@@ -49,7 +49,8 @@ export default class FtConceptButton {
 		) {
 			this.button.setAttribute('aria-label', state
 				? this.options.ariaLabelPressedText
-				: this.options.ariaLabelUnpressedText;
+				: this.options.ariaLabelUnpressedText
+			);
 		}
 	}
 


components/ft-concept-button/src/js/ft-concept-button.js
```diff
@@ -47,7 +47,7 @@ export default class FtConceptButton {
 			this.options.ariaLabelPressedText &&
 			this.options.ariaLabelUnpressedText
 		) {
-			this.button.ariaLabel = state
+			this.button.setAttribute('aria-label', state
 				? this.options.ariaLabelPressedText
 				: this.options.ariaLabelUnpressedText;
```
chee, line 52:
```suggestion
			this.button.setAttribute('aria-label', state
				? this.options.ariaLabelPressedText
				: this.options.ariaLabelUnpressedText
			);
```

JakeChampion: thank you!

PR #669, opened 2022-03-23
Move g-audio component into monorepo

joannaskao:
- Move g-audio from https://github.com/Financial-Times/g-audio into monorepo (and aimed to keep the entire history)
- Remove bower dependency and use npm
- Update o-icons dependency
- Add Percy integration


components/g-audio/.eslintrc
```diff
@@ -0,0 +1,41 @@
...
+  "parserOptions": {
+  "sourceType": "module"
+  },
+  "env": {
+  "es6": true,
+  "browser": true
+  },
+  "rules": {
+  "no-unused-vars": 2,
+  "no-undef": 2,
```
notlee, line 14: Can we delete this file in favour of [.eslintrc.cjs](https://github.com/Financial-Times/origami/pull/669/files#diff-afe2a5db742c132e436430734ead35c04f90c7273d1befc07922290bb280d874)?

components/g-audio/.npmignore
```diff
@@ -0,0 +1,15 @@
...
+/build/
+.idea/
+/demos/local
+/coverage
+.eslintrc.cjs
+.github/
+.gitignore
+.remarkrc.cjs
+.stylelintrc.cjs
+test
```
notlee, line 15: non-blocking, brings `.npmignore` inline with other components

```suggestion
.env
.gitignore
.npmignore
.eslintrc.cjs
.remarkrc.cjs
.stylelintrc.cjs
.prettierrc.toml
.eslintignore
.stylelintignore
stories/
src/tsx/
.github/
test/
demos/local/

```

components/g-audio/.scss-lint.yml
```diff
@@ -0,0 +1,124 @@
+# sass-lint config generated by make-sass-lint-config v0.1.2
+#
+# The following scss-lint Linters are not yet supported by sass-lint:
+# ElsePlacement, PropertyCount, SelectorDepth, UnnecessaryParentReference
```
notlee, line 4: Can this file be deleted in favour of `[.stylelintrc.cjs](https://github.com/Financial-Times/origami/pull/669/files#diff-f296198ff6e5bd5daaaff5e4ff476f894c8e4d666137736453936a82a680bc2c)?

components/g-audio/src/scss/_audio.scss
```diff
@@ -0,0 +1,83 @@
+$gaudio-horz-padding-inline: 5px;
+$gaudio-vert-padding-inline: 3px;
+
+$gaudio-horz-padding-block: 15px;
+$gaudio-vert-padding-block: 10px;
+
+$gaudio-button-background: #9e2f50;
+$gaudio-button-color: #f6e9d8;
+
+.g-audio.g-audio--initialized {
```
notlee, line 10: Could we wrap CSS in a `gAudio` mixin? We don't have a spec to enforce that any more but this would be the only Origami component which outputs CSS immediately on import.

notlee (changes requested):
Nice! 👏 Just one request to wrap the Sass in a mixin, to follow an Origami pattern so no CSS is output automatically on import. Plus a couple of little non-blocking suggestions.

components/g-audio/src/scss/_audio.scss
notlee: Sass guide here 👉 https://origami.ft.com/documentation/components/code/#sass

components/g-audio/.eslintrc
JakeChampion: I deleted it 👍

components/g-audio/.scss-lint.yml
JakeChampion: I deleted it 👍

components/g-audio/src/scss/_audio.scss
joannaskao: Thanks! I think I've just done this — can you review and see if I've done it right?

components/g-audio/src/scss/_audio.scss
```diff
@@ -7,77 +7,79 @@ $gaudio-vert-padding-block: 10px;
...
+		}
 
-	&.pause .g-audio--playbutton:before {
-		@include oIconsContent('pause', $gaudio-button-color, 25);
-		content: '';
-		font-size: 0.8em;
-		margin: -3px $gaudio-horz-padding-inline;
-		vertical-align: bottom;
-	}
+		&.pause .g-audio--playbutton:before {
```
notlee: Oh sorry, one more thing! This class should be namespaced too `g-audio-pause`

components/g-audio/src/scss/_audio.scss
joannaskao: Okay, have just added that!

notlee (approved):
🙌

PR #689, opened 2022-04-04
ACQ-1536: add some css needed for new drawer menu component

ekocian:
Add some CSS needed for a new component in drawer menu.

See https://github.com/Financial-Times/dotcom-page-kit/pull/940 for further details

### DIFF
--- a/components/o-header/demos/src/header.mustache
+++ b/components/o-header/demos/src/header.mustache
@@ -141,7 +141,7 @@
 
 
 		<div class="o-header__drawer-actions">
-			<a class="o-header__drawer-button"  href="#"">
+			<a class="o-header__drawer-button"  href="#">
 				subscribe
 			</a>
 		</div>
--- a/components/o-header/demos/src/header.mustache
+++ b/components/o-header/demos/src/header.mustache
@@ -136,6 +136,13 @@
 			{{/editions}}
 		</div>
 
+
+		<div class="o-header__drawer-actions">
+			<a class="o-header__drawer-button"  href="#"">
+				subscribe
+			</a>
+		</div>
+
 		<div class="o-header__drawer-search">
 			<form class="o-header__drawer-search-form" action="/search" role="search" aria-label="Site search">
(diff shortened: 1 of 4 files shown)


ekocian:
> Thanks @ekocian 🙏 Could you share the markup for this? I'll add it to a demo.
> 
> FYI here's the related PR: #690

image

@notlee This is how it looks included in header, not sure if clear and enough for you. Otherwise just let me know and I can complete it. Thanks!

notlee:
Thanks for opening the PR @ekocian! 🙇 I've added that to the demo. I also chose to update the name of the button class for this use-case. Hope that's okay? Using a new class name for the new drawer button reduces specificity slightly and makes it a little easier for us to adapt going forward, since markup changes are more difficult in Origami as we don't currently provide markup templates for o-header.

If this looks good to you I'll get it release, let me know

notlee:
FYI we use [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) messages to manage Origami releases so when I merge this I'll squash our commits and update your message. There's usually a warning on commit but it only shows after running `npm install` so I'm guessing you didn't see. I'll update the readme for future contributors! 🙌

components/o-header/demos/src/header.mustache
```diff
@@ -139,6 +139,13 @@
 			{{/editions}}
 		</div>
 
+
+		<div class="o-header__drawer-actions">
+			<a class="o-header__drawer-button"  href="#"">
```
notlee, line 144:
```suggestion
			<a class="o-header__drawer-button"  href="#">
```

ekocian:
> Thanks for opening the PR @ekocian! 🙇 I've added that to the demo. I also chose to update the name of the button class for this use-case. Hope that's okay? Using a new class name for the new drawer button reduces specificity slightly and makes it a little easier for us to adapt going forward, since markup changes are more difficult in Origami as we don't currently provide markup templates for o-header.
> 
> If this looks good to you I'll get it release, let me know

Thanks @notlee! I think change the class name is working perfectly fine with this changes. Feel free to release it whenever you want
image

Issue #688, opened 2022-04-04
Update o-header L breakpoint (tablet/desktop) and spacing for subscribe button changes


notlee:
More details: https://github.com/Financial-Times/dotcom-page-kit/pull/939

Related PR: https://github.com/Financial-Times/origami/pull/685

notlee:
- nice and easy to add a new breakpoint
- at 980px the entire myft button is cropped
- 1080px landscape ipad looks ok-ish, but spacing is not so good
- I think the idea is that the left/right column is the same size and the center column is in the middle, looks a bit off here
may need to spend some time sorting out header spacing

![Screenshot 2022-03-31 at 13 56 29](https://user-images.githubusercontent.com/10405691/161528973-7e346c3e-5588-42c5-a2bf-d515c16ca806.png)

PR #690, opened 2022-04-04
feat: update o-header with new subs button for annon readers

notlee:
Markup changes here are non-breaking of previous markup.
Updates will be rolled out across ft.com via pagekit

Ticket: https://financialtimes.atlassian.net/browse/ACQ-1535
Closes: https://github.com/Financial-Times/origami/issues/688

### DIFF
--- a/components/o-header/src/scss/features/_top.scss
+++ b/components/o-header/src/scss/features/_top.scss
@@ -1,38 +1,41 @@
 @mixin _oHeaderTop() {
 	.o-header__top-wrapper {
-		display: table;
-		width: 100%;
-	}
-
-	.o-header__top-column {
-		display: table-cell;
-		vertical-align: middle;
+		display: grid;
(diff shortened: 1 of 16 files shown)


components/o-header/src/scss/features/_top.scss
```diff
@@ -71,20 +72,14 @@
 		}
 	}
 
-	.o-header__nav-item:first-child {
-		> .o-header__top-link--menu,
-		> .o-header__top-link--search,
-		> .o-header__top-link--myft {
-			margin-left: -#{$_o-header-icon-size-large * 0.25px}; // remove space built into icon image, -10px size given 40px icon.
+	.o-header__top-link--hide-s {
```
JakeChampion, line 75: question: what does `hide-s` mean? does it mean hide subscription?

notlee: hide on small viewports (don't love the name but following an existing pattern)

PR #692, opened 2022-04-06
fix: center ft logo in o-header when right/left columns are missing

notlee:
this pattern is used on login/signup pages for a full size logo


JakeChampion (approved):
lgtm

PR #695, opened 2022-04-07
backstage: recombine github test and percy workflows

JakeChampion:
Let me know what you think to this :-)

The reason we created separate workflows was because github could only rerun entire workflows.
github added the ability to rerun only the failed jobs within a workflow!
switching from multiple test workflows to one workflow enables us to make full use of this new github rerun functionality

previously if multiple test or percy workflows failed, we would have to go and manually rerun each individual workflow
by having a single test workflow and percy workflow we will be able to rerun all the failed test jobs in one go

image


notlee (approved):
cool!

PR #698, opened 2022-04-08
fix: add missing system_code query parameter using the $system-code global variable

JakeChampion:


notlee (approved):

chee:
@JakeChampion let's also take the trailing slash off here? https://github.com/Financial-Times/origami/blob/bcaf6ab39ea6df0553c1f89eca9bf6321bc6d3b9/components/o-fonts/src/scss/_variables.scss#L9

notlee:
I think this will affect font cache / pagekit's front preloading? Perhaps the system code should be `origami` in this case.

chee:
> I think this will affect font cache / pagekit's front preloading? Perhaps the system code should be `origami` in this case.

@Financial-Times/platforms  is this true?

notlee:
Because each project will set its own system code I think fonts would be downloaded again moving from the front page to the article page for example?

apaleslimghost:
yeah, that's correct. maybe we could introduce an `$o-fonts-system-code` that defaults to `$system-code` but would let us override it to (e.g.) `page-kit`?

Issue #485, opened 2021-12-16
Add "main" field to package.json


aendra-rininsland:
`package.json` files in Origami tend to specify a `browser` field but not a `main` field. This makes it substantively more difficult to use with build tools, particularly more opinionated ones like NextJS.

Is there any reason why Origami doesn't provide a `main` field in `package.json`, particularly when it already provides a `browser` field referencing the transpiled output? In my experience, pretty much all front-endian npm packages I've ever consumed have pointed the `main` field at the transpiled output.

notlee:
Hey 👋 

## Browser Field

[The `package.json` docs say](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#browser) the following about the browser field, I guess providing a `main` field too would defeat the point. But @JakeChampion or @chee might have more context about the decision to use the `browser` field over `main`:
>If your module is meant to be used client-side the browser field should be used instead of the main field. This is helpful to hint users that it might rely on primitives that aren't available in Node.js modules. (e.g. window)

Origami component JS does only support a browser environment, the first thing it does is `document.addEventListener` which would error in an SSR context. So I guess it makes sense to be explicit by using the `browser` field.

## NextJS

It looks like `NextJS` initialises 2 instances of webpack? One that resolves the `browser` field and one that doesn't. I think that because... 

The following config in `next.config.js`:
```
module.exports = {
  reactStrictMode: true,
  webpack: (config, options) => {
    console.log(config.resolve.mainFields);
    return config;
  }
}
```
Outputs:
```
ready - started server on 0.0.0.0:3000, url: http://localhost:3000
[ 'browser', 'module', 'main' ]
[ 'main', 'module' ]
```

I found by adding `browser` to the `mainFields` of both webpack instances (?) I could get `oHeader` setup in a NextJS app:
```
module.exports = {
  reactStrictMode: true,
  webpack: (config, options) => {
    console.log(config.resolve.mainFields);
    config.resolve.mainFields.push('browser');
    return config;
  }
}
```

That doesn't feel right though. 🤔 

For reference here's the JS for my test NextJS page

```
import React from 'react'

class Header extends React.Component {
	constructor(props) {
		super(props);
	}

	async componentDidMount() {
		const oHeader = (await import('@financial-times/o-header')).default;
		this.headerInstance = new oHeader(this.el);
		console.log(this.headerInstance);
	}

	componentWillUnmount() {
		this.headerInstance.destroy();
	}

	render() {
[...]

PR #712, opened 2022-04-20
Checks for document before adding o.DOMContentLoaded event listener in o-toggle

aendra-rininsland:
At present, o-toggle is used in both o-header and o-footer.

I want to be able to SSR headers and footers, because there's not much point in SSR if you need a client-side effect to render the header and footer (resulting in large CLS, part of the point of SSR being to avoid). However, because of the Origami practise of accessing `document` as a module-importing side-effect, this is utterly impossible at present and causes e.g. NextJS to throw `ReferenceError: document is not defined`

If Origami is going to continue to follow its idiom of adding event listeners as a side effect on module import, could you please at least existence check for `document` before accessing it? This would solve like 90% of my problems.

### DIFF
--- a/components/o-toggle/main.js
+++ b/components/o-toggle/main.js
@@ -5,6 +5,6 @@ const constructAll = () => {
 	document.removeEventListener('o.DOMContentLoaded', constructAll);
 };
 
-if (document) document.addEventListener('o.DOMContentLoaded', constructAll);
+if (typeof document !== 'undefined') document.addEventListener('o.DOMContentLoaded', constructAll);
 
 export default Toggle;
--- a/components/o-toggle/main.js
+++ b/components/o-toggle/main.js
@@ -5,6 +5,6 @@ const constructAll = () => {
 	document.removeEventListener('o.DOMContentLoaded', constructAll);
 };
 
-document.addEventListener('o.DOMContentLoaded', constructAll);
+if (document) document.addEventListener('o.DOMContentLoaded', constructAll);
 
 export default Toggle;
--- a/components/o-toggle/main.js
+++ b/components/o-toggle/main.js
@@ -5,6 +5,8 @@ const constructAll = () => {
 	document.removeEventListener('o.DOMContentLoaded', constructAll);
 };
 
-if (typeof document !== 'undefined') document.addEventListener('o.DOMContentLoaded', constructAll);
+if (typeof document !== 'undefined') {
+	document.addEventListener('o.DOMContentLoaded', constructAll);
+}
 
 export default Toggle;


components/o-toggle/main.js
```diff
@@ -5,6 +5,6 @@ const constructAll = () => {
 	document.removeEventListener('o.DOMContentLoaded', constructAll);
 };
 
-document.addEventListener('o.DOMContentLoaded', constructAll);
+if (document) document.addEventListener('o.DOMContentLoaded', constructAll);
```
chee, line 8: this'll still error if document is not defined.

```suggestion
if (typeof document !== 'undefined') document.addEventListener('o.DOMContentLoaded', constructAll);
```

this should work though, right?

chee, line 8:
```suggestion
if (typeof document !== 'undefined') {
	document.addEventListener('o.DOMContentLoaded', constructAll);
}
```

maybe something like this

aendra-rininsland: Updated as per this suggestion.

chee:
i'm fine with merging this, we should do it on the rest of our components too. @notlee any thoughts ?

notlee:
Something I think we could do is provide a `main` field in package.json for all components, which doesn't add the event listener.

```
import Toggle from './src/js/toggle.js';
export default Toggle;
```

Orrrr... an empty js file.
```
```

Jake didn't like that because it suggests our components support a node environment, as we're
providing an entry point. On the other hand I like that we're explicit we expect a component to be 
resolvable in a node environment, for SSR, but that nothing should happen.

Relates to: https://github.com/Financial-Times/origami/issues/485#issuecomment-996148199

This approach works though! I wonder if a `main` field in package.json which points to something
empty is more explicit.

notlee:
Either way yeah let's do the same for all components :)

notlee:
Something I missed out is that with this approach it still requires [tampering with Next.js config](https://github.com/Financial-Times/origami/issues/485#issuecomment-996126999) to get it to resolve the `browser` field in a node env.

notlee (approved):

chee:
> Something I missed out is that with this approach it still requires [tampering with Next.js config](https://github.com/Financial-Times/origami/issues/485#issuecomment-996126999) to get it to resolve the `browser` field in a node env.

i prefer making the code check for window because there are many different bundlers and i don't know how they all behave, and i don't want to end up with people thinking they have working code but not (because it finds the empty main and doesn't error, and then their front-end bundle doesn't work) and i don't want to spend time checking every bundler and builder to see how it behaves and we know this approach works already

aendra-rininsland:
Just to leave a final thought about `main` versus `browser`: with the addition of [`package.exports`](https://webpack.js.org/guides/package-exports/) package.json is increasingly used as part of bundler module resolution methods and I'd argue `package.main` is now less "the NodeJS entry point" and more "the CommonJS version of the entry point", with `package.module` being the ESM equivalent. Further with `package.exports` you can conditionally load the same entry point depending on environment; see https://webpack.js.org/guides/package-exports/#target-environment

So you could leave off both `main` and `module`, leave `browser` as it is currently, then provide `exports` like:

```
"browser": "./main.js",
"exports": {
  "browser": "./main.js",
  "node": "./src/index.js"
}
```

This way modern tooling will consume `package.exports` and choose the right version (🤞 ), existent tooling already setup to grab `browser` will continue to do so. 

...Though worth noting though that my experience with NextJS today (20 April 2022, for posterity) is that it completely disregards the conditional loading and defaults to CommonJS if you have .cjs files in the same directory and with the same name as the entrypoint, was extremely confused for a few hours this afternoon! @chee's point about not being able to anticipate how any given bundler will act is well-taken; I think my above suggestion will work with Next but I haven't verified yet. 

Further reading that might be helpful: https://webpack.js.org/guides/package-exports/#common-patterns

Issue #159, opened 2021-09-16
Deprecate o-toggle?


notlee:
The `hidden` option of [o-expander](https://github.com/Financial-Times/o-expander) also allows content to be toggled between hidden and shown states but it doesn't have any callback functionality outside events.

To reduce maintenance and increase the distinction between these: Should `o-toggle` be deprecated? Or should the `hidden` option be removed from `o-expander`? Look into how these components are used to make a decision.

chee:
my opinion is that o-expander is too complex, and it has features that i couldn't find being used anywhere. i think we can drop `data-o-expander-shrink-to` and `data-o-expander-item-selector`. it has the option to shrink to a specific max-height, and a specific number of list items, and an option to disable setting aria-attributes. i'm not sure it should have those. i think it would be great to talk to some designers about toggleable content. i guess we call these something like a disclosure component.

also i think possibly the code for o-toggle is too complex, i feel like we could write something that is smaller and simpler and more robust for this and doesn't have two separate javascript classes for the toggler and the target.

a simpler expander looks like:
- label when collapsed
- label when expanded
- a heading/summary (optional)
- child content

a toggle looks like:
- label for the button
- child content

they are quite different, but without guidance it might be confusing which one to use. they are both essentially a custom implementation of the native `<details/>` disclosure element which actually exists already in all our supported browsers (it used to have accessibility problems but i'm not sure it still does). i personally prefer the `o-toggle` shape if we could maybe have it redesigned slightly (perhaps it should have a ▶️/🔽  icon for showing when it is open and closed)

i'm just sort of thinking out loud here but i think maybe i would rather deprecate _o-expander_ (because i think o-toggle is closer to what we would want), the . so we'd talk to design about the idea of a disclosure component and then have something like:

```
<div data-o-component="o-disclosure">
  <button type="button o-toggle-button" aria-expanded="false" aria-controls="toggle-target">
    <span class="o-disclosure-icon">
      ▶️
    </span>
    Dogs
  </button>
  <div class="o-disclosure-target" id="toggle-target" aria-hidden="true">
    <ul>
      <li> border terrier
      <li> border collie
      <li> irish wolfhound
    </ul>
  </div>
</div>
```

which is not far off one of the current shape of `o-toggle`, and so maybe it would still be called `o-toggle`. but basically i'm suggesting we get a design and then we drop _all_ of the different options for o-toggle and o-expander and allow only the exact shape of that design

notlee:
[...]

PR #719, opened 2022-04-22
Remove most of the features of o-toggle

chee:
- removes visibility/display options
- removes most of the javascript (¡¡now it just toggles a single aria-attribute!!)
- removes the `callback` feature !!
- makes the css a bit smaller !!
- adds a storybook :)

it's a major because of the deleted features, [¡here's the migration guide!](https://github.com/Financial-Times/origami/blob/e205bf8ff631a92e53d06b45769f4471e94cd758/components/o-toggle/MIGRATION.md)


chee:
before we merge this, let's check if anyone has a toggle button whose hidden content does not immediately follow the DOM.

chee:
> before we merge this, let's check if anyone has a toggle button whose hidden content does not immediately follow the DOM.

done! it was! updated the code to allow that!!!!!!

notlee:
@chee sorry I've totally forgotten where we got to with this work? Is it ready for review or did we find other usecases to consider? Oh! Was it o-header/o-footer? My instinct is to not use o-toggle as dependency

chee:
> @chee sorry I've totally forgotten where we got to with this work? Is it ready for review or did we find other usecases to consider? Oh! Was it o-header/o-footer? My instinct is to not use o-toggle as dependency

yes, that was it. o-header and o-footer depend on it. it's nice that they fail their tests when i changed o-toggle without checking. i like the monorepo.

notlee:
@chee please update this with a summary of your findings and screenshots of usecases etc.

chee:
moving this conversation to #159

PR #725, opened 2022-05-11
feat: o-buttons add ghost mono / inverse themes

notlee:
These are being added to the Figma UI Kit


notlee:
@Financial-Times/origami-core Percy is regenerating all snaps again!? Ignore these Percy runs

chee (approved):

PR #726, opened 2022-05-11
Move ftdomdelegate to the Origami mono repo

notlee:
Git history maintained in the same way as our other packages (I think):
`git subtree add -P libraries/ftdomdelegate git@github.com:Financial-Times/ftdomdelegate master`

Then I deleted its old `.github` worflows and the lock file, added a bugs key to package.json, and ran `npm run regenerate` (what a joy of a command).

See commit: 90af9b0755438be2b70dc9988a19dc5d09ffd31d
From: https://github.com/Financial-Times/ftdomdelegate


chee (approved):
wooooo!

notlee:
@chee sorry about the force push. Thought I could get in there before your review but you were too on it. I had forgotten to add the version number to package.json

chee:
> @chee I think this will have to be a merge commit to keep ftdomdelegate history? I can temporarily allow merge commits but just wanted to double check

yeah, that should be fine. we could just leave merge commits on. ¿are they even a problem? who is out here looking at the tree

PR #727, opened 2022-05-12
feat!: Remove o-comments reliance on Build Service v2

notlee:
o-comments relied on v2 of the Build Service to build `src/scss/coral-talk-iframe.scss` for the Coral Talk iframe, which would break when we drop Bower support.
https://github.com/Financial-Times/origami/tree/o-comments-v8.3.3/components/o-comments#styling-of-coral-talk-iframe

There’s no direct migration path to Build Service v3 in this case as it’s no longer possible to use the build service to bundle arbitrary sass files in the component’s src.

Instead output the coral talk iframe css as part of `main.scss` and rely on [Coral's containerClassName option](https://docs.coralproject.net/cms) to scope CSS to the iframe. Provide an option in the Sass interface to not output the styles intended for the Coral iframe, to reduce final project css bundle sizes.

Pros:
- All comment CSS co-located in one project.
- Simple to implement and document.
- No special treatment in other projects (e.g. Build Service exception) to document and maintain.

Cons:
- A little redundant CSS in the iFrame embed.
- A slightly less nice Sass interface (the option to exclude iframe styles should be set).

### DIFF
--- a/components/o-comments/MIGRATION.md
+++ b/components/o-comments/MIGRATION.md
@@ -1,5 +1,14 @@
 ## Migration
 
+### Migrating from v8 to v9
+
+To migrate update your projects use of the `oComments` sass mixin. Set its `coral-talk-iframe` option to `false`.
+
+```diff
+-@include oComments();
++@include oComments($opts: ('coral-talk-iframe': false));
+```
+
 ### Migrating from v7 to v8
 
 Support for Bower and version 2 of the Origami Build Service have been removed.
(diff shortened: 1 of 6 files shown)


notlee:
*do not merge*: The [containerClassName](https://docs.coralproject.net/cms) should be set to `o-comments-coral-talk-container` first, if possible

GlynnPhillips (approved):
Looks good to me. Thanks for helping with this Lee.

Once released I can test and upgrade the consumers

PR #731, opened 2022-05-16
Add stories for o-date demos

ivomurrell:
The original o-date demo is unusual in that multiple variations are included in a single demo, with different formatting settings included in one chunk of HTML. There are demos split out into ones written in declarative and imperative styles, though the imperative demos are already hidden so I haven't ported them over. I've instead split each line of the declarative demo into its own story so that their arguments can be fiddled with individually.

o-date also has support for 'copy options', where you can intersperse computed timestamps with some custom copy. This is done via inserting `span`s with some identifying attributes (as long as optionally setting custom formatting.) I can think of a few ways we could make this fit more ergonomically with the new JSX templates (converting string templates into HTML – too abstracted, using friendlier naming conventions and then mapping them – not sure how stable this would be and seems unnecessary,) so have just left the children to be forwarded to the js code for now.

Made an effort to make the Storybook controls reflect the semantics of the `Date` props, so have properly typed them, though adding support for the Storybook date picker means we need to convert between ISO strings and UNIX timestamps, which is fun.

I've kept the inline description of each format in the stories for continuity though perhaps they should go into a documentation field somewhere instead?

### DIFF
--- a/components/o-date/src/tsx/date.tsx
+++ b/components/o-date/src/tsx/date.tsx
@@ -31,3 +31,18 @@ export function Date({
 		</time>
 	);
 }
+
+export interface DatePrinterProps {
+	format?: DateFormat;
+}
+
+export function DatePrinter({
+	format,
+	children,
+}: React.PropsWithChildren<DatePrinterProps>): JSX.Element {
+	return (
+		<span data-o-date-printer data-o-date-format={format}>
+			{children}
+		</span>
+	);
+}
--- a/components/o-date/src/tsx/date.tsx
+++ b/components/o-date/src/tsx/date.tsx
@@ -1,0 +1,33 @@
+export type DateFormat =
+	| 'today-or-yesterday-or-nothing'
+	| 'date-only'
+	| 'time-ago-limit-4-hours'
+	| 'time-ago-limit-24-hours'
+	| 'time-ago-abbreviated'
+	| 'time-ago-abbreviated-limit-4-hours'
+	| 'time-ago-no-seconds'
+	// hack to allow any string (for custom date formatting) while still
+	// suggesting the common format variants in IDEs
+	// from https://github.com/microsoft/TypeScript/issues/29729#issuecomment-505826972
+	| (string & {});
+
+export interface DateProps {
+	dateTime: string;
+	format?: DateFormat;
+}
+
+export function Date({
+	dateTime,
+	format,
+	children,
+}: React.PropsWithChildren<DateProps>): JSX.Element {
+	return (
+		<time
+			data-o-component="o-date"
+			className="o-date"
+			dateTime={dateTime}
+			data-o-date-format={format}>
+			{children}
+		</time>
+	);
+}
--- a/components/o-date/stories/date.stories.tsx
+++ b/components/o-date/stories/date.stories.tsx
@@ -45,10 +45,10 @@ export default {
 
 const Template: DateStory = ({description, ...args}) => {
 	useEffect(() => {
-		let tabs = javascript.init();
+		let dates = javascript.init();
 		return function cleanup() {
-			tabs = Array.isArray(tabs) ? tabs : [tabs];
-			tabs.forEach(tab => tab.destroy());
+			dates = Array.isArray(dates) ? dates : [dates];
+			dates.forEach(date => date.destroy());
 		};
 	});
 	return (
--- a/components/o-date/stories/date.stories.tsx
+++ b/components/o-date/stories/date.stories.tsx
@@ -1,0 +1,118 @@
+import type {ComponentMeta, Story} from '@storybook/react';
+import withHtml from 'origami-storybook-addon-html';
+import {ComponentProps, useEffect} from 'react';
+import {withDesign} from 'storybook-addon-designs';
+import javascript from '../main';
+import {Date as ODate} from '../src/tsx/date';
+
+// the storybook args are slightly different to the Date props:
+// 1. dateTime is a UNIX timestamp, the format storybook uses in its controls
+// 2. the description field appends a description of the timestamp inline
+type DateStory = Story<
+	Omit<ComponentProps<typeof ODate>, 'dateTime'> & {
+		dateTime: number;
+		description: string;
+	}
+>;
+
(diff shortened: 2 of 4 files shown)


components/o-date/src/tsx/date.tsx
```diff
@@ -0,0 +1,35 @@
...
+	| 'date-only'
+	| 'time-ago-limit-4-hours'
+	| 'time-ago-limit-24-hours'
+	| 'time-ago-abbreviated'
+	| 'time-ago-abbreviated-limit-4-hours'
+	| 'time-ago-no-seconds'
+	// hack to allow any string (for custom date formatting) while still
+	// suggesting the common format variants in IDEs
+	// from https://github.com/microsoft/TypeScript/issues/29729#issuecomment-505826972
+	| (string & {});
```
notlee, line 12: ooo interesting! nice

components/o-date/stories/date.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+		fallback: {control: 'text'},
+		// the description doesn't need to be editable
+		description: {table: {disable: true}},
+	},
+	parameters: {controls: {sort: 'requiredFirst'}},
+} as ComponentMeta<typeof ODate>;
+
+const Template: DateStory = ({description, ...args}) => {
+	useEffect(() => {
+		let tabs = javascript.init();
```
notlee, line 49: oops, tabs should be renamed

components/o-date/stories/date.stories.tsx
```diff
@@ -0,0 +1,119 @@
+import type {ComponentMeta, Story} from '@storybook/react';
+import withHtml from 'origami-storybook-addon-html';
+import {ComponentProps, useEffect} from 'react';
+import {withDesign} from 'storybook-addon-designs';
+import javascript from '../main';
+import {Date as ODate} from '../src/tsx/date';
+
+// the storybook args are slightly different to the Date props:
+// 1. dateTime is a UNIX timestamp, the format storybook uses in its controls
```
notlee, line 9: Helpful 👍

components/o-date/stories/date.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+export const Multiple: DateStory = Template.bind({});
+Multiple.args = {
+	dateTime: new Date('1912-04-15T05:18Z').getTime(),
+	format: 'date-only',
+	children: (
+		<>
+			<span data-o-date-printer></span>{' '}
+			<span
+				data-o-date-printer
+				data-o-date-format="time-ago-limit-4-hours"></span>{' '}
```
notlee, line 114: Hmm I wonder if this is an opportunity to rethink o-date. We basically want nested oDates so they share a date but show different formats. Can we pass props through to children from a parent date at all? FYI I'm about to run away for the day and haven't thought this through – just a mind dump :D

notlee (commented):
Thank you this is looking really great! :D It's super satisfying to see separate demos with a date control

components/o-date/stories/date.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+};
+
+export const TodayOrYesterday: DateStory = Template.bind({});
+TodayOrYesterday.args = {
+	format: 'today-or-yesterday-or-nothing',
+	description: 'using the o-date-format option',
+};
+
+export const FourHourLimit: DateStory = Template.bind({});
+FourHourLimit.storyName = '4 Hour Limit';
```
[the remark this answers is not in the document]
ivomurrell: I figured that the point of the demo was to show what happened when times fall out of the 24 or 4 hour limit but it's definitely a bit confusing!

components/o-date/stories/date.stories.tsx
ivomurrell: If the children are components themselves and the parent date is wired up properly then sure! Though not sure how that works when all the date logic is framework agnostic?

components/o-date/src/tsx/date.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export interface DateProps {
+	dateTime: string;
+	format?: DateFormat;
+	fallback?: string;
```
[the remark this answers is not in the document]
ivomurrell: Your call! Made the fallback prop when I was hoping we could move away from passing children entirely with a different abstraction for the copy options though that didn't work out. Thought it might still be useful as an abstraction and as a prompt to fill it in

components/o-date/stories/date.stories.tsx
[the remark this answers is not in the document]
notlee: True. I figured it would be better to see something then use the controls to try out the limit behaviour, but I can see an argument either way! I'll leave it to your judgement

components/o-date/src/tsx/date.tsx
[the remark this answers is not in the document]
notlee: >Thought it might still be useful as an abstraction and as a prompt to fill it in

Hmm good shout on the prompt, but maybe that would be better left to documentation. The downside of supporting both methods of providing a fallback is we need to document the differences and priority fallback vs. children.

components/o-date/stories/date.stories.tsx
notlee: I'm not sure either and would need to play around 😬 Is it possible to iterate `children` as a standard array, tell if they're another oDate component, and bind new props to them without anything React-y? That sounds wild I'm going to assume not.

I guess another more or less gross approach could be to provide a separate JSX template for printer elements? At least encapsulates the data attributes and would have the same format prop as `ODate`

ivomurrell: Nah, `children` is meant to be an opaque object so you need to use functions like [`React.Children.map`](https://reactjs.org/docs/react-api.html#reactchildrenmap) or [`React.Children.toArray`](https://reactjs.org/docs/react-api.html#reactchildrentoarray) to get it to look like an array. And there might be some minification that goes on in production that means the component names aren't stable, though idk enough about the frontend to say!

But yeah good idea, I'll make some printer JSX templates.

notlee (approved):
nice work, thanks :D

PR #732, opened 2022-05-17
Add stories for o-icons demos

ivomurrell:
All the different icons are selectable from a drop-down menu instead of all being rendered at once. The list of options is derived from the `fticons` imageset.

Custom icons are a little weird as they're just a span calling with a CSS class set, and the real work happens in the Sass. It doesn't seem worth creating a TSX component for that. It's also a bit harder to see what's going on under the hood compared to other components as we surface the generated HTML in a Storybook addon, but not the CSS. You can still easily inspect the code in your browser though, of course.


components/o-icons/src/tsx/icon.tsx
```diff
@@ -0,0 +1,9 @@
+export interface IconProps {
+	// currently no way to get a const type from a JSON import in TypeScript so
+	// let's stick with a string type for now
+	icon: string;
+}
```
chee, line 5: not for this PR but we could generate this `interface` as part of a build step

components/o-icons/src/tsx/icon.tsx
```diff
@@ -0,0 +1,11 @@
+import {IconName} from './data';
+
+export interface IconProps {
+	// currently no way to get a const type from a JSON import in TypeScript so
+	// let's stick with a string type for now
+	icon: IconName;
```
chee, line 6:
```suggestion
	icon: IconName;
```

components/o-icons/src/scss/_icon-list.scss
```diff
@@ -64,7 +64,5 @@ $o-icons-list: (
 	video,
 	warning-alt,
 	warning,
-	written-text,
-	eye-open,
-	eye-closed
+	written-text
```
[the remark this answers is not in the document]
chee: 👁️👀👁️

notlee:
Just done a little rebase there.

notlee (changes requested):
Sorry almost merged but I see chee noticed something happened to the eye icons. When they're restored I think this is good to go :) 

https://github.com/Financial-Times/origami/pull/732/files#r874880314

components/o-icons/stories/icons.scss
```diff
@@ -0,0 +1,14 @@
+@import "@financial-times/o-icons/main";
+@include oIcons();
+
+.demo-icon-custom {
+	@include oIconsContent('lock');
+}
+
+.demo-icon-custom-color {
```
notlee, line 8: Please delete the custom colour story! Sorry I didn't think to mention. We decided to only show demos in storybook that are actually approved by design and part of the FT design system, which means we don't show icons with random "this is possible" colours

components/o-icons/src/scss/_icon-list.scss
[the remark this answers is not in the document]
ivomurrell: This script uses the imageset from https://github.com/Financial-Times/fticons (which supposedly replaces `o-icons`...? huh) which doesn't include the newly added 👁️ icons. Maybe we should move away from that data source for the script in the future?

components/o-icons/scripts/build-icon-list.cjs
```diff
@@ -3,16 +3,19 @@
...
 
 function runScript() {
-	let {images} = loadJson(`${__dirname}/../node_modules/@financial-times/fticons/imageset.json`);
+	let {images} = loadJson(
+		`${__dirname}/../../../node_modules/@financial-times/fticons/imageset.json`
+	);
 	images = images.filter(image => !iconBlacklist.includes(image.name));
 	writeDemoData(images, `${__dirname}/../demos/src/data.json`);
 	writeSassIconList(images, `${__dirname}/../src/scss/_icon-list.scss`);
+	writeTsInterface(images, `${__dirname}/../src/tsx/data.ts`);
```
notlee, line 18: 🙌

PR #737, opened 2022-05-17
backstage: correct redirects, https://github.com/Financial-Times/orig…

notlee:
…ami/pull/736


chee (approved):
lol! this is very true ✅

PR #746, opened 2022-05-30
Add stories for o-labels demos

notlee:
Still need to make a nice interface for controlling dates in indicator
and timestamp labels (mostly in the Storybook UI but I also need to
confirm whether the JSX template is too restrictive or not by forcing
users to use o-date.)


notlee:
Nice, I think we can make the Stories more discoverable for users though. What do you think?

Switching the live indicator control is a bit confusing since it doesn't currently do anything visually. Let's hide the indicator control since we have a demo for each type of indicator?
Screenshot 2022-09-29 at 09 11 17

"Label" isn't a very helpful name for a story as it isn't clear what kinds of labels are included there – the user can't scan story names to find what they need. On reflection we should probably split this up to mirror the existing demos [Content States](https://registry.origami.ft.com/components/o-labels@6.2.2#demo-content-states), [Lifecycle States](https://registry.origami.ft.com/components/o-labels@6.2.2#demo-lifecycle-states), and... default(?) for core. [Service Tier States](https://registry.origami.ft.com/components/o-labels@6.2.2/?brand=internal#demo-service-tier-states); [Colour Palette States](https://registry.origami.ft.com/components/o-labels@6.2.2/?brand=internal#demo-colour-palette-states); [Support Status States](https://registry.origami.ft.com/components/o-labels@6.2.2/?brand=internal#demo-support-status-states), and... default(?) for internal. Guess that will mean needing to specify `argTypes` specifically in the Story
Screenshot 2022-09-29 at 09 11 02

notlee:
Looks like I can't leave a review @akomiqaia as I opened the PR but this looks good to me, approved! :shipit: Feel free to approve yourself and merge when you're ready

akomiqaia (approved):

PR #748, opened 2022-06-08
O messages tsx

akomiqaia:
I had to add `destroy()` as mentioned [here](https://github.com/Financial-Times/origami/pull/523).


components/o-message/src/tsx/message.tsx
```diff
@@ -0,0 +1,84 @@
...
+	if(!showCloseButton) {
+		dataAttributes['data-o-message-close'] = false;
+	}
+
+  return (
+		<InnerLayout
+			condition={hasInnerLayout}
+			// I had to add inline style to get the same feeling as in the registry.
+			// class demo-inner-message does not set those styles
+			wrapper={children => <div className="demo-inner-message" style={{margin: "auto", maxWidth: "400px"}}>{children}</div>}>
```
akomiqaia, line 64: This feels like a hack for displaying component. Maybe you have an idea why `demo-innner-message` did not apply styling?

components/o-message/stories/message.stories.tsx
```diff
@@ -0,0 +1,192 @@
...
+	title: 'Components/o-message',
+	component: Message,
+	decorators: [withDesign, withHtml],
+	// These params are copied from banner.tsx. need to find corresponding design in figma
+	parameters: {
+		design: {
+			type: 'figma',
+			url: 'https://www.figma.com/file/MyHQ1qdwYyek5IBdhEEaND/FT-UI-Library?node-id=0%3A1489',
+		},
+		guidelines: {
```
akomiqaia, line 20: where do i find this for this component?

components/o-message/stories/message.stories.tsx
```diff
@@ -0,0 +1,192 @@
...
+  } as ComponentMeta<typeof Message>;
+
+const Story = args => {
+	useEffect(() => {
+		let messages = javascript.init(null, null)
+		return function cleanup() {
+			messages = Array.isArray(messages) ? messages : [messages];
+			messages.forEach(message => message.destroy());
+		};
+	}, [args.showCloseButton])
```
akomiqaia, line 35: I copied this code from `Banner.tsx` and it works. I get the idea but `.destroy()` seems to be unnecessary. since message does not have function that destroys it. do we know what happens if we don't use `useEffect()` and just initialise javascript?

components/o-message/stories/message.stories.tsx
chee: don't worry about this right now, you can just set this to the empty object

components/o-message/src/tsx/message.tsx
chee, line 64: these .tsx templates are for public consumption. if there is something we want specific to the demos, then that should be defined in ◌́message.stories.tsx◌́ and not be part of the component itself.

nothing with a demo class should exist in the exported component.

the reason this is not setting the styles is that class is defined here: https://github.com/Financial-Times/origami/blob/main/components/o-message/demos/src/demo.scss

and that file is not imported.

components/o-message/src/tsx/message.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+
+export function Message({
+	type = 'alert',
+	state = 'success',
+	content = undefined,
+	primaryAction = undefined,
+	secondaryAction = undefined,
+	hasInnerLayout=false,
```
chee, line 50:
```suggestion
	inner=false,
```

let's make this the simpler `inner` to match the modifier class name

components/o-message/src/tsx/message.tsx
```diff
@@ -0,0 +1,84 @@
...
+}: MessageProps) {
+	const dataAttributes = {};
+
+	if(!showCloseButton) {
+		dataAttributes['data-o-message-close'] = false;
+	}
+
+  return (
+		<InnerLayout
+			condition={hasInnerLayout}
```
chee, line 61: i don't think we need to do this, setting `inner` should only add the `o-message--inner` modifier to the `o-message` element

components/o-message/src/tsx/message.tsx
```diff
@@ -0,0 +1,84 @@
...
+	detail: string;
+	hightLight?: string;
+	additionalInfo?: string;
+}
+
+type MessageAction = {
+	copy: string;
+	url: string;
+	openInNewWindow?: boolean
+}
```
chee, line 13:
```suggestion
type MessageAction = {
	label: string;
	url: string;
	openInNewWindow?: boolean
}
```

i think we've used the name `label` in other components for this

components/o-message/src/tsx/message.tsx
```diff
@@ -0,0 +1,84 @@
+import { Fragment } from 'react';
+
+type MessageContent = {
+	detail: string;
+	hightLight?: string;
```
chee, line 5:
```suggestion
	highlight?: string;
```

components/o-message/src/tsx/message.tsx
chee, line 13:
```suggestion
type MessageAction = {
	text: string;
	url: string;
	openInNewWindow?: boolean
}
```

actually, let's make it `text` here to match [the javascript api of this component ](https://registry.origami.ft.com/components/o-message@5.2.2/readme?brand=core#options)

components/o-message/src/tsx/message.tsx
chee: i answered this elsewhere, but it's because that class is defined in the component's demo.scss

components/o-message/stories/message.stories.tsx
chee:
```suggestion
	useEffect(() => void javascript.init(), [])
```

this is probably fine

components/o-message/src/tsx/message.tsx
chee: we also want to avoid importing that file. this inner-message wrapper business should probably be moved to the stories file, and if you find you need css for the container you can add it in the sass file imported by the story

akomiqaia:
@chee your review was very useful and updated code according to your recommendations.

notlee:
FYI I just rebased on main so we can build the internal brand storybook for this branch

components/o-message/src/tsx/message.tsx
```diff
@@ -0,0 +1,88 @@
...
+		| 'warning-light'
+		| 'inform'
+		| 'inform-inverse';
+	content: MessageContent;
+	primaryAction?: MessageAction;
+	secondaryAction?: MessageAction;
+	inner?: boolean;
+	showCloseButton?: boolean;
+	autoOpen?: boolean;
+	parentElement?: string;
```
notlee, line 30: I think we can delete these 2 options? It looks like they are not needed for the template and are only relevant when creating a message with the JavaScript API, without existing DOM elements.

components/o-message/src/tsx/message.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+type MessageAction = {
+	text: string;
+	url: string;
+	openInNewWindow?: boolean;
+};
+
+type MessageProps = {
+	type: 'action' | 'alert' | 'notice';
+	state:
```
notlee, line 15: Not all states are supported by all brands and we should reflect that in the Storybook / typescript definitions. For example `warning` states are only supported for the internal brand and visually breaks when selected using a core brand storybook build.

There's a table in the readme with details of which brands support which message types/layouts/states.
https://registry.origami.ft.com/components/o-message@5.2.2/readme?brand=internal#message-types

I've never had to solve this one before so don't have an approach to recommend! But I see Ivo has given it some thought in [his draft PR for o-labels](https://github.com/Financial-Times/origami/pull/746/files#diff-aec22530edf46a8b8cda01fdcdf1c6e1ff297ad05d794b2ee8b4951b1a4c5713R3) :) Maybe have a look at that and see what you think.
Screenshot 2022-06-14 at 16 17 56

components/o-message/src/tsx/message.tsx
```diff
@@ -0,0 +1,88 @@
...
+		| 'error'
+		| 'feedback'
+		| 'warning'
+		| 'warning-light'
+		| 'inform'
+		| 'inform-inverse';
+	content: MessageContent;
+	primaryAction?: MessageAction;
+	secondaryAction?: MessageAction;
+	inner?: boolean;
```
notlee, line 27: Similar to the brand issue mentioned before, not all message types support `inner`:
https://registry.origami.ft.com/components/o-message@5.2.2/readme?brand=internal#message-types

Is it possible to make this prop depend on the value in another prop? If not maybe we need a typescript type for each message type e.g `ActionMessageProps` which extends a base type e.g. `MessageProps`?

notlee (changes requested):
Nice work on this @akomiqaia! 👏 It's looking great but I noticed it's possible to create messages which aren't supported with a combination of brand/type/layout/state, we need to work on the Storys and typescript interface to prevent that. o-message is a more complicated one to start on than I realised!

components/o-message/src/tsx/message.tsx
```diff
@@ -0,0 +1,166 @@
...
+			<div className="o-message__container">
+				<div className="o-message__content">
+					<p className="o-message__content-main">
+						{content.highlight && <Highlight text={content.highlight} />}
+						{content.detail}
+					</p>
+					{inner && content.additionalInfo && (
+						<AdditionalInfo text={content.additionalInfo} />
+					)}
+					<div className="o-message__actions">
```
notlee, line 136: Both actions are optional, so we should remove the whole of `<div className="o-message__actions"> ... </div>` if no actions are provided.

components/o-message/stories/message-alert.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+		secondaryAction: {
+			text: 'Text link',
+			url: '#',
+			openInNewWindow: false,
+		},
+	},
+} as ComponentMeta<typeof AlertMessage>;
+
+const innerDecorator = Story => (
+	<div style={{margin: 'auto', maxWidth: '400px'}}>{Story()}</div>
```
notlee, line 33: Screenshot 2022-06-24 at 11 26 06

This demo markup is shown in the html tab, I expect we'll end up with users copying this and thinking all inner messages should have a 400px max width 😅 

- It would be great if we could find a way to exclude decorators from the html tab. Or failing...
- We might consider adding a demo class like `demo-inner-message-container` instead of inline styles, with the css in `stories/message.scss`. As least when users copy and paste this the class is obviously only for demo purposes and the demo styles won't apply in their project

notlee (commented):
Nice this is great! Again nice spot on `includeStories`. I've picked up on 2 minor things which would be nice to address before we release if you don't mind? But this is good to go! 😄

components/o-message/stories/message-alert.stories.tsx
akomiqaia: I had this class before and removed it. But you are right I did not really think about others copping it with max-width. I will update the code again. Can investigate if we can exclude the decorators from html markup. But this HTML render is plugin written by origami right?

notlee: Yeah that's right, see `apps/storybook/addons/html/index.js`

PR #749, opened 2022-06-13
feat: o-share demos for storybook

akomiqaia:


components/o-share/stories/share.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+export const Horizontal: ComponentStory<typeof Share> = Story.bind({});
+export const Small: ComponentStory<typeof Share> = Story.bind({});
+Small.args = {
+	small: true,
+};
+export const Inverse: ComponentStory<typeof Share> = Story.bind({});
+
+Inverse.args = {
+	inverse: true,
+};
```
notlee, line 47:
```suggestion
};
Inverse.parameters = {
	origamiBackground: 'slate',
};
```

This will set the background colour for the story to slate by default :) so the inverse demo is visible

notlee (approved):
Nice work :D See comments for a couple of non-blocking suggestions

PR #752, opened 2022-06-15
fix: o-tooltip show close button when opened on construction

notlee:
>A previous change made it so the close button doesn’t show if a tooltip is shown on hover or focus, but does show on click. It accidentally made it so if a tooltip is open by default on page load there is no close button, when there should be.

Relates to the following PR:
https://github.com/Financial-Times/origami/pull/419


notlee:
@fenglish Thanks! Chee is ill today and @akomiqaia is new ✨ so I think you probably know Origami better than the team today, feel free to approve! 🙈 If not @akomiqaia please take a look 😄

fenglish (approved):

PR #759, opened 2022-06-17
fix: o-fonts, support self hosted fonts

notlee:
The latest o-fonts major updated `$o-fonts-path` to point to
v3 of the Origami Build Service. This required setting font file
names in a query parameter, rather than sending a request for a
file explicity.

`https://www.ft.com/__origami/service/build/v3/font?version=1.12&font_name=MetricWeb-Regular&font_format=woff2&system_code=origami`

vs.

`https://www.ft.com/__origami/service/build/v2/files/o-fonts-assets@1.7.0/MetricWeb-Regular.woff`

This meant users could no longer set `$o-fonts-path` to self host fonts. 
This is something Financial-Times/ft-app does to support offline fonts.

As a fix this commit deprecates `$o-fonts-path` and replaces it with
`$o-fonts-build-service-path`, to specify a Build Service url to load
fonts from – this could be customised to point to a different version
of o-fonts on the Build Service, or a self hosted version of the Build
Service, but is unlikely to be used and perhaps could be removed in a
future major version. `$o-fonts-self-host-path` has been added to
mimic the previous behaviour of `$o-fonts-path` for self hosted fonts.

### DIFF
--- a/components/o-fonts/README.md
+++ b/components/o-fonts/README.md
@@ -204,7 +204,7 @@ _Note: If your project has multiple Sass entry points call `oFontsVariantsInclud
 
 Note: font files are contained in a separate, private repository ([o-fonts-assets](https://github.com/Financial-Times/o-fonts-assets)).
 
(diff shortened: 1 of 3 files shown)


akomiqaia (approved):
looks good to me but I might not know much about this

chee (approved):
this seems fine to me.

PR #767, opened 2022-06-23
o-loading story

chee:
This is written as an mdx file! how exciting!

One of the demos for o-loading was an example of how to use it accessibly, so i've written it as documentation instead.

Actually this way of creating the stories is really nice and I kind of want to use it for all of them.

image

NOTE FOR RABBIT: before merging this, check if the dark background is set up correctly on the light loader stories.


components/o-loading/src/tsx/loading.tsx
```diff
@@ -0,0 +1,10 @@
+export interface LoadingProps {
+	size?: 'mini' | 'small' | 'medium' | 'large';
+	theme?: 'light' | 'dark';
+}
+
+export function Loading({theme, size}: LoadingProps) {
+	return (
+		<div className={`o-loading o-loading--${theme} o-loading--${size}`}></div>
```
notlee, line 8: Size and theme are optional so we might end up with some `o-loading--` class names applied here

components/o-loading/stories/loading.stories.mdx
```diff
@@ -0,0 +1,121 @@
...
+		<Loading size="medium" theme="light" />
+	</Story>
+	<Story name="Light large">
+		<Loading size="large" theme="light" />
+	</Story>
+</Canvas>
+
+## Accessibility considerations
+
+When using the loader, you should mark the region it relates to as busy with
```
notlee, line 53: Very exciting! A much better way to communicate this then the old demo 😄 I'm glad to see a start to mdx docs. How did you imagine we'd split content across docs/readme? I had imagined docs being more about design / usage guidelines. This is interesting as it's a technical implementation detail which usually live in the readme. Still about context and usage within an app though. I suppose the readme will get pared back to summary / references to other places / api notes

notlee (changes requested):
Very exciting! 🙌 

Sadly the html / controls tabs aren't working at the moment:

Screenshot 2022-06-24 at 11 37 41
Screenshot 2022-06-24 at 11 45 00

It's be great to configure a slate background for the light demos too!

components/o-loading/stories/loading.stories.mdx
chee: the docs in the `GUIDANCE` subsection will be about design, accessibility and usage guidelines (the design system).

the docs in the `COMPONENTS` subsection will be specific stuff about the component. The README is something that will be seen on npmjs and on github. It might not even appear in the Storybook. It might be generated _from_ the Storybook.

components/o-loading/src/tsx/loading.tsx
chee: in the readme it says 

> you will need to supply the theme and the size modifiers in your markup

so i will remove the question marks from size and theme. they aren't optional

components/o-loading/stories/loading.stories.mdx
chee: does that make sense?

chee:
@notlee

notlee (approved):

PR #770, opened 2022-06-28
backstage: remove the whitelist function & `TRACKING_ATTRIBUTES`

MD-L:
### Remove the whitelist function & `TRACKING_ATTRIBUTES`

**Context:**

The Data Insight team have provided us (Apps team) new tracking attributes to add to the FT app:

1. `context.content.uuid` (which would be the same as `context.content.id`)
2. `context.content.asset_type` (which would be `article` if `audioSubtype` is `amy`, or `audio` if it is a `podcast`).

We currently rely on [o-audio](https://github.com/Financial-Times/origami/blob/main/components/o-audio/src/js/tracking.js), and cannot add any new tracking attributes on our side, as there is a whitelist function that only allows attributes from `TRACKING_ATTRIBUTES`.

**In this PR:**

- I removed this whitelist check of `trackingProperties`, as well as the removal of the  tracking attributes list (all within [L56-86](https://github.com/Financial-Times/origami/blob/main/components/o-audio/src/js/tracking.js#L56-L86)) .
- Then, on [L91](https://github.com/Financial-Times/origami/blob/9f398471d63c9d9231b23d5233494d2af3009d87/components/o-audio/src/js/tracking.js#L91) - `this.trackingProperties = trackingProperties` instead.


notlee:
Thanks for the PR and descriptive PR message! This looks good to me. Before I approve could you update this commit message to be `fix:`instead of `backstage:` please? 

Backstage won't trigger a release when we merge, it's intended for changes which aren't user facing. 

Fix will [trigger a patch release](https://github.com/Financial-Times/origami/blob/main/CONTRIBUTING.md#commit-type-prefixes) when we merge 🎉  we make [semver releases](https://semver.org/#summary)

MD-L:
> Thanks for the PR and descriptive PR message! This looks good to me. Before I approve could you update this commit message to be `fix:`instead of `backstage:` please?
> 
> Backstage won't trigger a release when we merge, it's intended for changes which aren't user facing.
> 
> Fix will [trigger a patch release](https://github.com/Financial-Times/origami/blob/main/CONTRIBUTING.md#commit-type-prefixes) when we merge 🎉 ( we make [semver releases](https://semver.org/#summary))

Thanks @notlee, I have changed the title of the PR now... although I'm not sure how to properly change my commit title 😅

MD-L:
> @MD-L I'd recommend `git commit --amend`, to change the message, then `git push --force` to overwrite with your commit here in Github :)

Thanks again @notlee I think it's corrected now :)

GlynnPhillips (approved):
This also looks good to me and +1 to Lee's comment about good descriptions, thank you

notlee (changes requested):
Thanks @MD-L! Sorry I didn't notice before but a test is failing because of this change. Would you mind updating the test to make sure custom event attribute _can_ be set?
https://github.com/Financial-Times/origami/blob/main/components/o-audio/test/tracking.test.js#L241-L246

FYI to run tests locally use `npm run test -w components/o-audio`

MD-L:
Updated! Please let me know if anything else is needed :)

notlee (approved):
Thanks @MD-L! 🎉 :shipit:

PR #771, opened 2022-06-29
fix: alternate to a lighter background

### LINKED ISSUE #743: Accessibility Error , insufficient contrast color on links in o-table.
When running accessibility test with PA11Y in next article we are having the error below. Please review it when you can.

The problem is that links on table that are in darker rows do not have enough contrast with the background.

```
Errors in https://ft-next-arti-ci-371-upd-wbz3kc.herokuapp.com/content/3f8a4ee9-0c0b-45b5-8fb6-4766d947a5c4:

 • This element has insufficient contrast at this conformance level. Expected a
   contrast ratio of at least 4.5:1, but text in this element has a contrast
   ratio of 4.14:1. Recommendation:  change text colour to #066f79.

   (#site-content > div:nth-child(3) > div:nth-child(2) > div >
   div:nth-child(20) > div > div:nth-child(1) > table > tbody > tr:nth-child(2)
   > td:nth-child(1) > a)
```

image

chee:
this is more accessible when there is a link in the text of the row

closes #743


notlee (approved):
~WHEAT 60!~
WHITE 60!

PR #773, opened 2022-07-08
feat: o-tracking, add test mode minus debug logs

CLTPayne:
#### Context:
* The Apps team sends non live events to a splunk index specifically for test data. Thus we need to ensure our development events have `is_live: false`. 
* In old versions of o-tracking we were able to set `is_live` in the initialisation config. This is no longer possible -  https://github.com/Financial-Times/origami/commit/5b69c97aa5d015046b26fc7d95aaa3b3852bf462
* The App started using the intended `test` config option but developers are finding the o-tracking debug logs particularly noisy (and our `ft-app` console can already be a little noisy sometimes 😬) 
* This option was discussed https://financialtimes.slack.com/archives/C02FU5ARJ/p1657105962524879. 

#### In this PR:
* I've added an additional config option to trigger `is_live: false` while leaving the `test` implementation as intended. 
* I found naming this new config option extremely hard. Very open to better names or other implementation ideas!?

#### Tests:
Updated for this new config option

### DIFF
--- a/libraries/o-tracking/README.md
+++ b/libraries/o-tracking/README.md
@@ -90,6 +90,18 @@ const config = {
 oTracking.init(config);
 ```
 
+o-tracking also has a `test-data` mode to mark events as test events without the debugging information in the console:
+
+```js
+import oTracking from '@financial-times/o-tracking';
+
+const config = {
+    test_data: true, // Mark the events as test events without extra debug logging.
+    ...
+};
+oTracking.init(config);
+```
+
 #### Methods
 
 ##### oTracking.init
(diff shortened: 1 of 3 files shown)


rowanbeentje (approved):
The naming here is definitely the trickiest part 😅 I think `test_data` works well though - complements the existing `test` config, and I think the docs explain the differences well.

Thank you!

Over to the Data team... 😁

notlee (approved):

FranciscoRuiz-FT (approved):
Hey @CLTPayne, sorry for the delay on this I am not totally familiar with all the details of this library so I needed dedicated time to check if this change could have some side effects but it is looking good to me. Thanks for adding the tests and updating the documentation.

About the config name, I agree, it is really tricky and I cannot think about something better, actually, I think the current `test` should be more like `debug` and this one you are proposing could be `test` or a specific log level for the `debug`, but that's something we can address in the future. For now, I believe your proposal looks like the way to go.

CLTPayne:
Thanks for the reviews all!

PR #775, opened 2022-07-13
Storybook/o teaser v3

akomiqaia:
Closing [original](https://github.com/Financial-Times/origami/pull/762) PR in favour of this PR


components/o-teaser/src/tsx/custom-slot.tsx
```diff
@@ -0,0 +1,16 @@
+import DOMPurify from 'dompurify';
+import { CustomSlot } from './Props';
+
+const render = (action: JSX.Element | string) => {
+	// Allow parent components to pass raw HTML strings
+	if (typeof action === 'string') {
+		return <span dangerouslySetInnerHTML={{__html: DOMPurify.sanitize(action)}} />;
```
notlee, line 7: With the aim to create more generic JSX templates we can't keep `dangerouslySetInnerHTML` as that's a React specific api. 🤔 Guess we update this to only allow a JSX element and to avoid a breaking perhaps have the React wrapper (x-teaser / our storybook demo) use [html-react-parser](https://github.com/remarkablemark/html-react-parser) to convert arbitrary html to an element

components/o-teaser/src/tsx/title.tsx
```diff
@@ -0,0 +1,48 @@
...
+					className: 'js-teaser-heading-link',
+					'aria-label': ariaLabel,
+				}}>
+				{displayTitle}
+			</Link>
+			{indicators && indicators.accessLevel === 'premium' ? (
+				<span>
+					{' '}
+					<span className={premiumClass}>Premium</span>
+					<span className="o-normalise-visually-hidden">&nbsp;content</span>
```
notlee, line 43: We might want to replace `o-normalise-visually-hidden` with a different o-teaser specific class, or document in the readme that oNormalise must be output by the project before including o-teaser (existing component do the former)

components/o-teaser/stories/teaser.stories.tsx
```diff
@@ -0,0 +1,141 @@
...
+	controls: {exclude: [...controlsToExclude, 'metaPrefixText', 'altTitle']},
+};
+
+export const Opinion: ComponentStory<typeof Teaser> = Story.bind({});
+Opinion.args = opinionArgs;
+Opinion.argTypes = argTypes;
+
+Opinion.parameters = {
+	controls: {
+		exclude: subtractArrays(controlsToExclude, ['showHeadshot', 'headshot']),
```
notlee, line 66: Non-blocking suggestion: It might be more clear which controls are excluded if we copy/paste the array for each story which uses `subtractArrays`? It doesn't look like there are too many

components/o-teaser/src/tsx/link.tsx
```diff
@@ -0,0 +1,29 @@
...
+	customElements?: any;
+	url?: string | undefined;
+	attrs?: {
+		'data-trackable'?: string;
+		'aria-label'?: string;
+		'aria-hidden'?: 'true' | 'false' | undefined;
+		className?: string;
+		tabIndex?: number;
+	};
+	children?: React.ReactNode[] | React.ReactNode | string;
```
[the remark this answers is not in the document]
akomiqaia: I think you are right and we shuold use `JSX.Element` instead of `React.ReactNode`.  

React node itself is a representation of the virtual DOM. So ReactNode is the set of all possible return values of a component. 

```ts
type ReactNode =
  | ReactChild
  | ReactFragment
  | ReactPortal
  | boolean
  | null
  | undefined;
```

`JSX.Element` is more generic type. The key difference is that props and type are typed as any in JSX.Element.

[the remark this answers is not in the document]
akomiqaia: [PR that fixes](https://github.com/Financial-Times/origami/pull/775/commits/366555c6983cea93199980fdb1515066fa32913f)

components/o-teaser/src/tsx/title.tsx
notlee: Yeah, probably just `o-teaser-visually-hidden`

components/o-teaser/stories/teaser.stories.tsx
```diff
@@ -0,0 +1,259 @@
...
+export default {
+	title: 'Components/o-teaser',
+	component: Teaser,
+	decorators: [withDesign, withHtml],
+	parameters: {},
+} as ComponentMeta<typeof Teaser>;
+
+const Story: ComponentStory<typeof Teaser> = args => {
+	console.log();
+
```
notlee, line 35: whoops! :D 
```suggestion
```

components/o-teaser/stories/teaser.stories.tsx
```diff
@@ -0,0 +1,259 @@
...
+	if (args.showCustomSlot) {
```
notlee, line 36:
```suggestion
	if (typeof args.showCustomSlot === 'string') {
```

It worked!? Nice! I think we only want to parse if `showCustomSlot` is a string though? It could be a TSX Element.

notlee (approved):

PR #778, opened 2022-07-19
Storybook/o header

akomiqaia:
Template is based on [dotcom header](https://github.com/Financial-Times/dotcom-page-kit/tree/main/packages/dotcom-ui-header) API. We copied storybook data from dotcom repository using following commands: 

```bash
git fetch https://github.com/Financial-Times/dotcom-page-kit.git main:tmp-storybook-data
git checkout tmp-storybook-data
git subtree split --prefix=packages/dotcom-ui-header/src/__stories__/story-data -b tmp-dotcom-header-storybook-data
git checkout storybook/o-header
git subtree add -P components/o-header/stories/storybook-data tmp-dotcom-header-storybook-data

```

Few things came up while working on this branch and it's compatibility to other team's registries. I will document this separately but will leave my notes with this PR too: 

Origami tsx template rely on JSX runtime and importing tsx templates will heavily depend on your setup. You will need typescript ^v4.1. Which is necesary to specify “jsx”: “react-jsx” in tsconfig.

[React transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) enables developers not to import react and on runtime it added jsx wrapper instead of React.createElement. Which means we can write JSX templates and make them agnostic to a framework. Which has proven to be quite difficult but it is possible and it will require newer version (^v4.1) of typescript installed on your repo. If you are using Babel you will need @babel/core version ^7.9 to use react-jsx runtime.

If you get some errors that might say that react-jsx runtime does not exists you can try to resolve alias in your webpack configuration:

```js
config.resolve.alias = {
   'react/jsx-runtime': require.resolve('react/jsx-runtime')
 }
```
For vue3 users this process should be seamless if your project is already setup to support TSX/JSX. If not have a look at Vue’s tutorial to manually implement jsx support.


components/o-header/src/tsx/user.tsx
```diff
@@ -0,0 +1,21 @@
...
+		<div className="o-header__row o-header__anon" data-trackable="header-anon">
+			<ul className="o-header__anon-list">
+				{userNavItems.map((item, index) => (
+					<li className="o-header__anon-item" key={`link-${index}`}>
+						<a
+							className="o-header__anon-link"
+							href={item.url || undefined}
+							data-trackable={item.label}>
+							{item.label}
+						</a>
```
akomiqaia, line 15: @notlee for example this component uses `data-trackable` on parent element and then adds additional `data-trackable` on each item based on their label. 

Nesting of UsersActionsNav is following: 
DefaultHeader > HeaderWrapper > Header > UsersActionsNav

if `DefaultHeader` is taking `extraHeaderProps` prop how do we identify which data-attribute goes where? We might have a case where maybe we need to pass data attribute to `UsersActionsNav` sibling component. 

It seems like we need to give users an option to define attributes per component and `extraHeaderProps` can be a complex object that will pass down all the necessary attributes this way. example: 

```js
const extraHeaderProps = {
	UsersNavAction: {
		'header-anon': {
			attributes: {'data-trackable': 'header-anon'},
		},
		'anon-link': {
			attributes: label => ({'data-trackable': label}),
		},
	},
	AnotheComponent: {
		attributes: {},
	},
};

```

notlee: Ooo this is an interesting problem. 🤔 Here comes a bit of a brain dump.

I like the approach for header owned elements like `header-annon` but feel less good about `anon-link`. It's kinda tricky to mentally map `extraHeaderProps.UsersNavAction['anon-link']` to the nav data / which element will get the attribute. I'm wondering if it's going to be hard to document the callbacks for `extraHeaderProps`.

```tsx
// get data
const data = fetch('https://');

// describe attributes
const extraHeaderProps = {
	UsersNavAction: {
		'header-anon': {'data-trackable': 'header-anon'},
        // if we go with this maybe passing the item would be 
        // more flexible for other usecases
		'anon-link': item => ({'data-trackable': item}),
	},
	AnotheComponent: {},
};

// the rest
const props = {
	data,
	extraHeaderProps,
	variant: 'whatever',
	showSubNavigation: true
};

<DefaultHeader {...props} />
```

Wouldn't it be nice to have an attributes prop on `NavMenuItem`, which keeps all data related to a nav item together, like [disableTracking](https://github.com/Financial-Times/dotcom-page-kit/blob/2102d0305b9e8874b818aa2feeb20662d2f88c65/packages/dotcom-ui-footer/src/components/partials.tsx#L13)? 

Problem with that `NavMenuItem` I believe derives from the `data` prop, which is passed through straight from an api request, I think. We'd have to make users map over `data['navbar-right-anon'].items` to add an `attributes` prop which... I also don't love 😅 

```tsx
// get data
const data = fetch('https://');

// add attributes
data['navbar-right-anon'].items.map(item => {
    item.attributes = {'data-trackable': item.label};
    return item;
});

data['navbar-right-anon'].attributes = {'data-trackable': 'header-anon'};

// the rest
const props = {
	data,
	variant: 'whatever',
	showSubNavigation: true
};

<DefaultHeader {...props} />
```

Is that an option you considered? Any thoughts on pros / cons? Maybe other ideas?

akomiqaia: I actually like your idea. But problem I can think of is elements can be nested in many divs or any other html tag and users might need very specific html elements to get targeted? The problem is that we just don't know how and where attributes will be used. 

we can by default support item related attributes to be picked up from data object but I feel like it might be better if teams will write custom functions and use query selectors to apply attributes where needed. something like this maybe: 

```jsx
const wrapper = props => {
	useEffect(() => {
		const headerAnon = document.querySelector('.o-header__anon');
		headerAnon.setAttribute('data-trackable', 'header-anon');

		const links = document.querySelectorAll('.o-header__anon-link');
		links.forEach((link) => link.setAttribute('data-trackable', 'LABEL'));
	}, []);
	return <DefaultHeader {...props}></DefaultHeader>;
};

```

notlee: I quite like that, it does seem more clear and readable. One thing we should work towards though is the ability to make markup changes without needing to coordinate a major release and I think this approach would hinder that 🤔  We could get there with these tsx templates and by eventually providing an Origami React library to wrap components. If we did that the React library wouldn't need a major release for markup changes, which is currently one of the most difficult and most time consuming changes to rollout. However if we take this approach and encourage users to manipulate the DOM outside component templates we'd forego the ability to make markup changes in a minor or patch release, as we may break people's query selectors.

notlee (approved):

Issue #617, opened 2022-02-07
Create Storybook UI to switch brands


notlee:
Component styles vary per brand. So do their features. For example the "editorial" demo in o-quote
is only shown when the core brand is selected. We should be able to filter demos / story controls by 
brand in storybook too.
https://registry.origami.ft.com/components/o-quote@5.2.1

notlee:
I wonder if this could be a dropdown in storybook or if we have some environment variable and publish multiple storybooks perhaps...

chee:
> or if we have some environment variable and publish multiple storybook builds perhaps...

indeed https://github.com/Financial-Times/origami/blob/18223a71cfd2fa285b420c85c28b0234eab3ee33/apps/storybook/config/main.js#L27

notlee:
Oh cool, I knew you investigated multiple brand support but wasn't sure where that ended. I guess we can use the same env to hide/show stories! Do you have a plan for rolling out a brand switcher or is it something that needs more thought?

chee:
We discussed this in November.

I spent like 2 days working on a way of making a drop down for brand but it proved to be very-difficult-to-impossible to do in a way that could work because the client side and server-side are different environments.

So then I elected to use an environment variable so we could do 3 builds of the Storybook.

A possible path forward is to have a drop down that links you to the current page on the other builds.

We can also possibly use [Storybook Composition](https://storybook.js.org/docs/react/sharing/storybook-composition) to bring the other brands into one Storybook.

But then we decided to leave the env var in place, postponing all that work until after design tokens because it would probably be easier to express this in a world with design tokens

notlee:
It's now possible to navigate to a storybook for all brands from origami.ft.com so closing this. We can improve this by having a switcher within the storybook in the future

PR #782, opened 2022-07-29
feat: add internal component system in storybook UI

akomiqaia:
Partially implements: #617 

To enable multi brand support in our storybook I decided to use [SB composition](https://storybook.js.org/docs/react/sharing/storybook-composition), which means that now each brand is running on it's own server and are referenced in main storybook deployment. Once we do first deployment of brand components we will need to update storybook config file to insert correct URLs. In production performance should not be issue but for local development there might be some issues since we need to run them on their separate servers (especially if we start supporting multiple brands). 

Other alternatives are:
- Use `[Vite](https://storybook.js.org/docs/react/builders/vite)` for dev server, that will be faster than `webpack` HMR but this won't solve an issue of multiple servers on local machine.  
- Use `[npx sb extract](https://storybook.js.org/docs/react/builders/vite)` and generate `stories.json` with other files that can be served from a server but we will loose hot reload. 
- Build [custom builder ](https://storybook.js.org/docs/react/builders/builder-api) for our use case but I found this one too complex and it might not be exactly what we need.


notlee (commented):
Because of the 3 issues below do you think composition is still the right choice for us? 

Instead of the wrapper storybook we could provide a landing page which points to each brand's storybook and potentially within each storybook have a way to switch to the other brand's storybook.

We might want to use composition in the future within a brand if our business case to expand Origami brand support goes well

PR #788, opened 2022-08-11
feat: add ts compiler for button and fix ts errors inside button.tsx

akomiqaia:
This PR should generate `react-lib` folder inside the component that should include tsx transpiled code. 

*Button compilation is for having proof of concept that react based teams can import compiled code without a lot of effort.*


notlee:
We investigated this issue further:

* Our aim was to help the platforms team replace their x-dash jsx with origami tsx.
* Publishing for React wouldn't help because they use x-engine to support React, Preact, and hyperons.
* We could compile tsx to jsx without assuming a runtime and could do this in the future as needed.
* Right now we're not doing that. The Spark team are adopting the tsx templates. Platforms would still need to do more work to use jsx – i.e. configure `@rollup/plugin-node-resolve`, set a Pragma function and include it globally

PR #790, opened 2022-08-16
alter o-buttons tsx to solve compile failures in stricter environments

chee:
rabbit saying hello

hello!

changes in this pr:

- remove `ButtonTemplate` (it fails to compile because of the lack of overlap in the union)
- add a `makeClassNames` function to reduce some of the duplication between LinkButton and Button created by removing ButtonTemplate
- attach the `onClick` handler ourselves to handle React runtime expecting the handler to be a `MouseEventHandler` rather than a function, this should always work without requiring a `import type {} from "react"` in the agnostic templates
- use `null` as the default value for `href` so we can do `href={href}` rather than `{...(href ? {href} : {}}`. functionally the same i think but looks a little prettier

### DIFF
--- a/components/o-buttons/src/tsx/button.tsx
+++ b/components/o-buttons/src/tsx/button.tsx
@@ -31,17 +31,7 @@ interface LinkButtonProps extends ButtonProps {
 	href: string;
 }
 
-function ButtonTemplate({
-	label,
-	type = 'secondary',
-	size = '',
-	theme,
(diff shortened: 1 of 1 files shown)


chee:
damn that little hedgehog. it just wants your money

notlee (approved):

notlee:
such a hungry hedgehog

PR #798, opened 2022-08-24
Run prettier on all the things

notlee:
We don't want Prettier changing files as we work on them as the diff becomes tricky to read. Let's run Prettier on everything within a commit that has no other changes.


notlee:
Fun. Looks like our Prettier and eslint config conflicts
https://github.com/Financial-Times/origami/runs/7997280953?check_suite_focus=true

akomiqaia:
oh this seems fun :D I thought all the components were copied from their original repos and for some reason assumed we did not want to touch them and let's why it was ignored. It always remained as a surprise to why in workspaces tsx template formmating worked.

notlee:
Out latest lint rules only run on changed files so this has also bought up other issues unrelated to Prettier. This PR needs more work.

notlee:
Closing in favour of https://github.com/Financial-Times/origami/pull/876

PR #800, opened 2022-08-30
feat: storybook landing page

akomiqaia:


apps/website/pages/storybook.html
```diff
@@ -0,0 +1,42 @@
...
+title: Origami Storybook
+description: Storybook for Origami components and services
+permalink: /storybook/
+layout: o-layout-landing
+---
+
+<div class="o-layout__hero o-layout__hero--muted o-layout-typography">
+	<h1>{{page.title}}</h1>
+	<p>
+		Storybook is allowing us to make the most of a large ecosystem of tools to
```
notlee, line 11: Instead of sell storybook here I think it would be more helpful to concisely explain to our users what Storybook is. Maybe something like this?

 ```suggestion
		Interactive component demos and documentation.
```

File: apps/website/pages/storybook.html [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+<div class="o-layout__main o-layout-typography">
+	<div class="o-layout__overview">
+		<div class="o-layout-item">
+			<p>
+				We’re writing JSX templates for our components for the first time, which
+				we may make public in the near future. For teams who also use JSX
+				templates, this could hugely reduce the complexity of implementing
+				Origami components in a project and make our most tricky kind of
+				migration (markup changes) simple when upgrading from one major version
+				of a component to another.
```
notlee, line 25: This focuses quite heavily on the tsx template benefit but doesn't explain what Storybook is for, or how it differs from the registry. Users who land on this page could be confused.

Things I'd expect users to learn:
1. What Storybook is (above)
1. When to view the registry vs. storybook 
1. That it's a work in progress, incomplete, but will replace the registry
1. Why (could link to`/blog/2021/11/05/newsletter/#storybook`.)
1. If in doubt, browse the registry (link)

What do you think?

apps/website/pages/storybook.html
```diff
@@ -0,0 +1,42 @@
...
+			</p>
+		</div>
+		<div class="o-layout-item">
+			<h2>Supported Brands</h2>
+			<li>
```
notlee, line 30: These`li` list items are missing a parent list element `ul`

apps/website/pages/storybook.html
```diff
@@ -0,0 +1,42 @@
...
+				<a href="https://origami.ft.com/storybook/brands/core/">
+					Core
+				</a>
+			</li>
+			<li>
+				<a href="https://origami.ft.com/storybook/brands/internal/">
+					 Internal
+				</a>
```
notlee, line 38: A relative link here means they'll work in the preview build! :D 
```suggestion
				<a href="/storybook/brands/core/">
					Core
				</a>
			</li>
			<li>
				<a href="/storybook/brands/internal/">
					 Internal
				</a>
```

PR #803, opened 2022-08-31
Ci 1158 alphaville article page branding

juanSanchezAlcala:
Changes in color needed on o-topper for new Alphaville branding.
[ticket](https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1316?modal=detail&selectedIssue=CI-1158)
image


components/o-topper/src/scss/_mixins.scss
```diff
@@ -41,6 +41,9 @@
 				}
 			}
 		}
+		.o-topper--branded.o-topper--color-matisse .o-topper__standfirst {
+			opacity: 1;
+		}
```
notlee, line 46: Thanks @juanSanchezAlcala! This PR looks good. I'd just like to query the necessity of this change with the design team. We may want to keep an opacity on the standfirst for consistency across the toppers. I've checked and with the existing opacity there is still enough contrast to meet WCAG level AA guidelines

notlee: right: this pr. left: with existing opacity 
Screenshot 2022-08-31 at 16 09 17

notlee:
```suggestion
```

Confirmed with Olga that it's good to drop this and stick to the standard opacity :)

notlee (approved):

PR #805, opened 2022-09-01
fix: upgrade n-map-content-to-topper version in o-topper

juanSanchezAlcala:
upgrade n-map-content-to-topper version in order to be able tu use alphaville branding
[ticket](https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1316?modal=detail&selectedIssue=CI-1158)

### DIFF
--- a/components/o-topper/package.json
+++ b/components/o-topper/package.json
@@ -22,7 +22,7 @@
     "watch": "bash ../../scripts/component/watch.bash"
   },
   "peerDependencies": {
-    "@financial-times/n-map-content-to-topper": "^1.5.0",
+    "@financial-times/n-map-content-to-topper": "^2.1.0",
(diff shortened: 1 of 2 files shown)


notlee:
This is a breaking change as we're updating a peer dependency to a new major version which may cause a dependency conflict for o-topper users.

In this case I don't think `n-map-content-to-topper` needs to be a peer dependency. Could you move it to `dependencies` to avoid any potential conflicts?

juanSanchezAlcala:
> This is a breaking change as we're updating a peer dependency to a new major version which may cause a dependency conflict for o-topper users.
> 
> In this case I don't think `n-map-content-to-topper` needs to be a peer dependency. Could you move it to `dependencies` to avoid any potential conflicts?

sure, done

notlee (approved):

PR #809, opened 2022-09-08
807 whitelabel dev server failing

akomiqaia:


notlee (approved):
Nice! Just some indentation to fix there before merging :) 

FYI this could be a `backstage` commit rather than `fix` as it only affects the story and we don't need to release a new version of the component.

notlee:
Helpful as always, Percy 👀

notlee:
@akomiqaia When we merge to main I think release please will look at all commits and release a patch because of the`fix` commits. We can avoid that if we squash and merge, which will squash all the commits into one new `backstage` commit

PR #828, opened 2022-10-06
533 o forms alert fix

### LINKED ISSUE #533: o-forms: ensure inline errors are assigned the 'alert' role - ID: DAC_Status_Message_01
# Issue
This is from page 96 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-forms@9.0.1&demo=checkboxes&system_code=origami-repo-data&brand=master>

The error message ‘Please select at least one option’ has been placed in a span below the input fields. For sighted users they are able to identify the error message correctly by the colour red but for screen reader users the message is not announced as an alert.
 
# Proposed Solution:
Ensure screen reader users are able to identify when errors have occurred in the checkboxes. In this case, the span which contains the error message should be assigned the role=”alert”. This will enable screen reader users to identify the error message correctly.

akomiqaia:
closes #533


notlee (approved):
I noticed that radio buttons and checkboxes had `required` set so, when making a de-selection, they would fail validation and their boarder would go red. They would not show an error message as we didn't configure one the demo json data. We didn't notice before because o-forms js was only initialised on the interactive demo. 

I made a commit to never make a multi-input field required in the demo.

notlee:
@akomiqaia To update the demos and readme in the registry we have to make a release. I'd suggest releasing as patch. You could add a new `fix` commit or squash and merge, writing a new `fix:` commit message.

PR #829, opened 2022-10-07
fix: o-banner, relative space close button position

notlee:
When `$o-spacing-relative-units:true` o-banner would output a large negative position for the close button due to... bad math in Sass. Instead we can use the `calc` function.


OBakir90 (approved):
LGTM! thanks for sorting that so quick

PR #832, opened 2022-10-10
backstage: add focus rings for toggle

### LINKED ISSUE #540: o-forms: focus state for o-toggle is too small - ID: DAC_Focus_Visible_01
# Issue
This is from page 122 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-forms@9.0.1&demo=toggle&system_code=origami-repo-data&brand=master>

The toggle button has a grey square to show users that focus has been placed on the element. This focus indicator is insufficient and may affect keyboard only and low vision users, who may find difficulty in identifying where the focus has been placed in their 
journey.

> “The highlighting does not appear correctly when my focus is on the first toggle, I thought there was no highlight but if you look in the left there is the slight grey highlight so it's going to the wrong place. Users would expect the highlight focus to be around the element fully so they can see where their focus is on the page.” 

# Proposed Solution:
Ensure the focus indicator is visible on the whole of the input. This will enable low-vision and keyboard alone users to distinguish where the focus is placed in their journey.
The focus indicator should also meet a minimum contrast ratio of 3.1 against the toggle input.

akomiqaia:
closes #540


components/o-forms/src/scss/_toggle.scss
```diff
@@ -60,6 +60,16 @@
...
 			}
+			&:focus-visible {
+				outline: none;
+				box-shadow: none;
+			}
+
+			&:focus + .o-forms-input__label::before {
+				@if oBrandIs('core') or oBrandIs('internal') {
+					outline: 4px solid oColorsByUsecase(focus, 'outline'); // oCoulorsByUsecase doesn't support whitelable brand
+				}
```
notlee, line 71: We can pass `null` to the 3rd [fallback parameter](https://registry.origami.ft.com/components/o-colors@6.4.2/sassdoc?brand=core#function-ocolorsbyusecase) to handle this.

```suggestion
				outline: 4px solid oColorsByUsecase('focus', 'outline', $fallback: null);
```

For the whitelabel brand that will output this, and the colour will be inherited:

```
				outline: 4px solid;
```

notlee (approved):

PR #833, opened 2022-10-10
Storybook/o forms radio box inputs

OBakir90:
Have tried to include [all demos](https://www.ft.com/__origami/service/build/v3/demo?component=o-forms%409.4.0&demo=radio-box&system_code=origami-repo-data&brand=core) in registry using same inputs listed [here](https://github.com/Financial-Times/origami/blob/main/components/o-forms/demos/src/data/radio-box.json)


akomiqaia:
Really nice work @OBakir90. Seems like you have done a lot of work. Just found few issues with storybook demos and added some suggestions at the end.

Screenshot 2022-10-10 at 16 21 07
In the demos it seems like user needs to click twice to switch the state on initial render. After that it works fine and immediately switches between states.

Screenshot 2022-10-10 at 16 20 50
Similar issue as in the first one. if I have `daily` selected and then I try to go to `monthly` first it jumps to `weekly` and then after another click it selects `monthly`.

Screenshot 2022-10-10 at 16 25 46
The error message is still present when one of the options is selected. Maybe we can switch classes based on selection state? 

Also just a general thought about the demos. In our [demo-box-styled-radio-inputs](https://registry.origami.ft.com/components/o-forms@9.4.0#demo-box-styled-radio-inputs) registry this demos are grouped and maybe we want similar thing to happen inside the storybook?

components/o-forms/stories/forms.stories.tsx
```diff
@@ -0,0 +1,349 @@
...
+		};
+	}, []);
+	return <Forms {...args} />;
+};
+
+export const BoxRadioButton: ComponentStory<typeof Forms> = FormsStory.bind(
+	{}
+);
+
+export const HiglightBoxRadioButton: ComponentStory<typeof Forms> = FormsStory.bind(
```
akomiqaia, line 32:
```suggestion
export const HighlightBoxRadioButton: ComponentStory<typeof Forms> = FormsStory.bind(
```

components/o-forms/stories/forms.stories.tsx
```diff
@@ -0,0 +1,349 @@
...
+		},
+		title: {
+			mainTitle: "Box Radio Button",
+			promptTitle: "This is a box radio button sample"
+		}
+	},
+
+}
+
+HiglightBoxRadioButton.args = {
```
akomiqaia, line 94:
```suggestion
HighlightBoxRadioButton.args = {
```

components/o-forms/src/tsx/forms.tsx
```diff
@@ -0,0 +1,103 @@
...
+				)}
+			</span>
+			<span className={addNewClassNames('o-forms-input o-forms-input--radio-box', field.modifiers, 'o-forms-input--')}>
+				{children}
+				{field.error && (
+					<span className="o-forms-input__error" >{field.error}</span>
+				)}
+				{field.state && (
+					<span className="o-forms-input__state " />
+				)}
```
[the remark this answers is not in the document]
akomiqaia: No problem at all and thanks for updating the code. But I think my previous comment was a bit confusing. Not all the span elements need a `role` attribute. `span` element has no semantic meaning and sometimes we add `role` attribute to provide meaning to it that is understood by screen readers. here is more info about [aria-roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles). [This PR](https://github.com/Financial-Times/origami/pull/828/files) only updates `error` and `state` span elements.

akomiqaia:
@OBakir90 This is looking super nice. And good work on storybook controls. Just one little thing to mention is `saved` control does not work on certain components and it seems like it was intended only for  `StateBoxRadioButton`. It will be better experience for users if `saved` control is disabled and gets shown only in `StateBoxRadioButton`. Here is doc on how to [disable controls](https://storybook.js.org/docs/react/essentials/controls#disable-controls-for-specific-properties).

components/o-forms/stories/forms.stories.tsx
```diff
@@ -0,0 +1,205 @@
...
+export const ErrorBoxRadioButton: ComponentStory<typeof Forms> =
+	FormsStory.bind({});
+
+export const StateBoxRadioButton: ComponentStory<typeof Forms> =
+	FormsStory.bind({});
+
+BoxRadioButton.args = {
+	inputs: [
+		{
+			type: 'radio',
```
notlee, line 48: love that it's possible to modify add more inputs in the demos :D

components/o-forms/src/tsx/forms.tsx
```diff
@@ -0,0 +1,139 @@
...
+	inputs?: inputs;
+	highlight?: boolean;
+}
+
+interface FormFieldProps {
+	children?: JSX.Element | JSX.Element[];
+	title?: title;
+	field?: field;
+	centered?: boolean;
+	saved?: boolean;
```
notlee, line 38: These props may need a little refactor before we make them public. For example:

I think only box-styled radio inputs support a state? Could the prop be on a more specific component than `FormField`?

Do we need `save` and `field.state`, or could 1 `state` property work (it wouldn't make sense to have saved = false and field.state = saving)? 

Do we need `field.modifiers` or could the modifiers be determined based on the `state` or other props?

notlee (changes requested):

KamranMuniree:
Thanks Orhan for your hard work on this. We will be closing this PR in favour of [this](https://github.com/orgs/Financial-Times/projects/83#card-86533701) which has been built on the foundations of your work.

PR #834, opened 2022-10-10
backstage: add aria-label and role=status for screen readers

### LINKED ISSUE #537: o-forms: save/saving state not accessible to screen reader users - ID: DAC_Status_Message_04
# Issue
This is from page 116 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-forms@9.0.1&demo=radio-box&system_code=origami-repo-data&brand=master>

The radio buttons with the inline saved state may be problematic for screen reader users as the content which displays the ‘saved’ or ‘saving’ states has been added using CSS. As there is no text alternative, screen reader users would be unable to view the state of the buttons in the same way as a sighted user.

 
# Proposed Solution:
Ensure that there is an alternative available for screen reader users such as a status message. Further information can be found in [Using role=status to present status messages](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22).

akomiqaia:
closes #537

### DIFF
--- a/components/o-forms/demos/src/interactive-suffix.mustache
+++ b/components/o-forms/demos/src/interactive-suffix.mustache
@@ -1,12 +1,12 @@
 <form action="" data-o-component="o-forms">
-	<label class="o-forms-field" for="suffix">
+	<span class="o-forms-field">
 		<span class="o-forms-title">
-			<span class="o-forms-title__main">Text input with suffix</span>
+			<label for="suffix" class="o-forms-title__main">Text input with suffix</label>
 		</span>
 
 		<span class="o-forms-input o-forms-input--text o-forms-input--suffix">
 			<input id="suffix" type="text" name="text" value="" required>
-			<button class="o-buttons o-buttons--secondary o-buttons--big">Go</button>
+			<button class="o-buttons o-buttons--secondary o-buttons--big">Submit</button>
 		</span>
-	</label>
+	</span>
 </form>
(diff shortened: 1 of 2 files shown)


components/o-forms/README.md
```diff
@@ -295,7 +295,7 @@ For a box-like, rectangular radio button replace the `o-forms-input--radio-round
 
 Box radio buttons may also, optionally represent a negative choose by applying the `o-forms-input__label--negative` modifier class to its label. This is useful for a yes/no input.
 
-Loading states are also supported by box radio buttons. We recommend using [`o-forms` JavaScript](#javascript) to add states to an existing form. However to add a state manually add an additional `o-forms-input__state` element and a state modifier class to the `o-forms-input` element, `o-forms-input--saving` or `o-forms-input--saved`.
+Loading states are also supported by box radio buttons. We recommend using [`o-forms` JavaScript](#javascript) to add states to an existing form. However to add a state manually add an additional `o-forms-input__state` element and a state modifier class to the `o-forms-input` element, `o-forms-input--saving` or `o-forms-input--saved`. To make loading states accessible for screen readers we recommend adding `role=status` and `aria-label="[STATE]"` attributes.
```
notlee, line 298: It's not clear which element to add the role/aria-label's too. How about this?

```suggestion
Loading states are also supported by box radio buttons. We recommend using [`o-forms` JavaScript](#javascript) to add states to an existing form. However to add a state manually add an additional `o-forms-input__state` element with `role=status` and `aria-label="[STATE]"` attributes. In addition, add a state modifier class to the `o-forms-input` element, `o-forms-input--saving` or `o-forms-input--saved`. 
```

notlee (approved):

PR #839, opened 2022-10-12
Update o-label tsx / stories and make public

notlee:
[fix: Update o-label tsx and stories](https://github.com/Financial-Times/origami/commit/bb10cb98518c0ad3871da9a5c73630878afde4bf)

- Create a new TSX component for each "kind" of label e.g. content;
support status; service tier. This makes it more clear which states
are part of a group and improve code completion to show the
different options for a kind of label.
- In the storybook for each label list the options as controls,
rather than show all labels in the story. Use a placeholder label
which matches the state. This makes it easier to find a label
for a specific purpose. For example the "Support Status Label"
story previous showed 2 identical looking labels at the same time
for active + maintained. Within inspecting the HTML it was not
clear that these were 2 different labels for different purposes
– they could diverge visually in the future, so it's important
to select the correct one.
Screenshot 2022-10-12 at 15 22 26
- Label copy is provided with a child rather than a `text`
prop. This arguably aligns more closely with html
`<BaseLabel>my label</BaseLabel>` instead of `<BaseLabel text="my label" />`
- Removes nested `time` elements from the indicator and time
label template.
- Time is provided as a prop rather than a child, as there are
multiple slotted children and it wasn't clear why `status` was
a prop and `timestamp` was not. The `status` prop has also been
made optional, it will now fallback to the indicator.
`<OIndicatorLabel indicator="live" timestamp={MyTimeComponent} />`
instead of
`<OIndicatorLabel indicator="live" status="live">{MyTimeComponent}</OIndicatorLabel>`

[docs: Remove nested time element from o-labels](https://github.com/Financial-Times/origami/commit/b003d2a7907b8df5294b44b60b26da52b4f63bf6)

[feat: o-labels, make tsx public](https://github.com/Financial-Times/origami/commit/6ada055893895cf0a7ad910d80fcc86eb0e345c0)


notlee:
I've sneakily made o-date tsx public too, as that's intended for use with the indicator label :)

components/o-labels/src/tsx/label.tsx
```diff
@@ -1,72 +1,101 @@
...
+	return SimpleLabel(props)
+}
+export function ColourLabel(props: ColourLabelProps): JSX.Element {
+	return SimpleLabel(props)
+}
+export function ContentLabel(props: ContentLabelProps): JSX.Element {
+	return SimpleLabel(props)
+}
+export function LifeCycleLabel(props: LifeCycleLabelProps): JSX.Element {
+	return SimpleLabel(props)
```
akomiqaia, line 72: Do we have to import  different labels? Seems like `SimpleLabel` is fine as long as it has correct props?

components/o-labels/stories/content.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+import withHtml from 'origami-storybook-addon-html';
+import {withDesign} from 'storybook-addon-designs';
+import {ContentLabel as ContentLabelTsx} from '../src/tsx/label';
+import './labels.scss';
+
+const brand = process.env.ORIGAMI_STORYBOOK_BRAND;
+const ComponentDescription = {
+	title: 'Components/o-labels',
+	component: ContentLabelTsx,
+    includeStories: (brand === 'core' ? /.*/ : []),
```
akomiqaia, line 11: The formating is a bit off here.

notlee: 👍 I need to sort out that prettier pr...

akomiqaia (approved):

notlee:
helpful as always percy 
Screenshot 2022-10-12 at 16 03 48

PR #845, opened 2022-10-13
fix: run percy on push to main

notlee:
percy still is not running on push to main – because we are using an absolute path? 
relates to: ca1990720c17c0d9e110591d076a7973c3b5d34b

### DIFF
--- a/.github/workflows/percy.yml
+++ b/.github/workflows/percy.yml
@@ -12,9 +12,9 @@ on:
   push:
     branches: main
     paths:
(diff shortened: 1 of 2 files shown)


notlee:
>Path patterns must match the whole path, and start from the repository's root.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths

akomiqaia (approved):

PR #846, opened 2022-10-14
fix: storybook build / brand unique stories

notlee:
Looks like our Storybook setup is a little borked. We used a dynamic includeStories property to include different stories depending on which brand we’re building the StoryBook for, It works fine locally but the prod build errors because Storybook uses static analysis during the prod build to determine which stories to include/exclude – it doesn’t execute the code, just looks for an array or regex. Looks like someone reported the same issue / expectations as us back in Jul.

Is only an issue when `storyStoreV7` or `buildStoriesJson` is enabled in storybook config. `storyStoreV7` will be the default in the future. So instead of using a dynamic `includeStories` property let's include stories based on directory structure:

`stories/*.stories.tsx` - stories for all brands
`stories/[brand]/*.stories.tsx` - stories for a specific brand 
`stories/shared/.tsx` - exports stories for re-export to support muliple specific brands (no `.stories` extension)

Previous broken use of `includeStories`:
https://github.com/Financial-Times/origami/blob/b33ef6ee611a09eef5e3ec08c04a1fc744b54df0/components/o-labels/stories/content.stories.tsx#L11

Example error in CI:
https://github.com/Financial-Times/origami/actions/runs/3236117619/jobs/5301479773

Source of error (storybook uses static analysis):
https://github.com/storybookjs/storybook/blob/ea70649db9bce2ed4ad2b1eaa13a380fb8cc8e57/code/lib/csf-tools/src/CsfFile.ts#L25

Others had the same issue/expectation:
https://github.com/storybookjs/storybook/issues/18792

### DIFF
--- a/apps/storybook/config/main.js
+++ b/apps/storybook/config/main.js
@@ -1,59 +1,59 @@
-const { readFile, access } = require('node:fs/promises');
-const {constants } = require('node:fs');
-const path = require('node:path');
-const process = require('node:process');
-const globby = require('globby');
+const {readFile, access} = require("node:fs/promises")
+const {constants} = require("node:fs")
+const path = require("node:path")
+const process = require("node:process")
+const globby = require("globby")
 
-const brand = process.env.ORIGAMI_STORYBOOK_BRAND || "core";
+const brand = process.env.ORIGAMI_STORYBOOK_BRAND || "core"
 /**
  * Node.JS no longer has an fs.exists method.
  * Instead we use the fs.access method and check we can read the file.
  * fs.access will throw an error if the file does not exist.
  * @param {string} file file-system path to the file you are wanting to check exists or not
  * @returns {Promise.<boolean>} Whether the file exists
-*/
+ */
 async function fileExists(file) {
 	try {
-		await access(file, constants.R_OK);
-		return true;
+		await access(file, constants.R_OK)
+		return true
 	} catch (error) {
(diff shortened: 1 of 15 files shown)


apps/storybook/config/main.js
```diff
@@ -158,13 +166,12 @@ module.exports.webpackFinal = async function webpackFinal(config) {
...
+module.exports.env = config => ({
+	...config,
+	ORIGAMI_STORYBOOK_BRAND: brand,
 })
 
-
 module.exports.features = {
-	buildStoriesJson: true
+	buildStoriesJson: true,
+	storyStoreV7: true,
```
notlee, line 176: this is the only real change here

apps/storybook/config/main.js
```diff
@@ -66,28 +66,36 @@ module.exports.core = {
...
+		const storiesForBrand = brands.includes(brand)
+			? await globby(
+					[
+						`${componentDirectory}/stories/*.stories.@(mdx|js|jsx|ts|tsx)`,
+						`${componentDirectory}/stories/${brand}/*.stories.@(mdx|js|jsx|ts|tsx)`,
+					],
+					{
+						gitignore: false,
+						expandDirectories: false,
+						deep: 1,
```
notlee, line 90: this also changed

akomiqaia (approved):

PR #848, opened 2022-10-14
517 dac date error handling

### LINKED ISSUE #517: o-forms: improve error handling for date inputs - ID: DAC_Error_Handling_01
# Issue
This is from page 75 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-forms@9.0.1&demo=inverse-form&system_code=origami-repo-data&brand=core>

When users enter incorrect data into the date input fields, the error handling is confusing. When users begin to enter into the first ‘DD’ input field a message appears which reads ‘Please fill out this field with the required format (DD/MM/YYYY)’. For sighted users they are able to see this message, but it is not announced to screen reader users.
           
When users enter data into the second input field ‘MM’ the message "Please fill out this field with the required format (DD/MM/YYYY)" disappears and the colour of the input field changes colour to green which would indicate to sighted users that they have entered the information correctly.
But if the user enters the date in the wrong format an error summary appears at the top of the page as expected. It would benefit all users if the use of colour in the input fields was removed as it may be confusing for users who may think the green colour indicates successful validation.

 
# Proposed Solution:
Remove the JavaScript functions which add colour and the error message to the input fields. Use inline HTML error messages in conjunction with the error summary. Further information can be found in the Error messages – Date inputs, Error message and Error summary in the GOV.UK Design System.

### LINKED ISSUE #536: o-forms: usability issue on date demo - ID: Date_Inputs_01
# Issue
This is from page 108 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-forms@9.0.1&demo=date&system_code=origami-repo-data&brand=master>

The invalid / inline date input fields all announce as ‘invalid entry’ when screen reader users enter into them. This may be confusing and more time consuming for screen reader users.

# Screen Reader comments

> Usability comment: Several of the date input fields on this page reported ‘invalid entry’ as part of their error handling. This was confusing because the edit fields contained no data.
> Thus, this announcement may confuse users into thinking that the edit fields are already contain information.
> It would be helpful if such announcements only appeared when errors actually occurred. This would provide confidence to users and ensure accuracy in error reporting.
> Please note that this issue was present with NVDA. And VoiceOver but did not occur to VoiceOver.

 
# Proposed Solution:
Ensure that the date inputs only announce ‘invalid entry’ when the entry is invalid or remove it completely.

akomiqaia:
According to DAC report our date input had few accessibility issues. Before it was relayed on HTML5 validation but now error message behaviour is controlled by custom javascript. That checks if Date is a valid date and if the format entered by user is valid. 

closes #517 
closes #536


components/o-forms/test/input.test.js
```diff
@@ -83,7 +87,9 @@ describe('Input', () => {
 			proclaim.isTrue(parentClass(dateField, 'invalid'));
 
 			dateField.value = 10;
-			dispatch('input', dateField);
+			monthField.value = 10;
+			yearField.value = 2022;
+			dispatch('blur', dateField);
```
akomiqaia, line 92: The new validation checks if the date itself is valid rather than if input pattern matches regex patterns defined before so I had to modify the tests a bit to pass them.

components/o-forms/src/js/input.js
```diff
@@ -39,18 +39,56 @@ class Input {
...
+	validateDate() {
+		const elements = this.parent.querySelectorAll('input');
+		const [date, month, year] = Array.from(elements).map(element => element.value);
+
+		const dateObj = Date.parse(`${date}/${month}/${year}`);
+		if (isNaN(dateObj)) {
+			this.toggleParentClasses("invalid");
+			return false;
+		}
+
```
notlee, line 68: The `Input` class seems to have a lot of responsibilities. It looks like it's intended to represent a single html `input` (not an `.o-forms__input` element, which may contain multiple html `inputs`). So it's unexpected that it checks the validity of inputs other than itself (date, month, year) and then modifies its parent elements.

Have you considered returning a boolean value here to indicate the  individual html `input`'s validity, and moving everything else to `forms.js`? We could use [validateFormInputs](https://github.com/Financial-Times/origami/blob/38586847f0160fa43bd298fcda51a83e95a71d41/components/o-forms/src/js/forms.js#L119) to determine the overall `.o-forms__input` validity based and toggle classes

components/o-forms/src/js/input.js
```diff
@@ -39,18 +39,56 @@ class Input {
...
+			this.toggleParentClasses("valid");
+		}
+
+		return true;
+	}
+
+
+	validateDate() {
+		const elements = this.parent.querySelectorAll('input');
+		const [date, month, year] = Array.from(elements).map(element => element.value);
```
notlee, line 61: What if a user has made a mistake in their html and the elements aren't found?

components/o-forms/src/js/input.js
```diff
@@ -39,18 +39,56 @@ class Input {
...
-			this.parent.classList.remove(this.className.invalid);
-			this.parent.classList.add(this.className.valid);
+			this.toggleParentClasses("valid");
+		}
+
+		return true;
+	}
+
+
+	validateDate() {
```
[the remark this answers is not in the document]
notlee: Kinda related, it looks like we've lost the red border as the css is using the `:invalid` selector to highlight only the invalid `inputs`within `.o-forms-input`.

`.o-forms-input--invalid :invalid:not(:disabled)`

Screenshot 2022-10-21 at 09 18 02

components/o-forms/src/js/input.js
```diff
@@ -39,18 +39,56 @@ class Input {
...
+		const elements = this.parent.querySelectorAll('input');
+		const [date, month, year] = Array.from(elements).map(element => element.value);
+
+		const dateObj = Date.parse(`${date}/${month}/${year}`);
```
notlee, line 63: Hmmm something isn't quite right here!

12th Dec is okay?
Screenshot 2022-10-21 at 09 15 15

But 13th isn't?
Screenshot 2022-10-21 at 09 14 59

notlee (changes requested):

components/o-forms/src/js/input.js
notlee: `alert('Not like that!')` 😂 ?

components/o-forms/README.md
```diff
@@ -443,15 +443,15 @@ To create our date input use a [multiple input](#multiple-input-fields) field st
 
 	<span class="o-forms-input o-forms-input--date">
 		<label for="date">
-			<input id="date" class="o-forms-input__day-part" type="text" inputmode="numeric" pattern="0[1-9]|[12]\d|3[01]" name="my-date" aria-label="Day (DD)"/>
+			<input id="date" class="o-forms-input__day-part" type="text" inputmode="numeric" name="my-date" aria-label="Day (DD)"/>
```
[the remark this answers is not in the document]
akomiqaia: that is a good point. I will bring them back but I might need to update JS as well so is we are running JS we can remove pattern attribute from input elements. Also some of this regex needs updating I think last time I tried to test it regex was a bit wrong.

components/o-forms/src/js/input.js
akomiqaia: maybe I misunderstood what you meant by this comment and somewhere I do agree with this explanation. But Date input is 3 different inputs that we want to look at it as one. I have been looking at the code and it seems like form.js `validateFormInputs` runs on submit rather than each update of an input. We could update code but then checking whole form validity every-time we type something does not sound good.

components/o-forms/src/js/input.js
akomiqaia: that is good point. Also jsut thinking about edge cases. Do we have use cases when our date fields are used only to get month and year. The code that i wrote will validate dates only that have all 3 inputs presented in the markup. It will fail to validate `12(mm) 2022(yyyy)` or `dd mm` inputs. should we take this into consideration?

components/o-forms/src/js/input.js
akomiqaia: `Date.parse` seems not to like `date/month/year` order and instead it expects it to be `month/date/year`.

notlee: Ah okay. Looks like parse behaviour for [non-standard date strings could vary per browser](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#fall-back_to_implementation-specific_date_formats).  Looks like we'll want to [pass in YYYY-MM-DD](https://tc39.es/ecma262/#sec-date-time-string-format)?

components/o-forms/src/js/input.js
notlee: >Date input is 3 different inputs that we want to look at it as one

Yeah that's why this has become a bit (more) confusing. Until now `input.js` represented validation for only one html `input`. Not a set of html inputs working together for one `.o-forms__input`. 

>I have been looking at the code and it seems like form.js validateFormInputs runs on submit rather than each update of an input.
>We could update code but then checking whole form validity every-time we type something does not sound good.

Good spot, I didn't notice that, we still want to check validity of form fields as they're being modified at the moment so my suggestion is no good.

Refactoring without a breaking change is difficult because Inputs are passed through Forms publicly as [Forms.prototype.formInputs](https://github.com/Financial-Times/origami/blob/38586847f0160fa43bd298fcda51a83e95a71d41/components/o-forms/src/js/forms.js#L18). Let's chat through some options (maybe we plan for a future major release)

notlee (approved):

PR #854, opened 2022-10-18
feat: o-labels, support custom labels with tsx templates

notlee:
It was not possible to use templates with custom labels. To support this we can allow any state on the BaseLabel tsx component.

```scss
@include oLabelsAddState('some-example', (
    'background-color': oColorsByName('claret')
));
```

```jsx
<BaseLabel state="some-example" size="big">my custom label</BaseLabel>
```


Behnazz (commented):
LGTM 🎉

akomiqaia (approved):

PR #861, opened 2022-10-21
ADSDEV-1204 Cookie Banner DAC_Reflow_06

ytcleon:
This PR is to improve reflow accessability of the cookie banner in `o-cookie-message` by reducing the size of padding when viewport size is 320px.

JIRA ticket: https://financialtimes.atlassian.net/browse/ADSDEV-1204

### DIFF
--- a/components/o-cookie-message/src/scss/_mixins.scss
+++ b/components/o-cookie-message/src/scss/_mixins.scss
@@ -63,15 +63,8 @@
 		padding: $_o-cookie-message-spacing;
 		padding-top: oSpacingByIncrement(7);
 		max-width: none;
-	}
-
-	@media (max-width: 320px) {
-		.o-cookie-message__inner {
-			padding-top: oSpacingByIncrement(1);
-			padding-bottom: oSpacingByIncrement(1);
-			padding-left: oSpacingByIncrement(2);
-			padding-right: oSpacingByIncrement(2);
-		}
+		max-height: 50vh;
+    overflow: scroll;
 	}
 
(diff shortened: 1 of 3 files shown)


notlee:
@ytcelon I don't think this solves the problem highlighted by DAC? If I set my browser to 1280x1024 and zoom to 400% it still looks like this (the cookie message can't be scrolled, but the page behind scrolls):

Screenshot 2022-10-21 at 11 56 30

What do you think of updating `.o-cookie-message` to ensure it doesn't cover more than 50% of the viewport height and is scrollable?
```
    max-height: 50vh;
    overflow: scroll;
```
Screenshot 2022-10-21 at 11 58 03
Screenshot 2022-10-21 at 11 58 14

notlee (commented):
see https://github.com/Financial-Times/origami/pull/861#issuecomment-1286803302

ytcleon:
Thank you for the suggestion. I tested it within a viewport of 1280 X 1024, so that may be why my commit does not work. The scrollable approach gives much better result. Happy to make the change.

components/o-cookie-message/src/scss/_mixins.scss
```diff
@@ -65,6 +65,11 @@
 		max-width: none;
 	}
 
+	.o-cookie-message__content-outer {
+		max-height: 50vh;
+		overflow: scroll;
```
notlee, line 70:
```suggestion
		overflow: auto;
```

notlee (approved):

PR #863, opened 2022-10-25
feat: Add link variant to ft-concept-button 

notlee:
1. Hide the current concept button styles from Origami because it is not used
and is a candidate for deprecation. That style seems to be only still used in
one place, via n-myft-ui, on the contact preferences page. In addition it is a
link, not a button, like the one here in Origami.
2. Add ft-concept-button__link to fulfil the same purpose as the above but
on the newsletter page. This will be the preferred style going forward. It's
teal to match the colour of our links.

Screenshot 2022-11-02 at 17 10 51

### DIFF
--- a/components/ft-concept-button/src/tsx/concept-button.tsx
+++ b/components/ft-concept-button/src/tsx/concept-button.tsx
@@ -14,6 +14,16 @@ export interface ConceptButtonProps {
 	// for props that aren't appropriate in Origami
 	extraButtonProps: any;
 }
+export interface ConceptLinkProps {
+	// link text
+	label: string;
+	href: string;
+	// for attributes which aren't appropriate in Origami
+	attributes?: {
+		[attribute: string]: string | boolean;
+	};
+	onClick?: Function;
+}
 
 function ConceptButton({
 	label,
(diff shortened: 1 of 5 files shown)


components/ft-concept-button/stories/concept-button.stories.tsx
```diff
@@ -8,6 +8,15 @@ export default {
 	title: 'Components/ft-concept-button',
 	component: ConceptButton,
 	decorators: [withDesign, withHtml],
+	// @todo Deprecate concept pill button?
+	// It looks like the concept pill has only ever
+	// been intended as a link to concepts.
```
notlee, line 13:
```suggestion
	// @deprecated The concept pill is a candidate for deprecation.
	// They were only ever intended as use as links button use a 
	// `button` element` in the template. The concept pill style
	// is only used in one place via n-myft-ui, in a dubious way.
	// We would prefer the teal style of `ft-concept-button_link`
	// going forward.
```

components/ft-concept-button/src/tsx/concept-button.tsx
```diff
@@ -14,6 +14,16 @@ export interface ConceptButtonProps {
 	// for props that aren't appropriate in Origami
 	extraButtonProps: any;
 }
+export interface ConceptLinkProps {
+	// link text
+	label: string;
+	href: string;
+	// for attributes which aren't appropriate in Origami
+	attributes?: {
```
andy-little, line 22: ooo, that's good, I should be doing that in the forms

PR #866, opened 2022-10-31
feat: add accesible mode that shows img html element instead of backg…

juanSanchezAlcala:
I have add an accesible new feature that adds img html elements instead of using a pseudo html element ::before to add the  background image fro the icons.
This is done dynamically after DomContentLoaded event is fired. This is not the best approach but there's a lot of scss logic that we will need to move to javascript in case we want to calculate the url of the icons from javascript. So for the moment maybe this could be the best approach.
[ticket](https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1316?modal=detail&selectedIssue=CI-1373)

### DIFF
--- a/components/o-share/src/js/share.js
+++ b/components/o-share/src/js/share.js
@@ -130,6 +130,16 @@ function Share(rootEl, config) {
 			aElement.appendChild(spanElement);
 			liElement.appendChild(aElement);
 			ulElement.appendChild(liElement);
+			if(config.accessible){
+				if (document.readyState === "complete" || document.readyState === "loaded") {
+					addAccesibleIcons(aElement);
+				}
+				else{
+					document.addEventListener('DOMContentLoaded',() => {
+						addAccesibleIcons(aElement);
+					});
+				}
+			}
 		}
 		oShare.rootEl.appendChild(ulElement);
 	}
@@ -179,6 +189,7 @@ function Share(rootEl, config) {
 				config.titleExtra = rootEl.getAttribute('data-o-share-titleExtra') || '';
 				config.summary = rootEl.getAttribute('data-o-share-summary') || '';
 				config.relatedTwitterAccounts = rootEl.getAttribute('data-o-share-relatedTwitterAccounts') || '';
+				config.accessible = rootEl.getAttribute('data-o-share-accesible') !== null;
 			}
 			render();
 		}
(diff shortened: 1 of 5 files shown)


notlee (changes requested):
Thanks for the PR! Unfortunately we're going to need to come up with a different approach, this one has a few issues.

First the hover and focus states are visually broken. The `img` is black and doesn't change on user interaction:
before, white on twitter blue after, black on twitter blue

Second because the images are added dynamically with JavaScript this won't improve the experience for users with JS turned off, or where JavaScript fails for whatever reason.

Third, `render` is never called when [full markup is provided by the user](https://registry.origami.ft.com/components/o-share@8.3.0/readme?brand=core#full-markup). That includes the ft.com article page.

Whatever approach we take we shouldn't add an `accessibility` option, as that means there's an option for an inaccessible component. It should always be accessible.

I see 3 potential approaches here:
1. We set an alternative icon for Windows High Contrast mode using the `-ms-high-contrast` media query. This will solve the specific users issue but as it's not standards based may not work for all high contrast based assistive technology – we could move to add support based on the new(ish) forced colours media query over the medium term. [I opened a draft PR to demonstrate](https://github.com/Financial-Times/origami/pull/868).
2. We include the design team and add a text label to all icons. When high contrast mode removes the background / icon there will still be text. 
3. We redevelop o-share with inline `img` elements and a `src` which points to the Image Service, and figure out how to change its colour on interaction. We release a major change with migration guide and update the apps which use them, including the article page. (This is my least favourite option because it's the most work; requires other teams to migrate and they may not; and I'm not sure how we'd manage hover states etc.)

What do you think? Could you discuss 1 and 2 with DAC and your team?

juanSanchezAlcala:
> Thanks for the PR! Unfortunately we're going to need to come up with a different approach, this one has a few issues.
> 
> First the hover and focus states are visually broken. The `img` is black and doesn't change on user interaction: before, white on twitter blue after, black on twitter blue
> 
> Second because the images are added dynamically with JavaScript this won't improve the experience for users with JS turned off, or where JavaScript fails for whatever reason.
> 
> Third, `render` is never called when [full markup is provided by the user](https://registry.origami.ft.com/components/o-share@8.3.0/readme?brand=core#full-markup). That includes the ft.com article page.
> 
> Whatever approach we take we shouldn't add an `accessibility` option, as that means there's an option for an inaccessible component. It should always be accessible.
> 
> I see 3 potential approaches here:
> 
> 1. We set an alternative icon for Windows High Contrast mode using the `-ms-high-contrast` media query. This will solve the specific users issue but as it's not standards based may not work for all high contrast based assistive technology – we could move to add support based on the new(ish) forced colours media query over the medium term. [I opened a draft PR to demonstrate](https://github.com/Financial-Times/origami/pull/868).
> 2. We include the design team and add a text label to all icons. When high contrast mode removes the background / icon there will still be text.
> 3. We redevelop o-share with inline `img` elements and a `src` which points to the Image Service, and figure out how to change its colour on interaction. We release a major change with migration guide and update the apps which use them, including the article page. (This is my least favourite option because it's the most work; requires other teams to migrate and they may not; and I'm not sure how we'd manage hover states etc.)
> 
> What do you think? Could you discuss 1 and 2 with DAC and your team?

@notlee  I've found that there is another css attribute for that purpose [prefers-contrast](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast). Maybe that will solve the issue with the option number 1 since it seems standard and supported for almost all browsers

PR #868, opened 2022-10-31
fix: o-share, show icons with windows high contrast mode enabled

notlee:
Show icons with windows high contrast mode enabled.

`forced-colors` is the standard alternative but is quite new and not widely adopted.
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors We would need to update the way we colour icons (perhaps using masks and background colours) to use the text colour set by forced-colors. https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#system_colors

### DIFF
--- a/components/o-share/src/scss/_mixins.scss
+++ b/components/o-share/src/scss/_mixins.scss
@@ -24,6 +24,13 @@
 
 		.o-share__icon--#{$icon-name}:before {
 			background-image: url($service-url + $query + "&format=svg&tint=#{$color-encoded}");
+			@media screen and (-ms-high-contrast: active) {
+				background-image: url($service-url + $query + "&format=svg&tint=%23ffffff,%23ffffff");
+			}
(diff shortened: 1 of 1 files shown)


notlee:
Relates to: https://financialtimes.atlassian.net/browse/CI-1373

o-share actually already accounts for `-ms-high-contrast: active` and  `-ms-high-contrast: black-on-white`. 
The case above mentions "[for] users that rely on technologies such as ZoomText with colour inversion selected, and Windows High Contrast Mode to navigate, these images disappear" with a user comment of “When viewing the page in inverted mode, the social links completely disappear because I presume, they are black on a black background. Ideally these should be a different colour to black.” It sounds like the issue appears when using ZoomText, or ZoomText and Windows High Contrast Mode together. We'll need to do more to improve this. Like put images inline in html, or better add a text label to each.

PR #871, opened 2022-11-01
fix: fix DAC issues for header-ervices

### LINKED ISSUE #551: o-header-services: improve markup for drop-down menus - ID: DAC_Headings_And_Labels_Name_Role_Value_01
# Issue
This is from page 170 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on: <https://www.ft.com/__origami/service/build/v3/demo?component=o-header-services@5.0.1&demo=drop-down-navigation&system_code=origami-repo-data&brand=master>

The ‘DOCS’ and ‘SPECS’ links in the navigation region have both the links and the expandable buttons wrapped inside the same list item. As a result, both the links and the buttons are keyboard focusable.
When users attempt to activate the links using a mouse, the expandable menus open and then quickly close again, and users are only able to open the menus by clicking on the adjacent buttons. The labels for the buttons have duplicated aria-labels which read ‘Toggle dropdown menu’ which screen reader users may find confusing as it is not clear what the dropdown menu is opening.
Additionally, when the buttons are activated and the menus expand, screen readers are not informed of this behaviour.

# Screen reader comments:

> “Exploring the Primary navigation region both in and out of context with the arrow keys and the ‘B’ keystroke to cycle between buttons, I found that activating toggle buttons did not report that each expands to reveal new content.
> This is likely to confuse some Screen Reader users because they will have no way of knowing that the button reveals expanded content aside from using the arrow keys.
> Including this information when the toggle buttons activate should make comprehension and navigation more efficient.”
> 
> “Navigating through the toggle buttons with the ‘B’ keystroke, I found that the toggle buttons were not labelled. All I heard when moving through these buttons was the phrase ‘toggle dropdown button’.
> This was confusing when viewed out of context and may also confuse other Screen Reader users as there is no way to determine to which menu the toggle button refers out of context.
> Ensuring that toggle buttons are clearly labelled should help to avoid confusion and make navigation easier.”

>  “When tabbing through navigation, I have to tab into the dropdown arrow itself to open it fully. This is unusual behaviour unusual would expect it to be encapsulated with the title for it to open as one. So that they know which dropdown correlates with which title.

# Proposed Solution:
It would benefit all users if the links / buttons were two, individual focusable elements as opposed to two links and two buttons. Further information can be found in [Fly-out Menus - W3.org](https://www.w3.org/WAI/tutorials/menus/flyout/).

akomiqaia:
closes #551

### DIFF
--- a/components/o-header-services/demos/src/header.mustache
+++ b/components/o-header-services/demos/src/header.mustache
@@ -27,8 +27,8 @@
 		<ul class="o-header-services__primary-nav-list">
 			<li {{#drop-down}}data-o-header-services-level="1"{{/drop-down}}>
 				<a {{^drop-down}}aria-current="page"{{/drop-down}} href="{{href}}">Documentation</a>{{#drop-down}}<!--
-				--><button class="o-header-services__drop-down-button" type="button" name="button" aria-label="Toggle dropdown menu"></button>
-				<ul data-o-header-services-level="2" aria-hidden="true">
+				--><button class="o-header-services__drop-down-button" type="button" name="button" aria-controls="documentation-dropdown" aria-label="Toggle documentation dropdown menu"></button>
+				<ul id="documentation-dropdown" data-o-header-services-level="2" aria-hidden="true">
 					<li>
 						<a href="{{href}}">Principles</a>
 					</li>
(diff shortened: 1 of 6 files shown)


notlee:
It looks like this has broken the drawer on a mobile view. When toggled the drop down nav obscures top level nav items and their associated drop down button:


before:
Screenshot 2022-11-08 at 13 19 12

now:
Screenshot 2022-11-08 at 13 19 08

akomiqaia:
@notlee the last commit should fix this.

notlee (approved):
🙌

notlee (approved):

PR #872, opened 2022-11-01
O forms radiobox and text inputs

andy-little:
- created form basic components
- created the following input components: Checkboxes, Toggle, Date input, File input, Pseudo Radio Links, Select boxes, Text Area, Text input, Box-styled radio inputs, Round-styled radio inputs
- added storybook demos


components/o-forms/src/tsx/TextInput.tsx
```diff
@@ -0,0 +1,122 @@
...
+	ref?: any; /* Look up correct type */
+}
+export interface TextInputProps extends TypeTextInput, TypeFormField{
+}
+interface PrivateTextInputProps extends TypeTextInput {
+	id: string;
+}
+
+
+function PrivateTextInput({
```
andy-little, line 25: This is poorly named 🥴. This component is the actual TextInput. Originally I thought the users should have to nest an input component inside a <FormField>. I thought this might allow for more customisation, but we changed tact to abstract away the <FormField> within an input component... So technically this component doesn't really need to exist but I wanted to keep it around incase it turns out to be better to offer people the flexibility having separate input and formfield components

components/o-forms/src/tsx/Form.tsx
```diff
@@ -0,0 +1,148 @@
+import uniqueId from 'lodash.uniqueid';
```
akomiqaia, line 1: So far Origami's tsx templates are exported from [COMPONENT_NAME].tsx file.  Maybe this file should be called `Partials.tsx` or something different from`Forms.tsx` and rest of the components can be exported from here. So for example as origami user I should be able to import all components from one file: 
```js
import { BoxRadioBtns, TextInputs} from '@ft/Form.tsx'
```

So users does not have to look up files names and this way our exporting is consistent as well

components/o-forms/stories/boxRadioBtns.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+StateBoxRadioButton.args = {
+children: [
+		<BoxRadioBtn
+		name="default"
+		value="Daily"
+		checked/>,
+		<BoxRadioBtn
+		name="default"
+		value="Weekly"/>,
```
akomiqaia, line 122: Have you tried to render children conditionally inside the template itself? instead of passing them as SB controls? At he moment Children control in storybook is JSX element object and it is a bit hard to update values.

components/o-forms/stories/textBox.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+};
+
+export default {
+	title: 'Components/o-forms/text-box',
+	component: TextInput,
+	decorators: [withDesign, withHtml],
+	argTypes:{
+		onChange: hideArg,
+		ref: hideArg,
+		children: hideArg,
```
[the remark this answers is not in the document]
andy-little: This is to hide the children from the control panel :)

components/o-forms/stories/textBox.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+};
+
+export default {
+	title: 'Components/o-forms/text-box',
+	component: TextInput,
+	decorators: [withDesign, withHtml],
+	argTypes:{
+		onChange: hideArg,
+		ref: hideArg,
```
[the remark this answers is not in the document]
andy-little: In react if you don't want to have a controlled input you can use a ref instead. https://reactjs.org/docs/uncontrolled-components.html

components/o-forms/stories/textBox.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+    disable: true
+  }
+};
+
+export default {
+	title: 'Components/o-forms/text-box',
+	component: TextInput,
+	decorators: [withDesign, withHtml],
+	argTypes:{
+		onChange: hideArg,
```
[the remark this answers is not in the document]
andy-little: It's an event listener. When using client side react it is common to use controlled inputs so that the value of the form is in sync with your applications state. If you pass a value to the input component you wont be able to change the value in the browser with out doing something like:
`onChange = (e) => {setValue(e.target.value)`.
You can see how I have set it up in the textBox.stories :)
https://reactjs.org/docs/forms.html

components/o-forms/src/tsx/Form.tsx
andy-little: 👍🏻👍🏻👍🏻

components/o-forms/stories/boxRadioBtns.stories.tsx
andy-little: I have but typescript complains. I did something very similar to [this](https://storybook.js.org/docs/react/essentials/controls#fully-custom-args). But actually at the top is says !ts and !tsx
I have a couple of slightly less than ideal ideas how to solve this, hard to explain in text form though, might be nice to discuss at some point if poss.

akomiqaia:
Really nice work @andy-little . This was a big PR and took me some time.  There are few general things that I wanted to mention. 

The code formatting is a not correct in some files but this is not your fault at all. We had conflicting configurations and we still have not managed to get to the bottom of it. At the moment what I do is to comment `component/` inside `.prettierignore` and then format files using vscode formatter. Once done with formatting you can uncomment `component/` directory. This should not be happening but unfortunately we did not have time to have a good look at prettier and esling config. 

very well done on storybook as well. You have learned quite a lot in a very short amount of time and demos look great. Few suggestions that could make them perfect is to find a way to render controls relevant to a story. for example `Optional text` story has a control called `highlight` but switching it does have any effect on it. and each story has few controls that are not set. We want to hide controls that are not in use by story and or give them values that are meaningful for the component.  

Just a small suggestion and maybe something that is more of a personal choice but text input demos in sotrybook are very similar and it seems like some demos can be condensed. Like for `Optional input` story you already have a control that demos optional component and same with Passowrd, email, textarea you have `type` control that changes element in the demo.

andy-little:
Nice one, thanks Ako, very helpful :). I've fixed the formatting. Regarding the OptionalTextStory and the highlight control, I've made a slight adjustment, it _is_ working but you need to have some text in the field to notice the highlight, and if there is an error set it wont do anything. I've now tried to hide the highlight param if an error message is set, you can see this in the argTypes of the textbox story, for some reason though it doesn't work 🤔. Did you notice any other controls not working? I think some controls only work in conjunction with others being set. Tricky to know how to deal with that. That is one reason that I quite like having a bunch of examples. I also quite like it because it means you can easily browse and see what is possible/get some inspiration. I'm happy to cut down on them though :)

.prettierignore
```diff
@@ -1,4 +1,4 @@
-components/
+# components/
```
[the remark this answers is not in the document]
andy-little: oops

components/o-forms/package.json
```diff
@@ -51,5 +51,9 @@
     "npm": "7.11.1"
   },
   "percy": true,
-  "private": false
+  "private": false,
+  "dependencies": {
+    "@types/lodash.uniqueid": "^4.0.7",
+    "lodash.uniqueid": "^4.0.1"
```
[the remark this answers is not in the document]
andy-little: 👍🏻 done :)

components/o-forms/stories/boxRadioBtns.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+};
+
+export const BoxRadioButton = Template.bind({});
+export const NegativeHighlight = Template.bind({});
+export const MultipleBoxRadioButton = Template.bind({});
+export const DisabledBoxRadioButton = Template.bind({});
+export const ErrorBoxRadioButton = Template.bind({});
+export const StateBoxRadioButton = Template.bind({});
+
+BoxRadioButton.args = {
```
frshwtr, line 39: (I need to look into this deeper, this is more of a comment for myself)

There appears to be a Storybook control, state: saved and saving, but no state to return to the default where neither "Saving" or "Saved" are displayed. Look into where this is coming from and feed back accordingly (It's most likely being driven out by the component props)

frshwtr: Yes, storybook picked the control type based on the interface definition. Oddly it doesn't pick up that it's optional. Having the option to see the default state would help users see all the possible states.

components/o-forms/src/tsx/RadioBtns.tsx
```diff
@@ -0,0 +1,171 @@
...
+		wrapperProps.type = 'round';
+	}
+	return (
+		<FormFieldset {...fieldsetProps}>
+			<RadioBtnsBoxWrapper {...wrapperProps}>{children}</RadioBtnsBoxWrapper>
+		</FormFieldset>
+	);
+}
+
+export function RadioBtns({
```
frshwtr, line 112: Is there a use case where this component would be used instead of `RadioBtnsBox`?

components/o-forms/stories/radioBtns.stories.tsx
```diff
@@ -0,0 +1,88 @@
+import withHtml from 'origami-storybook-addon-html';
+import {withDesign} from 'storybook-addon-designs';
+import {ComponentStory, ComponentMeta} from '@storybook/react';
+import {useEffect} from 'react';
+import {RadioBtn, RadioBtns} from '../src/tsx/o-forms';
+import './forms.scss';
+import javascript from '../main.js';
+
+export default {
```
frshwtr, line 9: A lot of the demos for these radio buttons appear to be getting their controls from the `<RadioBtns .../>` container. Which are not related to the component being demoed. The container appears to exist to give the right styling/behaviour.

These controls can be removed so that a user only sees the controls specific to the component being demoed.

frshwtr: Actually, I think this demo should be simpler. `box-radio-buttons` is good to demo a working implementation of a combination of radio buttons. But I think the `radio-buttons` should be demonstrating just one instance of a button and its controls.

components/o-forms/stories/boxRadioBtns.stories.tsx
frshwtr: There's a reset button which seems to do the job 🤔

components/o-forms/src/tsx/RadioBtns.tsx
```diff
@@ -0,0 +1,171 @@
...
+		isVerticalCenter,
+	};
+	return (
+		<FormFieldset {...fieldsetProps}>
+			<RadioBtnsWrapper {...wrapperProps}>{children}</RadioBtnsWrapper>
+		</FormFieldset>
+	);
+}
+
+export function RadioBtn({
```
frshwtr, line 142: @notlee what's the convention with naming, are abbreviations allowed?

components/o-forms/stories/textBox.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+import {TextInput} from '../src/tsx/o-forms';
+import {Button} from '../../o-buttons/src/tsx/button';
+
+const hideArg = {
+	table: {
+		disable: true,
+	},
+};
+
+export default {
```
frshwtr, line 16: On this demo, I noticed that the class changes every time a new character is inputted. Is this normal?

components/o-forms/src/tsx/TextInput.tsx
frshwtr: This is a good point. I think it's good to keep the core of `TextInput` separate so it can exist as an Atom. Most of the form components seem to be molecules formed of labels and the core input. I'll have a think about the best naming here.

components/o-forms/stories/textBox.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+		highlightValid: {
+			if: {arg: 'errorMessage', truthy: false},
+		},
+	},
+} as ComponentMeta<typeof TextInput>;
+
+const Template: ComponentStory<typeof TextInput> = args => {
+	const [value, setValue] = useState(args.value);
+	const onChange = e => {
+		setValue(e.target.value);
```
frshwtr, line 35: Linked to [this comment](https://github.com/Financial-Times/origami/pull/872/files#r1082557641) The changing of the value in the text field here will re-set state and trigger a re-render. For the text box, this will mean that the [uuid generation](https://github.com/Financial-Times/origami/pull/872/files#diff-54480ccc29dc73c91a57814b3bdc5dd46e2c9fc4402f0dfea87502d0b3b9e0a5R87) will occur on every re-render, resulting in a new ID for the text box. 

This results in rapid changing of the class name for the text box label.

I have checked what affect this has on screen readers, as this ID links the label to the text input. It appears to have no adverse affect on VoiceOver on MacOS. Though it is worth noting that in Chrome's accessibility tree, the link changes to `undefined` for a very split second (almost impossible to notice without careful observation) before being set to the new value.

frshwtr: This is a react-specific issue and should not affect non-react users. It is only a problem when a user of this component uses `onChange` to update state owned by a parent and if the user also wants to target this ID. Given this is a component for a form, it is likely that a user may store the value of the input field into react state.

Does it impact our users? Not really. The re-generation of UUID on every user input introduces an unnecessary operation. Which might have a marginal hit to performance. But shouldn't be enough to cause issues.

**Options for next steps**

***1) Do nothing:*** though the re-generation is less than ideal. It shouldn't cause a huge enough burden to users. Screen readers will still have refs to pick up on. It will be worth auditing to see the full picture on what this means for accessibility.

***2) Have users pass in an ID***: this would mean the ID generation can be removed, but would pass the onus on setting the ID to the user. They would then need to ensure they are using best practise when using this component.

CC: @notlee I'm thinking of going down the path of option 1 for now. Perhaps we can revisit to optimise later?

notlee: That seems fair to me @frshwtr. If we become aware of any performance or usability issues we can introduce a requirement for ids. 

I like to keep in mind [W3C's platform design principles](https://www.w3.org/TR/design-principles/#priority-of-constituencies) when thinking about these kinds of problems:

>User needs come before the needs of web page authors, which come before the needs of user agent implementors, which come before the needs of specification writers, which come before theoretical purity.

users > product developers > origami team

components/o-forms/src/tsx/FileInput.tsx
```diff
@@ -0,0 +1,89 @@
+import {getInputClasses, uniqueId} from '../utils';
+import {FormError, InputProps, TypeFormField, FormField} from './Form';
+
+interface TypeFileInput extends InputProps {
+	highlightValid?: boolean;
+	errorMessage?: string;
+	accept?: string;
+	capture?: boolean | 'user' | 'environment';
+	multiple?: boolean;
+	ref?: any /* Look up correct type */;
```
frshwtr, line 10: TODO: this comment

components/o-forms/src/tsx/Select.tsx
```diff
@@ -0,0 +1,113 @@
+import {getInputClasses, uniqueId} from '../utils';
+import {FormError, InputProps, TypeFormField, FormField} from './Form';
+
+interface TypeSelect extends InputProps {
+	highlight?: 'valid' | 'invalid';
+	errorMessage?: string;
+	suffix?: JSX.Element;
+	isSmall?: boolean;
+	children?: JSX.Element[];
```
frshwtr, line 9: I think this might need to also allow one element:
```suggestion
	children?: JSX.Element | JSX.Element[];
```

frshwtr (approved):

PR #879, opened 2022-11-07
backstage: Do not validate o-forms date field before user has attempted each input.

notlee:
Previously it would always show an error when filling out the date field,
even before trying to give a month or year.

I'm not sure `Date.parse` was catching any errors that the regex isn't, and therefore we could simplify? 
In Firefox for example it's still possible to pass "31/11/2022" which is not actually a valid date. We could 
include a library to improve the validation further but perhaps should leave that for back-end logic for now?

This change also handles date fields which don't include all the inputs

before:
date-1

after:
date-2


akomiqaia (approved):

akomiqaia:
closes #536

PR #881, opened 2022-11-08
backstage: locally watch for any component / library change

notlee:
Rebuild watched components when any component / library changes. E.g. `npm run watch -w components/o-forms` will rebuild when o-normalise is updated.


notlee:
This could be smarter by watching only dependencies, but I think this is fine.

akomiqaia (approved):

PR #882, opened 2022-11-10
backstage: stop oForms initialising many times to avoid multiple instances of error summaries

### LINKED ISSUE #500: o-autocomplete: Error summary in demo is announced too many times for screen reader users and is not descriptive - ID: DAC_Errors_01
# Issue

This is from page 19 of the [audit report](https://drive.google.com/file/d/17wq6b0CIX41RA4muBFDzUDxE0-qzHBIH/view)

Issue can be found on:  <https://www.ft.com/__origami/service/build/v3/demo?component=o-autocomplete%401.3.0&demo=dynamic&system_code=origami-repo-data&brand=master>

When users press enter in the ‘Select your country’ input field, an alert message appears on the page which reads ‘There is a problem’. For screen reader users, the alert message is announced three times in succession which may be disorientating as aria-labelledby has been placed on the div which contains the error summary.
An error message has been placed below the input field with the message ‘Please fill out this field’ but screen reader users would not be able to identify the message as an error. The error message is also non-descriptive and asks the user to ‘Please fill out this field’.

# Proposed Solution:
Remove the aria-labelledby attribute from the div and ensure that the error message is descriptive and is assigned the role=alert to inform screen reader users of the error.

akomiqaia:
closes #500 

according to the DAC report, users were getting multiple instances of error summaries. This was a bug that was introduced in version o-autocomplete 1.6.1. The reason seems to be dispatching `o.DOMContentLoaded` events from demo JS files.


components/o-forms/src/js/forms.js
```diff
@@ -35,6 +35,21 @@ class Forms {
...
+			const inputsToRemove = this._formInputsCache.filter(input => !formElements.includes(input.input));
+			const elementsToAdd = formElements.filter(element => !this._formInputsCache.find((input) => input.input === element));
+			inputsToRemove.forEach(input => input.destroy());
+			this._formInputsCache = [
+				...this._formInputsCache.filter(input => !inputsToRemove.includes(input)),
+				...elementsToAdd.map(element => new Input(element))
+			];
+			return this._formInputsCache;
+		}
+		return null;
```
notlee, line 50: Previously `formInputs` was always an array, but now may be `null`. As a public property a user may rely on this being an array, for example calling `forEach`, and so this could break their code – we don't do a good job of documenting types in Origami components and should look to improve that 😬 

Instead we could return an empty array to indicate no form fields have been found.
```suggestion
		return [];
```

components/o-forms/src/js/forms.js
```diff
@@ -35,6 +35,21 @@ class Forms {
 			});
 		}
 	}
+	get formInputs() {
+		if (this.form) {
```
notlee, line 39: Little tip for the future, I like to return early. It avoids indenting the main body of a function :)

```js
if(!this.form) {
    return [];
}

// everything else
// everything else
// everything else
// everything else
// e.g another condition
if(thing) {
    console.log('hello');
}
```

vs

```js
if(this.form) {
    // everything else
    // everything else
   // everything else
   // everything else
   if(thing) {
        console.log('hello');
    }
}

return [];

```

components/o-forms/src/js/forms.js
```diff
@@ -184,8 +199,8 @@ class Forms {
 			});
 		}
 		this.form = null;
-		this.formInputs.forEach(input => input.destroy());
-		this.formInputs = null;
+		this._formInputsCache.forEach(input => input.destroy());
+		this._formInputsCache = null;
```
notlee, line 203:
```suggestion
		this._formInputsCache = [];
```

components/o-forms/test/forms.test.js
```diff
@@ -188,6 +188,46 @@ describe('Forms', () => {
...
+		});
+		it('removing input element should return correct number of formInputs elements', () => {
+			const originalFormInputsLength = formEl.querySelectorAll('input').length;
+			const fieldElement = formEl.querySelector('.o-forms-field');
+			const numberOfInputElements = fieldElement.querySelectorAll('input').length;
+			fieldElement.remove();
+			const formInputsLength = form.formInputs.length;
+			proclaim.equal(formInputsLength, originalFormInputsLength - numberOfInputElements);
+		});
+
```
notlee, line 211: Nice to see tests for this :)

notlee (changes requested):

components/o-forms/src/js/forms.js
akomiqaia: this is a good suggestion. Thanks

akomiqaia: [f8dc967](https://github.com/Financial-Times/origami/pull/882/commits/f8dc9679c069c92d421be5fb609d64cf9d2142d3)

components/o-forms/src/js/forms.js
akomiqaia: This is part of `destroy()` function. should this remain null?

notlee (approved):

PR #889, opened 2022-11-16
fix: refactor snappy styles for full-bleed-offset topper

juanSanchezAlcala:
refactor snappy styles in one method for full-bleed-offset topper


notlee:
Thanks for coming back to do refactoring 😄 Looks like will output the fluid grid styles and then if snappy mode is enabled output more styles to override them. That means for snappy grid projects we'll deliver more css than we need. Have you considered using a Sass if / else statement to only output the CSS needed with snappy mode enabled?

juanSanchezAlcala:
> Thanks for coming back to do refactoring 😄 Looks like will output the fluid grid styles and then if snappy mode is enabled output more styles to override them. That means for snappy grid projects we'll deliver more css than we need. Have you considered using a Sass if / else statement to only output the CSS needed with snappy mode enabled?

Yes sure, the reason of the refactor was mainly to have the snappy changes in one method so i could apply in the app due the app is working with snappy but not using that variable so i need to apply them calling that method.
Right now i have pushed some changes with conditionals in the code that would be override by the method

notlee:
I see, that's frustrating! In that case please remove the underscore `_` to rename `_oTopperThemeFullBleedSnappy` to `oTopperThemeFullBleedSnappy`. The underscore prefix is a naming convention that indicates a private Sass mixin which should not be used outside o-topper

notlee:
In fact @juanSanchezAlcala, what if we create a new global variable in `variables.scss` for the app to control this?

```scss
$o-topper-snappy: if($o-grid-mode == 'snappy', true, false) !default`;
```

Then use `$o-topper-snappy` in the if statements. The app can override with:

```scss
$o-topper-snappy: true;
@include '@financial-times/o-topper/main'
```

That's a little easier to understand and maintain I think, by avoiding having to override topper classes within the app.

notlee (approved):

PR #895, opened 2022-11-24
[CI-1411] Deep Portrait Left/Mid

AlbertoCuberoNavas:
Create a new topper called "Deep Portrait Left/Mid"

[https://financialtimes.atlassian.net/browse/CI-1411](https://financialtimes.atlassian.net/browse/CI-1411)


components/o-topper/src/scss/themes/_portrait.scss
```diff
@@ -0,0 +1,127 @@
...
+@mixin _oTopperThemePortrait {
+	.o-topper__read-next {
+		display: none;
+	}
+
+	&.o-topper--split-text-center {
+		@include oGridRespondTo(S) {
+			@include _oTopperAlignText(center);
+		}
+	}
```
notlee, line 12: Is this needed? It doesn't look like `o-topper--split-text-center` is used in the new demos?

Do these portrait toppers replace the existing`split-text` toppers? If so can we deprecate the split-text toppers and hide their demos?

notlee:
Hi @AlbertoCuberoNavas, sorry I haven't had time to do a full review. I'm on annual leave now but will be able to review this when I'm back on Wednesday. Just left a quick question for now about whether we can deprecate some existing toppers :)

notlee:
Hey @AlbertoCuberoNavas. It looks like there are some issues with the credit (`o-topper__image-credit`) and caption (`o-topper__image-caption`) elements – see the readme for information on how these two elements are used.

### Credit

Note that on the existing "Split Text Left" topper the `o-topper__image-credit` element sits over the image to the right, with white copy.
Screenshot 2022-12-01 at 12 05 59

By comparison, on the new "Split Text Portrait Left" it is positioned differently, and has low contrast against the background:
Screenshot 2022-12-01 at 12 07 12

### Caption

The `o-topper__image-caption` element should sit below the image, with contrasting colour, as it does in the existing "Split Text Left",:
Screenshot 2022-12-01 at 12 07 00

However it incorrectly displays at the top of the teaser for the new "Split Text Portrait Left":
Screenshot 2022-12-01 at 12 10 41

### Note

Final note, it would be great to update the demo with a portrait image like in the design, to demonstrate the benefits of this design :)

notlee:
"Split Text Left" and "Split Text Portrait Left" appears broadly the same on mobile. Is this intentional? If the design change is to better display portrait imagery I would expect the mobile layout to change too?

AlbertoCuberoNavas:
> "Split Text Left" and "Split Text Portrait Left" appears broadly the same on mobile. Is this intentional? If the design change is to better display portrait imagery I would expect the mobile layout to change too?

Its intentional. 

"Split Text Portrait Left" is a variation of "Split Text Left", so there are similarities between both design in differents aspects.

components/o-topper/demos/src/split-text-portrait-left.mustache
```diff
@@ -0,0 +1,27 @@
...
+
+	<figure class="o-topper__visual">
+		<picture class="o-topper__picture">
+			<source media="(max-width: 1440px)" srcset="https://www.ft.com/__origami/service/image/v2/images/raw/http://prod-upp-image-read.ft.com/3eceaf1e-61a3-11e7-91a7-502f7ee26895?source=origami-build-tools&amp;fit=scale-down&amp;quality=highest&amp;width=720 720w">
+
+			<source media="(min-width: 1440px)" srcset="https://www.ft.com/__origami/service/image/v2/images/raw/http://prod-upp-image-read.ft.com/6b7b8578-61a3-11e7-91a7-502f7ee26895?source=origami-build-tools&amp;fit=scale-down&amp;quality=highest&amp;width=800 800w">
+
+			<img alt="" class="o-topper__image" src="https://www.ft.com/__origami/service/image/v2/images/raw/http://prod-upp-image-read.ft.com/97a49d16-61e8-11e7-8814-0ac7eb84e5f1?source=origami-build-tools&amp;fit=scale-down&amp;quality=highest&amp;width=800">
+		</picture>
+		<figcaption class="o-topper__image-credit">© KMG/SWNS.com</figcaption>
```
notlee, line 24:
```suggestion
```

notlee (approved):
Nice, approved! :shipit: 

Only suggestion before merging is to remove `o-topper__image-credit` from the demo. Toppers should use only one of `o-topper__image-credit` or `o-topper__image-caption` – if there's just a credit use `o-topper__image-credit`, if there's a credit and caption put both in the  `o-topper__image-caption` element

components/o-topper/demos/src/split-text-portrait-left.mustache
AlbertoCuberoNavas: done!

PR #898, opened 2022-11-24
O layout tsx

akomiqaia:


components/o-layout/src/tsx/layout.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+type LayoutProps = {
+	layoutType: 'default' | 'bleed' | 'docs' | 'landing' | 'query';
+	header: ChildrenType;
+	mainContent: ChildrenType;
+	overviewSections: OverviewSectionsProps;
+	hero?: HeroProps;
+	sidebar?: SideBarProps;
+	footer: ChildrenType;
+	displayArticleList?: boolean;
```
notlee, line 39: This prop toggles a `o-layout-typography` class right? Could you explain why it's called `displayArticleList`?

Aside, perhaps we should also off some o-layout typography wrapper tsx component [o-layout-typography](https://registry.origami.ft.com/components/o-layout@5.2.4/readme?brand=core#typography)?

components/o-layout/src/tsx/layout.tsx
```diff
@@ -0,0 +1,138 @@
...
+};
+
+export function Layout({
+	layoutType,
+	header,
+	mainContent,
+	hero,
+	sidebar,
+	footer,
+	displayArticleList,
```
notlee, line 49: Some of these props only belong to specific layout types which is confusing. For example `hero` is only allowed with `layoutType: landing`. How about instead of a `layoutType` prop we have a `HeroLayout` component, `DocumentationLayout` component, etc? That way their props will be specific and whether they are required or not would be accurate

components/o-layout/stories/internal/internalLayout.stories.tsx
```diff
@@ -0,0 +1,172 @@
...
+LandingLayoutWithArticleList.argTypes = {
+	overviewSections: {
+		table: {disable: true},
+	},
+	sidebar: {
+		table: {disable: true},
+	},
+};
+
+export const QueryLayoutDemo = args => {
```
notlee, line 127: I think it's helpful to have a different story for each layout type with specific content to show off that layout and how it should be used 👍 We should hide the `layoutType` prop from controls though, because it can lead to some weird results. E.g. docs story with query type:
Screenshot 2022-12-01 at 16 18 48

components/o-layout/stories/internal/internalLayout.stories.tsx
```diff
@@ -0,0 +1,172 @@
...
+		Tabs.init();
+		HeaderServices.init();
+		Forms.init();
+		javascript.init();
+	});
+
+	return <QueryLayout {...args} />;
+};
+
+QueryLayoutDemo.storyName = 'Query Layout';
```
notlee, line 140: Navigation has been added in the query layout? I don't think that's intended?

Screenshot 2022-12-01 at 16 22 18

components/o-layout/src/tsx/landingPartials.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export function Hero({muted, heroContent}: HeroProps) {
+	const classNames = ['o-layout__hero o-layout-typography'];
+	muted && classNames.push('o-layout__hero--muted');
+	return <div className={classNames.join(' ')}>{heroContent}</div>;
+}
+
+export function OverviewSections({
+	sections,
+	hasAction,
```
notlee, line 11: We don't need a user to specify a `hasAction` prop, we can derive it based on whether any `sections` have an `actionElement`

components/o-layout/src/tsx/landingPartials.tsx
```diff
@@ -0,0 +1,54 @@
+import {HeroProps, OverviewSectionsProps, ArticleListProps} from './layout';
+
+export function Hero({muted, heroContent}: HeroProps) {
```
notlee, line 3: Have you considered passing `heroContent` as `children`?

```jsx
<Hero>
    <p>hello world!</p>
</Hero>
```

vs.

```jsx
const heroContent = <p>hello world!</p>;
<Hero heroContent={heroContent}/>
```

components/o-layout/src/tsx/layout.tsx
```diff
@@ -0,0 +1,138 @@
+import {Hero} from './landingPartials';
+type ChildrenType = JSX.Element | JSX.Element[];
+
+type SideBarProps = {
+	customNavHeadingSelector?: string;
+	customNavigation?: JSX.Element | JSX.Element[];
```
notlee, line 6: Similar to before, could `customNavigation` work well as `children`?

components/o-layout/src/tsx/layout.tsx
```diff
@@ -0,0 +1,138 @@
...
+};
+
+type OverviewSectionItemProps = {
+	element: ChildrenType;
```
notlee, line 10: I don't think `element` is very descriptive here, perhaps `content`? Although again maybe `children` would be a better interface to provide section content?

components/o-layout/src/tsx/layout.tsx
```diff
@@ -0,0 +1,138 @@
...
+	actionElement?: ChildrenType;
```
notlee, line 11: `actionElement` on the other hand sounds reasonably descriptive 👍

notlee (changes requested):
This came together more quickly than I expected, nice work 😄 I think there are areas we can improve though from a user experience perspective, as noted above. We should also review the controls to make sure they are all helpful. A lot of props are for children, which display this kind of nonsense in the controls! Perhaps some controls should be hidden. Maybe new controls should be configured within a story, as a more helpful demo for generating html to copy, and mapped to the relevant props.

Screenshot 2022-12-01 at 16 55 18

components/o-layout/src/tsx/layout.tsx
akomiqaia: @notlee it refers to [article list part](https://registry.origami.ft.com/components/o-layout@5.2.4/readme?brand=core#article-list) of the documentation.

components/o-layout/stories/internal/internalLayout.stories.tsx
akomiqaia: @notlee [documentation](https://registry.origami.ft.com/components/o-layout@5.2.4/readme?brand=core#query-sidebars) suggests that it supports sidebar navigation. Happy to update documentation and templates if this is no longer supported.

components/o-layout/src/tsx/layout.tsx
notlee: The `o-layout-typography` class can be used within any layout though, should the prop be more generic? FYI it can also can used multiple times throughout the page https://registry.origami.ft.com/components/o-layout@5.2.4/readme?brand=core#typography

components/o-layout/stories/internal/internalLayout.stories.tsx
notlee: Oh cool, good spot, that's fine :) Though it's disabled by default so it would be better to be off by default in the demos too

components/o-layout/src/tsx/layout.tsx
akomiqaia: [617c3a4](https://github.com/Financial-Times/origami/pull/898/commits/617c3a49fa2fc6e3810bc78113656fd49c2f0e4e) what do you think about this update?

A short summary of the commit: 
removed `SideBarProps` and instead moved `customNavHeadingSelector` prop to controls. For custom navigation, if users pass down any children to `DocsLayout` it will render custom navigation. I added some comments and an example inside a story file.

akomiqaia: or I guess we could use a decorator for this too and have a boolean control to display custom sidebar navigation?

components/o-layout/stories/internal/internalLayout.stories.tsx
akomiqaia: [a87d9e1](https://github.com/Financial-Times/origami/pull/898/commits/a87d9e1d48820b6378a769da7fe74362d36f7e87) made a few changes here so we could destroy sidebar nav correctly and also not to allow duplicate nodes in case of rerendering

components/o-layout/src/tsx/layout.tsx
akomiqaia: So yea this toggles `o-layout-typography` for landing layout. I changed the name of it but according to docs if we are using article list inside the main content we need to remove `o-layout-typography` class.

components/o-layout/src/tsx/layout.tsx
akomiqaia: [00cac54](https://github.com/Financial-Times/origami/pull/898/commits/00cac547d9e8402a9927c59d3f63e8748daa48b7) updated it to content. We could use children too but `OverviewSections` expects an array of objects.

components/o-layout/src/tsx/landingPartials.tsx
akomiqaia: [e25ba6f](https://github.com/Financial-Times/origami/pull/898/commits/e25ba6f314295805abc92df5c27d6edfb55a6274)

akomiqaia:
From the last discussion, we spoke about improving this PR. Changes discussed were:

- Hide controls that are not in use, remove hero controls ([e081ce7](https://github.com/Financial-Times/origami/pull/898/commits/e081ce74ad7d68c6d7b8a8aaa5fb6504ba7e99cd))
- Storybook demos with hero layout had the full-width grey background, that comes from [this style definition](https://github.com/Financial-Times/origami/blob/88b56eb5a8b7273f0a2001dc3ee420daa157d2f4/components/o-layout/src/scss/_grid-areas.scss#L47). At the moment stories that use hero are rendered at the fullscreen layout (padded by default) ([e081ce7](https://github.com/Financial-Times/origami/pull/898/commits/e081ce74ad7d68c6d7b8a8aaa5fb6504ba7e99cd))
- rename and then remove add styles control ([09c2034](https://github.com/Financial-Times/origami/pull/898/commits/09c2034b2e5958ef0b439e51f67d71dc2372975d))

notlee (approved):

Issue #902, opened 2022-11-25
o-header: user can leave open drawer accidentally and navigate content behind


notlee:
When a user of screen reading assistive technologies navigates through the expanded menu using the arrow keys, they are not trapped within the menu and can leave the menu returning to the page. However, this only occurs when navigating back up the page, as the menu has been situated at the end of the DOM. 

*See how we trap focus in o-overlay*

https://docs.google.com/document/d/17-d7DAAORFCaUbiHdFPqvrBerUKbMyhG/edit#bookmark=id.8nyxjz95mzek
https://financialtimes.atlassian.net/browse/CON-2116

![image-20221012-105433](https://user-images.githubusercontent.com/10405691/204035227-5930cd18-2127-4869-88e8-2962ca497d00.png)

akomiqaia:
I am a bit confused about this one. The DAC comment contradicts Screen reader comments. And your comment is about how content is listed. 929f9d1 commit adds aria-labels to listed content so screen readers will announce headings of the list items in the drawer. I could not recreate the issues mentioned about leaving navbar or not. focus seems to be trapped within the drawer and it can be closed by pressing ESC.

notlee:
I've corrected the description of this issue. It was an amalgamation of 2 issues 🙈 The issue here is that navigating the drawer backward using the keyboard (e.g. shift tab) leaves the open drawer and starts navigating the page behind. Focus should stay in the drawer until it's closed. Here's how we do that in o-overlay: https://github.com/Financial-Times/origami/blob/09adefeea8bd8b623da37fe8251fc3b9edbc379a/components/o-overlay/src/js/overlay.js#L346

https://user-images.githubusercontent.com/10405691/205663604-09d39c84-41cc-4d42-a789-3270753be52d.mp4

PR #903, opened 2022-11-28
backstage: remedy o header DAC issues

### LINKED ISSUE #901: o-header: inconsistent label for navigation menu
The menu button provided for users in the header, has been labelled as ‘open side navigation menu’. Once the menu is opened, the close button is labelled as ‘Close drawer menu’. This can be confusing for users of screen reading assistive technologies, as they are not informed that both components refer to the same menu. This can be increasingly difficult to discern when navigating the page out of context using the forms list using the names of forms alone.

*for this to be classed as done, we need to update origami demos and work with the platforms team to update pagekit*

https://docs.google.com/document/d/17-d7DAAORFCaUbiHdFPqvrBerUKbMyhG/edit#bookmark=id.9ropi7us8epm

image-20221012-144826

akomiqaia:
update o header close button label and title so it matches open menu label; closes #901

### DIFF
--- a/components/o-header/demos/src/header.mustache
+++ b/components/o-header/demos/src/header.mustache
@@ -172,35 +172,35 @@
 		{{/editions}}
 
 		{{#nav}}
-			<nav class="o-header__drawer-menu o-header__drawer-menu--primary">
-				<ul class="o-header__drawer-menu-list" {{#heading}}aria-label="{{heading.name}}"{{/heading}}>
-							{{#heading}}
-							<li class="o-header__drawer-menu-item o-header__drawer-menu-item--heading">{{name}}</li>
-							{{/heading}}
-							{{#items}}
-							<li class="o-header__drawer-menu-item {{#divide}}o-header__drawer-menu-item--divide{{/divide}}">
-								{{#hasChildren}}
-									<div class="o-header__drawer-menu-toggle-wrapper">
-										<a class="o-header__drawer-menu-link o-header__drawer-menu-link--parent o-header__drawer-menu-link--{{#selected}}selected{{/selected}}{{^selected}}unselected{{/selected}}" href="{{href}}" {{#selected}}aria-label="{{name}}, current page" aria-current="page"{{/selected}}>{{name}}</a>
-										<button class="o-header__drawer-menu-toggle o-header__drawer-menu-toggle--{{#selected}}selected{{/selected}}{{^selected}}unselected{{/selected}}" aria-controls="o-header-drawer-child-{{index}}">
-											Show more {{name}} links
-										</button>
-									</div>
-									<ul class="o-header__drawer-menu-list o-header__drawer-menu-list--child" id="o-header-drawer-child-{{index}}">
-										{{#children}}
-										<li class="o-header__drawer-menu-item">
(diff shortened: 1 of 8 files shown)


components/o-header/demos/src/header.mustache
```diff
@@ -171,37 +171,37 @@
...
-							{{/hasChildren}}
-						</li>
-						{{/items}}
-				{{/nav}}
-			</ul>
-		</nav>
+		{{#nav}}
+			<nav class="o-header__drawer-menu o-header__drawer-menu--primary">
+				<ul class="o-header__drawer-menu-list" {{#heading}}aria-label="{{heading.name}}"{{/heading}}>
+							{{#heading}}
```
notlee, line 177: Breaking this menu up into multiple smaller menus seems sensible!

1. Where did this come from? It's not part of https://github.com/Financial-Times/origami/issues/901?
2. Would it make more sense to move the `aria-label` to the `nav` element instead, for users navigating by landmark?


Screenshot 2022-11-30 at 17 05 45

notlee:
We're going to need to update page-kit (or ask the platform team to) so we can rollout these changes to ft.com. We'll also want to update the [navigation service example](https://www.ft.com/__origami/service/navigation/v2/docs/example), which some people copy markup from. I'll create an issue for that before merging this :)

notlee:
https://github.com/Financial-Times/origami/issues/907

components/o-header/demos/src/header.mustache
akomiqaia: not sure what you mean by the first question. I also worked on [this issue](https://github.com/Financial-Times/origami/issues/902).

for the second one, you are right. We should move it to the `nav` tag.

akomiqaia: one thing is that when I move `aria-label` to nav tag VoiceOver did not announce there was a list of items. but if we keep the `aria-label` on the `ul` tag then the voiceover reads the heading and announces that there are 11 (for example) items in this list.

notlee: Good point. Taking another look at this I see them as all one `nav`, but with multiple sections within that nav. What do you think about going back to one `nav` element for these and but splitting into multiple `ul` lists with a `aria-label`?

notlee: In fact, forget that.

The first list item isn't really an item of the nav – it's not a link, and more of a heading – so it shouldn't be in the list at all. Looks like we should use a `h2` heading level. DAC recommended this in their report and there's also a [WAI guide to navigation here](https://www.w3.org/WAI/tutorials/menus/structure/#label-menus) which does the same.

Maybe something like this?

```html
<nav>
	<h2 id="o-header-drawer-top-sections">Top sections</h2>
	<ul aria-labelledby="o-header-drawer-top-sections">
	    <!-- list items -->    
	</ul>
	<h2 id="o-header-drawer-top-sections">FT recommends</h2>
	<ul aria-labelledby="o-header-drawer-top-sections">
	    <!-- list items -->    
	</ul>
</nav>
<nav>
	    <!-- secondary stuff -->    
</nav>
<nav>
	    <!-- user stuff -->    
</nav>
```
drawer-markup

notlee: I've created a new issue to capture this:
https://github.com/Financial-Times/origami/issues/910

The [previous issue](https://github.com/Financial-Times/origami/issues/902) was about something else, but included the wrong description. 🙈

notlee:
👍 Merging to the WIP v11 branch. Solves [inconsistent labels](https://github.com/Financial-Times/origami/issues/910), more work is required to markup [visual navigation headings](https://github.com/Financial-Times/origami/issues/910)

PR #905, opened 2022-11-30
O header services

akomiqaia:


components/o-header-services/src/tsx/header-services.tsx
```diff
@@ -0,0 +1,97 @@
+import {PrimaryNav, SecondaryNav, SecondaryNavProps, NavItem, ListItem} from './navs';
+
+type TitleProps = {
+	title: string;
+	tagline?: string;
+	titleUrl?: string;
+	relatedContent?: ListItem[];
+	primaryNav?: boolean;
+	secondaryNav?: boolean;
```
notlee, line 9: I don't think these 2 parameters should be needed. They affectively give the user control over whether to show the burger menu or not, but it shouldn't be up to the user in this case, it should be up to the component to decide that based on navigation data provided.

Removing those 2 props will make the api simpler for users, and improve our ability to maintain the tsx.

1. Instead have you considered passing `primaryNavData` and `secondaryNavData` props to `Title`, so it can decide whether to display the burger menu based on their value?
 
2. An alternate approach could be to move the `hasHamburgerMenu` conditional to `HeaderServices`, and pass `hasHamburgerMenu` as a prop to `Title`. I think approach 1. above may be better as it encapsulates the logic for deciding if to show the burger icon in the `Title` component, which has the burger icon.

notlee:
Looks good :) One small suggestion to simplify the API a bit. Also I made a [commit to remove the padding around the story](https://github.com/Financial-Times/origami/pull/905/commits/adbcfcdaedc0672c29ed2153f860a4d0367d5fc7), so our header component demos go right to the edge (don't forget to git pull!)

components/o-header-services/src/tsx/header-services.tsx
akomiqaia: [e3dc2ee](https://github.com/Financial-Times/origami/pull/905/commits/e3dc2ee84592f27ed0ad1cca3989dee851226b3f) resolves this using the first approach

notlee:
@akomiqaia Nice 👍 
 
 I noticed on a second look that we could do a [similar thing for the dropDown prop](https://github.com/Financial-Times/origami/pull/905/commits/ab82e6384a85cde715fe747d8e06e2b2181ddd9e). I also realised [`theme` could be a better prop name than `modifier`](https://github.com/Financial-Times/origami/pull/905/commits/3afeda8f69a80b57b96ed6d7ce34ba773b13fae2) – modifier references the BEM naming convention (block, element, modifier), a modifier could be [all sorts of things including themes](https://origami.ft.com/documentation/components/customisation/#variants). 
 
 I went ahead and mind those changes whilst you were at lunch. Please take a look and let me know what you think, feel free to merge if you think those changes are okay 😄

PR #914, opened 2022-12-08
backstage: fix issues related to storybook controls.

akomiqaia:
This one was an interesting bug. In the end, ended up using [controls with complex values approach](https://storybook.js.org/docs/react/writing-stories/args#mapping-to-complex-arg-values). This approach could also be used in many different ways, such as using dropdown to show only supported components in certain scenarios.

The main issue was that we could not see some stories controls and it turned out that when defining `argTypes` for the storybook it didn't like using `''` as the label key. For example:
```js
argTypes: {
	theme: {
		control: {
			type: 'select',
			labels: {
				b2b: 'B2B',
				b2c: 'B2C',
				'': 'Default',
			},
		},
	},
},
```
using `''` as a label key made some other stories to error in control panel and stop displaying controls that were there before.


components/o-header-services/stories/header-services.stories.tsx
```diff
@@ -29,20 +33,16 @@ export default {
...
+			options: Object.keys(ThemeMapping),
+			mapping: ThemeMapping,
+			control: 'select',
 		},
 	},
 } as ComponentMeta<typeof HeaderServices>;
 
 const HeaderServicesStory = args => {
 	useEffect(() => void js.init(document.body), []);
+	console.log(args);
```
notlee, line 45:
```suggestion
```

notlee (approved):
Excellent :) Just one console.log to remove

PR #923, opened 2022-12-14
[CI-1411] Deep Portrait: topper name changed

AlbertoCuberoNavas:
[ticket](https://financialtimes.atlassian.net/browse/CI-1411)


notlee (approved):
This is technically a major / breaking change as we're removing `--split-text-portrait-left`, but I think it's sensible to release this as a patch since you're rolling this out and it won;t have been adopted elsewhere yet

AlbertoCuberoNavas:
> This is technically a major / breaking change as we're removing `--split-text-portrait-left`, but I think it's sensible to release this as a patch since you're rolling this out and it won;t have been adopted elsewhere yet

yes, right now it's not used in any place, so its safe to update it like a patch

PR #933, opened 2022-12-20
fix: require o-normalise 3.3.0 or higher

notlee:


components/o-expander/package.json
```diff
@@ -29,11 +29,11 @@
   },
   "peerDependencies": {
     "@financial-times/o-icons": "^7.0.0",
-    "@financial-times/o-viewport": "^5.0.0",
-    "@financial-times/o-normalise": "^3.2.0"
+    "@financial-times/o-normalise": "^3.2.0",
```
juanSanchezAlcala, line 32: I think this doesn't have much sense to have it on peerDependencies since we are importing a new version on devDependencies . What do you think ?

components/ft-concept-button/package.json
```diff
@@ -19,8 +19,8 @@
     "watch": "bash ../../scripts/component/watch.bash"
   },
   "peerDependencies": {
-    "@financial-times/o-typography": "^7.2.2",
-    "@financial-times/o-normalise": "^3.2.2"
+    "@financial-times/o-normalise": "^3.3.0",
```
juanSanchezAlcala, line 22: Here is as peerDependency but in most of other components is as devDependency , does it make sense have it as peerDependency here?

components/o-comments/package.json
```diff
@@ -29,7 +28,7 @@
     "@financial-times/o-colors": "^6.0.1",
     "@financial-times/o-editorial-typography": "^2.0.1",
     "@financial-times/o-forms": "^9.2.0",
-    "@financial-times/o-normalise": "^3.0.0",
+    "@financial-times/o-normalise": "^3.3.0",
```
juanSanchezAlcala, line 31: same here as first comment

components/o-ft-affiliate-ribbon/package.json
```diff
@@ -25,7 +25,7 @@
   "peerDependencies": {
     "@financial-times/o-colors": "^6.0.1",
     "@financial-times/o-grid": "^6.0.0",
-    "@financial-times/o-normalise": "^3.0.0"
+    "@financial-times/o-normalise": "^3.3.0"
```
juanSanchezAlcala, line 28: same as first comment

juanSanchezAlcala (changes requested):

notlee:
@juanSanchezAlcala we use `devDependencies` to indicate demo dependencies

juanSanchezAlcala (approved):

PR #940, opened 2022-12-21
backstage: trap focus in overlay using role=dialog and aria-modal=true

akomiqaia:
This issue was discovered while resolving #902. In both cases, we use JS to trap the focus and we listen if the tab key is pressed. For voiceOver the user usually will use `ctrl-option-[left/right] Arrow` and focus would leave the modal. To fix this we followed [web standard recommendations](https://www.w3.org/WAI/ARIA/apg/patterns/dialogmodal/). 

One to note is that once the overlay modal is open it is not automatically focused.

### DIFF
--- a/components/o-overlay/src/js/overlay.js
+++ b/components/o-overlay/src/js/overlay.js
@@ -65,16 +65,36 @@ const isVisible = function (element) {
 	return Boolean(element.offsetHeight);
 };
 
-const focusTrap = function (event ) {
+const filterFocusableElements = function (element) {
+	const elementVisible = isVisible(element);
+	// Inputs for radio and checkboxes are visually hidden,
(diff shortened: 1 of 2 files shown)


notlee (approved):

notlee:
Looks like there are some conflicts to fix here @akomiqaia, but otherwise looks good 👍

akomiqaia:
@notlee  what conflicts? :D

notlee:
@akomiqaia Oh, only in a previous commit – probably the [lock file change](https://github.com/Financial-Times/origami/pull/940/commits/dab757efe510451d2706428d07edeb27a4ac1857#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519). It's fine to squash though :)

PR #948, opened 2023-01-04
Use `rgba` for `rgba` value

chee:


umbobabo (approved):
Thank you Chee, good spot.

PR #951, opened 2023-01-05
930 o share icons

### LINKED ISSUE #930: o-share icons invisible with ZoomText Colour Inversion
See https://financialtimes.atlassian.net/browse/CI-1373

akomiqaia:
This PR changes how users can use the `o-share` component. Before it was possible to generate a URL based on the config provided but now this should be implemented as part of the HTML markup that users write. To remedy this DAC issue we had to inline our SVGs and apply `fill: currentColor` so the inverse theme would work on high contrast.

This approach will make `o-share` out of sync with `image-service` and the current solution is a temporary fix. We are working on [proposal](https://docs.google.com/document/d/1r1-7COAabIoJZSAIAu7vZne-SzcUqNGNSh7HlAlyVZY/edit#heading=h.ft9lhce3g3o9) that will automate syncing between o-share and image-service.

closes #930


components/o-share/demos/src/data.json
```diff
@@ -0,0 +1,39 @@
+{
+	"socialNetworks": [
+		{
+			"name": "Twitter",
+			"className": "twitter",
+			"link": "https://twitter.com/intent/tweet?url=http%3A%2F%2Fon.ft.com%2F1mUdgA2&text=US%20drugs&related=ftcompanies&via=FT",
+			"svgPath": "<path d='M417 720c193.2 0 298.9-160.1 298.9-298.9 0-4.5 0-9.1-.3-13.6 20.6-14.9 38.3-33.3 52.4-54.4-19.2 8.5-39.5 14.1-60.3 16.5 21.9-13.1 38.3-33.8 46.2-58.1-20.6 12.2-43.2 20.9-66.7 25.5-39.8-42.3-106.3-44.3-148.6-4.6-27.3 25.7-38.9 63.9-30.4 100.4-84.5-4.2-163.2-44.1-216.5-109.8-27.9 48-13.6 109.4 32.5 140.2-16.7-.5-33.1-5-47.7-13.1v1.3c0 50 35.3 93.1 84.3 103-15.5 4.2-31.7 4.8-47.4 1.8 13.8 42.8 53.2 72.1 98.1 72.9-37.2 29.2-83.1 45.1-130.5 45.1-8.4 0-16.7-.5-25-1.5 48 31 103.9 47.3 161 47.3'/>"
```
notlee, line 7: A future automated solution would be complicated if we had to modify the svg in some way. If we wrap the svg in an `o-share__icon__image` element are we able to inline the svg unmodified, and without separating the paths out here?

Along the lines of: 
```html
<div class"o-share__icon__image">
    <svg></svg>
</div>
```

```css
.o-share__icon__image > svg {
    fill: currentColor;
}
```

notlee:
This will be a major release so don't forget to write a MIGRATION guide. Don't forget to remind the reader that they can use the tsx, and if not can use Storybook to generate markup for the options they select

components/o-share/demos/src/data.json
akomiqaia: [23f6f47](https://github.com/Financial-Times/origami/pull/951/commits/23f6f47c61ea91fafb5bdb9a9b9845cbab2f2aae)

components/o-share/src/tsx/shareIcon.tsx
```diff
@@ -0,0 +1,158 @@
...
+				<span className="o-share__text" data-variant-label>
+					{label}
+				</span>
+			</button>
+		</form>
+	);
+}
+
+function generateDescriptiveLinkText(title: string, socialNetwork: IconType) {
+	// It seems like next article is not using pinterest and not sure how link or enterprise sharing is used and we might want to add something for the save button
```
akomiqaia, line 122: @notlee this is the part I am not sure about and a similar thing for line 148.

components/o-share/src/tsx/shareIcon.tsx
```diff
@@ -0,0 +1,158 @@
...
+		pinterest: `Share ${title} on Pinterest (opens a new window)`,
+		whatsapp: `Share ${title} on Whatsapp (opens a new window)`,
+		link: `Open link in new window`,
+		enterpriseSharing: `Share ${title} with your Enterprise Sharing tools (opens a new window)`,
+	};
+	return descriptiveLinkText[socialNetwork];
+}
+
+function generateSocialUrl(
+	config: Record<string, string>,
```
frshwtr, line 136: Declaring a more specific type here will make this function safer.

```tsx
interface SocialUrlConfig {
url: string;
title: string;
...
}

function generateSocialUrl(
    config: SocialUrlConfig
   ...
```

components/o-share/src/tsx/shareIcon.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+function CustomIcon({
+	icon,
+	label,
+	customAction,
+	title,
+	summary,
+}: CustomIconProps) {
+	return (
+		<form method="post" action={customAction}>
```
frshwtr, line 105: I think I need some more clarity on how this will work for consumers. Is the intention that a consumer will provide a ref to some custom action endpoint which will then trigger an action in their page?

components/o-share/src/scss/_mixins.scss
```diff
@@ -39,10 +39,10 @@
 			border-color: oColorsByUsecase($usecase, 'border');
 			background-color: oColorsByUsecase($usecase, 'background');
 			color: oColorsByUsecase($usecase, 'text');
-			&:before {
-				$color-usecase-encoded: str-slice(ie-hex-str(oColorsByUsecase($usecase, 'text')), 4);
-				background-image: url($service-url + $query + "&format=svg&tint=#{$color-usecase-encoded}");
-			}
+			// &:before {
```
frshwtr, line 42: Should this still be here?

components/o-share/src/tsx/shareIcon.tsx
```diff
@@ -0,0 +1,164 @@
...
```
notlee, line 103: I think we can remove `CustomIcon` and the `form` for now? Including from documentation

components/o-share/src/scss/_mixins.scss
```diff
@@ -26,10 +30,6 @@
 		$query: "?source=#{$system-code}";
 		$service-url: "#{$o-share-image-service-base-url}/#{$o-share-image-service-version}/images/raw/#{$scheme}:#{$icon-name}";
```
notlee, line 31: There's more we can delete here, we should no longer need to output and image for any state

components/o-share/MIGRATION.md
```diff
@@ -1,6 +1,95 @@
...
-### Migrating from v7 to v8
+## Migrating from v8 to v9
+
+Version 9 has markup changes and component API got updated as well. New version relies less on javaScript and improves accessibility on high contrast mode. The most important updates are:
+
+1. The icons are now inlined inside the templates, since this was our only known option to [fix accessibility issue](https://github.com/Financial-Times/origami/issues/930) on high contrast mode devices.
+2. Share link MUST be written in full markup instead of using js to automatically generate share.
+3. New TSX template requires social icons to be passed as children.
+
+New markup can be copied from Origami [registry](https://registry.origami.ft.com/components/o-share) or Build service users can use [storyBook demos](http://localhost:6969/?path=/story/components-o-share--horizontal) to generate markup based on storybook controls, that can be copied from `HTML` tab.
```
notlee, line 11: Correcting a link here:
```suggestion
New markup can be copied from the [Origami registry](https://registry.origami.ft.com/components/o-share) or [StoryBook](https://origami.ft.com/storybook/) via the`HTML` tab.
```

components/o-share/MIGRATION.md
```diff
@@ -1,6 +1,95 @@
 # Migration guide
 
-### Migrating from v7 to v8
+## Migrating from v8 to v9
+
+Version 9 has markup changes and component API got updated as well. New version relies less on javaScript and improves accessibility on high contrast mode. The most important updates are:
+
+1. The icons are now inlined inside the templates, since this was our only known option to [fix accessibility issue](https://github.com/Financial-Times/origami/issues/930) on high contrast mode devices.
+2. Share link MUST be written in full markup instead of using js to automatically generate share.
+3. New TSX template requires social icons to be passed as children.
```
notlee, line 9: This list describes what has changed and why which can be helpful context but could also distract from the steps a user needs to take to migrate. Could these be updated to be a kind of summary of steps which are detailed below?

```suggestion
o-share v9 incudes a number of changes to improve accessibility and the flexibility of its TSX templates. To migrate follow these steps and see below for more details:

1. If your project uses o-share TSX templates, migrate to the new template interface.
2. o-share no longer provides client side JavaScript to generate markup. If your project does not use o-share TSX templates ensure you output o-share markup in full. 
```

components/o-share/MIGRATION.md
```diff
@@ -1,6 +1,95 @@
...
+
+New markup can be copied from Origami [registry](https://registry.origami.ft.com/components/o-share) or Build service users can use [storyBook demos](http://localhost:6969/?path=/story/components-o-share--horizontal) to generate markup based on storybook controls, that can be copied from `HTML` tab.
+
+### Inlined SVG icons
+
+Version 8 was utilising `:before` selector and icon url was dynamically set using scss. This approach was not working on high contrast mode devices. We have decided to inline SVG icons instead. This is a breaking change since we have removed `:before` pseudo element and in version 9 we use following structure for `o-share` items:
```
notlee, line 15: This  describes what has changed and not what a user needs to do. Do they need to know we used to use the `:before` selector?

components/o-share/MIGRATION.md
```diff
@@ -1,6 +1,95 @@
...
+    icon="share"
+    showLabel={true}
+    label="Share"
+    customAction="#"
+    urlProps={shareIconProps}
+   />
+  </Share>
+```
+
+This example should render twitter social share icon and custom share button. More details about the props that you must pass down to each compnent can be found in our [readme](README.md#tsx-template).
```
notlee, line 88: We removed custom share buttons for now

components/o-share/MIGRATION.md
```diff
@@ -1,6 +1,95 @@
...
+ data-o-share-titleExtra="{{titleExtra}}"
+ data-o-share-summary="{{summary}}"
+ data-o-share-relatedTwitterAccounts="{{relatedTwitterAccounts}}"
+ data-o-share-location="{{locationOfShareComponent}}">
+</div>
+```
+
+Instead use full markup of the component:
+
+```html
```
notlee, line 53: This is a helpful example 👍

components/o-share/MIGRATION.md
```diff
@@ -1,6 +1,95 @@
...
+    <span class="o-share__text">
+     Share {{title}} on {{name}} (opens a new window)
+    </span>
+   </a>
+</li>
+```
+
+### Deprecate autogenerate share links
+
+Before version 9 users could automatically generate share links with following markup. But this requires JavaScript and from our lates DAC audit we determined that this approach could be deprecated.
```
notlee, line 36: This isn't quite true. We removed this feature as there were multiple ways to use the component which is difficult to maintain. We felt that tsx templates provide a reasonable alternative for users who do not want to manage markup. But probably we don't need quite so much detail in a migration guide away 😄 That might be better for the commit description and changelog

components/o-share/src/scss/_mixins.scss
frshwtr: @notlee Does this mean we no longer need to pass [these icons in as input](https://github.com/Financial-Times/origami/pull/951/files#diff-d0e99b9bcf76fd8eaf9ed1a2cb180ce9d0e6c1fed88aee299f83d49d1241a9a3R40)?

notlee: Good spot / thought. Perhaps we do as each icon has different colours on hover, and I assume we're still setting these colours. So that parameter would still allow a user to reduce the number of CSS selectors output, but the benefit would be much more limited.

frshwtr: Interesting. When I viewed in Storybook, they all had black bgs, I'll double check.

components/o-share/MIGRATION.md
frshwtr: I agree we can simplify this to just the information the user needs. @notlee Is there a decision log somewhere which we can reference should the user wish to read up on our decision to use SVGs?

notlee: Do you have a formal process/doc in mind regarding how a decision log should work? We don't have anything quite like that, but do document decisions in a few places depending on what they are.

I think for this I'd propose putting this information in the commit message and PR message of the change. It could also make an interesting blog post but in addition to, as a blog post isn't very discoverable when working on/with o-share. For significant or cross-component changes we'd usually have an issue with label `proposal` which would log a decision and the conversation around that.

frshwtr: In the past we've logged them on a `.md` in a repository which served as a hub for a given team (for us it would be something like `origami-team-home`. Notion is also a good option that can give some additional features such as filtering based on certain properties.

I'm thinking because the team is quite small, we might not need a hugely formal DL as we should be able to share knowledge and be on the same page easier than a larger team. Though DLs have use for external/in-source contributors.

For now we can log this in the commit and PR. Would this be the release please PR?

notlee (approved):
Looks good, thanks both! 🎉

components/o-share/MIGRATION.md
notlee: This commit and PR, it would be easier for a future person to `git blame` their way here

notlee:
Wait! @frshwtr [no tsx was ever publicly published for this component](https://github.com/Financial-Times/origami/pull/951/files#diff-0e109823ebdd3c35f03342ebe085eabc67037dd1aefa0a3c8990be4566825373L11). We can remove mention of TSX template changes from the migration guide :)

PR #952, opened 2023-01-09
backstage: introduce o-ft-affiliate-ribbon Storybook component

frshwtr:


components/o-ft-affiliate-ribbon/stories/ft-affiliate-ribon.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+} as ComponentMeta<typeof FtAffiliateRibbon>;
+
+export const StoryFtAffiliateRibbon: ComponentStory<typeof FtAffiliateRibbon> = () => {
+	return (
+		<FtAffiliateRibbon />
+	);
+};
+StoryFtAffiliateRibbon.storyName = 'Default FT Affiliate Ribbon';
+
+const FtAffiliateRibbon = () => {
```
frshwtr, line 23: I wasn't sure if this was the right place for this particular component. It seems most components have their html initialised through js code. Would this be the case for a component that (currently) exists solely as scss code when migrating?

notlee (approved):
Nice, thanks! One additional thing: We'll want to update[ .npmignore so src/tsx is published](https://github.com/Financial-Times/origami/blob/a206524b254d49474579459cd4e40c4920671e71/components/o-ft-affiliate-ribbon/.npmignore#L11). We don't publish tsx by default as there was a time we were using them experimentally only for demos, and wanted an opportunity to review before making public. Removing should be a `feat: ` commit, which will trigger a patch release on merge :)

PR #956, opened 2023-01-12
Convert unitless values into percentages in oColorsMix

oliverturner:
## Description
This PR ensures that `oColorsMix` calls the sass-native `mix` function with the required % unit for its `$weight` argument.

## Context
`sass@^1.52.3` raises warnings for any calls to `mix` made with a unitless value. In a project like ft-app this results in the terminal becoming flooded with aggressive red messages.


notlee (approved):
Thanks :D

PR #961, opened 2023-01-16
fix: o-video track comment

juanSanchezAlcala:
comment progress video track on 100%.


components/o-video/src/js/video.js
```diff
@@ -55,6 +55,7 @@ const dispatchedProgress = {};
 function shouldDispatch(video) {
 	const progress = video.getProgress();
 	const id = video.opts.id;
+	//100% progress doesn't seem to be lunch in chrome
```
notlee, line 58: 🥪 😄 
```suggestion
	//100% progress doesn't seem to be launched in chrome
```

notlee (approved):

PR #964, opened 2023-01-16
fix: o-header, fix tsx errors

notlee:
- Do not apply modifier classes which are falsely;
- Swap undefined `name` variable for `label`.

Sorry Prettier got to this and included other formatting changes.


components/o-header/stories/header.stories.tsx
```diff
@@ -1,34 +1,35 @@
...
-		currentPath: '/',
+		currentPath: "/",
 	},
 	argTypes,
 	decorators: [withDesign, withHtml],
 } as ComponentMeta<typeof MainHeader>;
 
 export const HeaderPrimary: ComponentStory<typeof MainHeader> = args => {
 	useEffect(() => void javascript.init(), []);
+	console.log(args);
```
frshwtr, line 32: Should this be here?

notlee: Oops, it should not be

components/o-header/src/tsx/drawer.tsx
```diff
@@ -254,13 +266,15 @@ function DrawerSubMenu({
 				/>
 				<button
 					className="o-header__drawer-menu-toggle o-header__drawer-menu-toggle--unselected"
-					aria-controls={`o-header-drawer-child-${idSuffix}`}>
-					Show more {name} links
+					aria-controls={`o-header-drawer-child-${idSuffix}`}
+				>
+					{`Show more ${label}`}
```
notlee, line 271: `name` replaced with `label`

components/o-header/src/tsx/top.tsx
```diff
@@ -9,16 +9,17 @@ export function HeaderWrapper({
 	additionalClassName?: string;
 	children: JSX.Element[] | JSX.Element;
 }) {
-	const headerClassNames = `o-header o-header--${
-		variant || 'simple'
-	} ${additionalClassName}`;
+	const headerClassNames = `o-header o-header--${variant || "simple"} ${
+		additionalClassName ? additionalClassName : ""
```
notlee, line 13: `additionalClassName` only applied if truthy

notlee: Which I wouldn't expect? In my case I was using esbuild and React Server to generate o-header markup. We need tests for these...

notlee: Oh I would expect because the string interoperation here

frshwtr (approved):

PR #973, opened 2023-01-25
feat: introduce o-editorial-layout typescript templates

frshwtr:


akomiqaia:
@frshwtr LGTM. 

Just one thing. What do you think about instead of nesting demos maybe we can list them under o-editorial-layout? each nested folder contains only one demo and their names are quite descriptive. 
Screenshot 2023-01-25 at 11 56 34

frshwtr:
> @frshwtr LGTM.
> 
> Just one thing. What do you think about instead of nesting demos maybe we can list them under o-editorial-layout? each nested folder contains only one demo and their names are quite descriptive. Screenshot 2023-01-25 at 11 56 34

Happy to do this :)

frshwtr:
@akomiqaia Looks like i forgot to push the change up earlier 😅  It's ready for re-review (slack is also isn't working for me right now)

akomiqaia (approved):

components/o-editorial-layout/stories/editorial-layout-heading.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export default {
+	title: 'Components/o-editorial-layout',
+	component: EditorialLayoutHeading,
+	decorators: [withDesign, withHtml],
+} as ComponentMeta<typeof EditorialLayoutHeading>;
+
+const Template = args => {
+	return (
+		<EditorialLayoutHeading {...args}>Documentation Heading</EditorialLayoutHeading>
```
notlee, line 15: o-editorial-layout builds on [o-editorial-typography](https://registry.origami.ft.com/components/o-editorial-typography) to provide margin and layout for editorial typography, for specific contexts such as an article page. 

I think this demo isn't very helpful because it shows a single element and no relationship to other elements. o-editorial-layout currently provides margins between typographical elements, asides, and figures.

In this case it might be better to re-create the existing demos more verbatum, to show these relationships?
https://registry.origami.ft.com/components/o-editorial-layout@2.3.2

One demo shows a wrapper class that styles html by semantic type select `h1`, `h2`, etc. 
The second demo shows how classes on each element can be used instead. This may be preferable where a user needs more control on where styles are applied, to avoid conflicts with other components within editorial content for example.

components/o-editorial-layout/src/tsx/editorialLayoutBody.tsx
```diff
@@ -0,0 +1,6 @@
+interface EditorialLayoutBodyArgs {
```
notlee, line 1: FYI we use kebab case for file names.

The components themselves look good though :)

notlee (changes requested):

components/o-editorial-layout/stories/editorial-layout-heading.stories.tsx
notlee: "Documentation Heading" may be a bit misleading too when we intend this component is used for editoiral content

components/o-editorial-layout/src/tsx/editorialLayoutBody.tsx
notlee: Oh! The other thing is this only supports the core brand, not intenral or whitelabel brands. We manage branded stories using a directory structure. Checkout [o-layout for an example](https://github.com/Financial-Times/origami/tree/main/components/o-layout/stories), which puts demos for the intenral brand within an `internal` stories sub-directory. We'll want a `core` directory in this case

components/o-editorial-layout/stories/editorial-layout-heading.stories.tsx
frshwtr: @notlee I can pull a headline from today's top story.

It might be nice to see if we can add a control which passes down to the children so the user can control it.

PR #975, opened 2023-01-25
fix: use kebab tsx naming, move stories to core/

frshwtr:


frshwtr:
old files did not delete, do not merge

frshwtr:
let's try that again

PR #978, opened 2023-01-26
fix: use correct conventions for naming tsx components, docs: move components to core, use verbatim examples from resgistry

frshwtr:
This PR will:
* Use kebab case for `o-editorial-layout` TSX components instead of camel.
* Move `o-editorial-layout` Storybook demos to a `core` sub-directory. So that they only show up as part of the `core` brand.
* Use verbatim examples from the [component registry](https://registry.origami.ft.com/components/o-editorial-layout@2.3.2) to show a clearer distinction between using the wrapper and the individual classes.


notlee (approved):

akomiqaia (approved):
Looks good. But moving everything under core folder is not strictly necessary in this case. Our storybook config still respects `origami.json` config which defines `o-editorial-layout` only for core brand. 

I think we introduced folder structured for complex situations where we could not use same controls for different brands. or some TSX templates would need particular props for different brands. Good example of this will be `o-message`. But if you look at `o-banner` it only supports core but nesting is not needed since it's just one brand. The rest is taken care by `origami.json`. 

In the end both options are fine. I guess we never agreed what would be teams practice around this. One thing that can stay out of the brand folders is maybe `scss` file. If there two brands they would use same `scss` file and mixin should figure out what brands what styles would use.  But again we have only one brand support and  I guess it can stay in core folder.

PR #981, opened 2023-01-27
O multi select [OR-16]

akomiqaia:


.prettierignore
```diff
@@ -1,4 +1,4 @@
-components/
+# components/
```
frshwtr, line 1: This might need uncommenting.

akomiqaia: yeah, will do that. Just need to keep it until PR is finished.

components/o-multi-select/main.scss
```diff
@@ -0,0 +1,90 @@
...
+/// Output all oMultiSelect features
+/// @param {Map} $opts [()] - A map of options to configure the output
+/// @access public
+/// @example scss
+///		@include oMultiSelect($opts: (
+///			// your opts here
+///		))
+@mixin oMultiSelect($opts: ()) {
+	// @include _oMultiSelectBase();
+	.o-multi-select_enhanced,
```
frshwtr, line 19: Should this be in kebab case like the other selectors on this rule?

akomiqaia: For mixins we use camelCase. Here is more [details about naming conventions](https://origami.ft.com/specification/v1/components/sass/#naming-conventions)

frshwtr: Ah sorry I'm referring to `.o-multi-select_enhanced`. The other two selectors on lines 20 and 21 have kebab all the way through whereas the final separator in this selector is underscore.

components/o-multi-select/demos/src/demo.mustache
```diff
@@ -0,0 +1,68 @@
+<div class="o-multi-select" data-o-component="o-multi-select">
```
frshwtr, line 1: This particular demo seems to have some unexpected behaviours in VoiceOver around displaying the combo box options. I'm not too sure why though. Perhaps we could look into it together.

akomiqaia: sure. I will have a quick look today again.

components/o-multi-select/main.scss
akomiqaia: @frshwtr [d0c21ff](https://github.com/Financial-Times/origami/pull/981/commits/d0c21ff1e4f324c300e9a970a30f01699a03e6e5) commit I tried to follow BEM naming conventions.

components/o-multi-select/demos/src/demo.mustache
akomiqaia: with that last changes accessibility issues should be resolved.

frshwtr: Hmm I still can't seem to select options with VoiceOver 🤔

components/o-multi-select/src/js/multi-select-options.js
```diff
@@ -0,0 +1,101 @@
...
+	const button = this.selectedOptions.querySelector(`#${option + index}`);
+	button.parentElement.remove();
+	this._updateState();
+}
+
+/**
+ * adds a item in a multi-select selected list.
+ *
+ * @private
+ * @param {HTMLElement} optionEl - The option element to remove.
```
frshwtr, line 39: These params all mention removing the element, but the function name mentions adding.

components/o-multi-select/src/tsx/multi-select.tsx
```diff
@@ -0,0 +1,64 @@
...
+			<div className="o-form o-multi-select o-multi-select--core">
+				<span className="o-forms-field">
+					<span className="o-forms-title">
+						<label className="o-forms-title__main" htmlFor="multiple">
+							Multiple select box
+						</label>
+					</span>
+
+					<span className="o-forms-input o-forms-input--select">
+						<select name="multiple" id="multiple" multiple>
```
frshwtr, line 17: This might need a generator for the id. A screen reader might struggle if there's more than one multi select on the page.

components/o-multi-select/demos/src/demo.mustache
akomiqaia: I see. there are issues in Safari. But chrome seems fine.  I will investigate Safari issues.

components/o-multi-select/src/tsx/multi-select.tsx
akomiqaia: This markup is for our demos in the registry. The users can copy this markup and add IDs by themselves but it might be a good idea to add them in our demos.

components/o-multi-select/demos/src/demo.mustache
akomiqaia: This one seems a bit tricky. We followed the best [practices recommended](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/) and these examples don't work properly in Safari on MacOS. Not sure how much we can do here.

frshwtr: Hmm @notlee any thoughts on this? I'm not sure what browser usage amongst VoiceOver users is.

I have confirmed this works on Chomre and Safari. Firefox doesn't let you click into the multi select dropdown.

On both chrome and safari, the VoicOver readout doesn't make it clear that clicking the button for an option will remove it from the list, I'll leave another comment closer to where I think a change could be.

components/o-multi-select/src/js/multi-select-options.js
```diff
@@ -0,0 +1,106 @@
...
+ * @private
+ * @param {string} option - The text content of the option.
+ * @param {number} index - The index of the option.
+ * @returns {{ li: HTMLElement, button: HTMLElement }} An object containing the newly created <li> and <button> elements.
+ */
+function createOptionButton(option, index) {
+	const li = document.createElement('li');
+	const button = document.createElement('button');
+	button.id = `${option}-${index}`;
+	button.setAttribute('aria-label', ` remove ${option} `);
```
frshwtr, line 74: This doesn't seem to have been applied in the demo 🤔 
image

components/o-multi-select/src/js/multi-select.js
```diff
@@ -0,0 +1,162 @@
...
+	 * @param {HTMLElement} [multiSelectEl] - The component element in the DOM
+	 * @param {Object} [options={}] - An options object for configuring the component
+	 */
+	constructor(multiSelectEl, options) {
+		this.multiSelectEl = multiSelectEl;
+		this._clearCore();
+
+		this.options = Object.assign(
+			{},
+			{},
```
frshwtr, line 22: One of these blank objects can be removed.

components/o-multi-select/src/js/multi-select.js
```diff
@@ -0,0 +1,162 @@
...
+} from './utils.js';
+import {updateState} from './state.js';
+import {handleOptionSelect, createOption} from './multi-select-options.js';
+
+class MultiSelect {
+	/**
+	 * Class constructor.
+	 *
+	 * @param {HTMLElement} [multiSelectEl] - The component element in the DOM
+	 * @param {Object} [options={}] - An options object for configuring the component
```
frshwtr, line 14: Options appear to be required based on [L27-31](https://github.com/Financial-Times/origami/pull/981/files#diff-2318176be42718afde4d03f7c26429c218adaa82dd6ed2f0d90cf62bfa334c13R27-R31). As a user, it isn't clear to me what those options are. Could the options be defined here?

components/o-multi-select/src/tsx/multi-select.tsx
frshwtr: Is that the case for our Typescript components? I thought we export them so projects can use tsx templates if they like.

.prettierignore
akomiqaia: [59ae843](https://github.com/Financial-Times/origami/pull/981/commits/59ae843edf5107423a089f07df962968e883dfa4)

components/o-multi-select/src/tsx/multi-select.tsx
akomiqaia: other demos don't seem to use IDs. This is a good point but maybe we rely on other devs to wrap TSX templates in their tags? @notlee what do you think about this?

components/o-multi-select/src/js/multi-select-options.js
akomiqaia: is this chrome? all 3 browsers I have display `aria-labels`

components/o-multi-select/src/js/multi-select.js
akomiqaia: [b10f054](https://github.com/Financial-Times/origami/pull/981/commits/b10f0544cdf87175108e3e67208aab3b3726e04c)

components/o-multi-select/src/js/multi-select.js
akomiqaia: [c41fc40](https://github.com/Financial-Times/origami/pull/981/commits/c41fc40668b3765bc78f689dbd70fafb14239ba6)

components/o-multi-select/src/js/multi-select-options.js
frshwtr: This was chrome yes. I have tried again and it seems to work. I might have had an older commit checked out.

components/o-multi-select/test/multi-select.test.js
```diff
@@ -0,0 +1,484 @@
...
+	});
+
+	it('has a static init method', () => {
+		assert.isFunction(MultiSelect.init);
+	});
+
+	it('should autoinitialize', done => {
+		const initSpy = sinon.spy(MultiSelect, 'init');
+		document.dispatchEvent(new CustomEvent('o.DOMContentLoaded'));
+		setTimeout(function () {
```
frshwtr, line 38: Using `setTimeout` like this can introduce some flakiness through race conditions which might be problematic if these tests run on a slow system (though I imagine the system would need to be very slow for this to fail).

I think it's OK for now but in future it would be interesting to see if we can extend chai to have a behaviour similar to [waitFor](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor) where it will retry the test until it passes or hits a timeout.

frshwtr:
This is looking good. It looks like the readme needs some content so I'll hold off approving for now.

It would also be good to get DAC to review this before we release this.

components/o-multi-select/demos/src/demo.js
```diff
@@ -0,0 +1,30 @@
...
+			'Fig',
+			'Grape',
+			'Guava',
+			'Huckleberry',
+		],
+	});
+	document.removeEventListener('o.DOMContentLoaded', constructAll);
+};
+
+if (typeof document !== 'undefined') {
```
frshwtr, line 24: Is this needed?

components/o-multi-select/demos/src/demo.mustache
```diff
@@ -0,0 +1,60 @@
...
+	<div class="o-form o-multi-select o-multi-select--core">
+		<span class="o-forms-field">
+			<span class="o-forms-title">
+				<label class="o-forms-title__main" for="multiple"
+					>Multiple select box</label
+				>
+			</span>
+
+			<span class="o-forms-combobox o-forms-combobox--select">
```
frshwtr, line 10: What's `o-forms-combobox`? There's no reference to it in the code base other than this file.

components/o-multi-select/demos/src/demo.mustache
```diff
@@ -0,0 +1,60 @@
+<div class="o-multi-select" data-o-component="o-multi-select">
```
frshwtr, line 1: This is a lot of markup for our users to duplicate in their repo. Increasing the possibility of variations in code which could cause errors. Could we generate more of the experience in Javascript?

components/o-multi-select/src/js/multi-select.js
```diff
@@ -0,0 +1,168 @@
...
+		this.multiSelectEl = multiSelectEl;
+		this._clearCore();
+
+		this.options = Object.assign(
+			{},
+			options || {
+				multiSelectOptions: MultiSelect.getDataAttributes(multiSelectEl),
+			}
+		);
+		if (!this.options.multiSelectOptions.length > 0) {
```
frshwtr, line 27: We should throw if `multiSelectOptions` isn't provided or if `multiSelectOptions` is an array of length 0.

components/o-multi-select/src/js/multi-select.js
```diff
@@ -0,0 +1,168 @@
...
+	 * @param {Object} options - An options object for configuring the component ({multiselectOptions: ['option1', 'option2']})
+	 */
+	constructor(multiSelectEl, options) {
+		this.multiSelectEl = multiSelectEl;
+		this._clearCore();
+
+		this.options = Object.assign(
+			{},
+			options || {
+				multiSelectOptions: MultiSelect.getDataAttributes(multiSelectEl),
```
frshwtr, line 24: `getDataAttributes` should be returning an object of options. In this case we are assigning all our options to `multiSelectOptions`. AFAIK this would mean our multi select options will be on `this.options.multiSelectOptions.multiSelectOptions`. We probably need to look into this a bit more.

frshwtr:
For some reason, the specificity of two rules: `.o-form-field label` and `.o-forms-input--radio-box label`, differ between this branch and main.

components/o-multi-select/src/js/multi-select.js
akomiqaia: do you think this is still relevant after updating the code? looking at `_getCoreOptions()` it seems it will always return an array which is used to assign options to `this.options.multiSelectOptions`

notlee:
Hey @akomiqaia, I made some small changes as I was doing a final review. Please take a read of my commit messages and let me know what you think, happy to revert anything.

I have one other major question before we look to release... when options are selected by an end-user, how do I as a product engineer actually get/use them!? 😱

components/o-forms/src/tsx/Form.tsx
```diff
@@ -144,3 +148,16 @@ export function FormError({errorMessage}) {
 		</span>
 	);
 }
+
+
+export function FormTemplate(props: FormTemplate ) {
+	const {children, ...rest} = props;
+	return (
+		<Form>
```
notlee, line 156: This shouldn't include the `Form` element. The user is likely to want to use this as part of a large `Form` and will want to set `Form` props.

Given that,`FormTemplate` is perhaps not the most descriptive name. This TSX component is more like [our input TSX components](https://github.com/Financial-Times/origami/blob/af76e1f116ef2757f8312ed6e7a8f6ab8ca47169/components/o-forms/src/tsx/TextInput.tsx#L118) (such as `TextInput`), so perhaps following that naming `GenericInput` would be more suitable?

components/o-forms/src/tsx/DateInput.tsx
```diff
@@ -1,5 +1,7 @@
-import {getInputClasses, uniqueId} from '../utils';
+import {getInputClasses} from '../utils';
 import {InputProps, FormError, FormFieldset, TypeFormField} from './Form';
+import {uidBuilder} from "@financial-times/o-utils";
```
notlee, line 3: o-forms needs to add `financial-times/o-utils` to its peer dependencies. I'm surprised this built...

components/o-multi-select/src/js/multi-select.js
```diff
@@ -0,0 +1,222 @@
+import {
+	onComboBoxKeyDown,
+	toggleDropdown,
+	updateCurrentElement,
+	_removeCurrentClass,
+	checkForDuplicates
+} from './utils.js';
+import {updateState} from './state.js';
+import {handleOptionSelect, createOption} from './multi-select-options.js';
+import {uidBuilder} from "@financial-times/o-utils";
```
notlee, line 10: o-multi-select needs to add `financial-times/o-utils` to its peer dependencies. I'm surprised this built...

components/o-multi-select/src/js/multi-select.js
```diff
@@ -0,0 +1,222 @@
...
+class MultiSelect {
+	/**
+	 * Class constructor.
+	 *
+	 * @param {HTMLElement} [multiSelectEl] - The component element in the DOM
+	 * @param {Object} options - An options object for configuring the component ({multiselectOptions: ['option1', 'option2']})
+	 */
+	constructor(multiSelectEl, options) {
+		this.multiSelectEl = multiSelectEl;
+		this.coreWrapper = this._getCoreWrapper();
```
notlee, line 23: Is there any reason this isn't private – do we expect a product developer to call `.coreWrapper`?

components/o-multi-select/src/js/multi-select.js
```diff
@@ -0,0 +1,222 @@
...
+		// data
+		this.idBase = this.comboEl.id;
+		this.totalNumberOfOptions = this.options.multiSelectOptions.length;
+
+		// state
+		this.numberOfSelectedOptions = 0;
+		this.activeIndex = 0;
+		this.open = false;
+
+		this.options.multiSelectOptions.forEach((option, index) => {
```
notlee, line 56: Same here, these should probably be private unless you think there's a good reason for them not to be.
An API to check `numberOfSelectedOptions` is perhaps something that would be useful to provide, but unless we have a usecase in mind we should make it private for now. It's easy to add to the public interface, it's hard to delete from – major release rollout.

notlee (commented):
As discussed we also need to take a look at the Safari/Voiceover experience, which seems... not good

components/o-forms/src/tsx/Form.tsx
akomiqaia: [843029b](https://github.com/Financial-Times/origami/pull/981/commits/843029b14ac32c76f6d954c48187ae4250952500)

components/o-multi-select/src/js/multi-select.js
akomiqaia: [c88ca4a](https://github.com/Financial-Times/origami/pull/981/commits/c88ca4a76151d8543e72d9e66edfc3f62c7d3448)

components/o-forms/src/tsx/DateInput.tsx
akomiqaia: [c88ca4a](https://github.com/Financial-Times/origami/pull/981/commits/c88ca4a76151d8543e72d9e66edfc3f62c7d3448)

components/o-multi-select/src/js/multi-select.js
akomiqaia: [de7f70b](https://github.com/Financial-Times/origami/pull/981/commits/de7f70bc5200fcccd845052b50905b9e3020b75b)

components/o-forms/src/tsx/Form.tsx
```diff
@@ -144,3 +151,16 @@ export function FormError({errorMessage}) {
...
 	);
 }
+
+
+export function GenericInput(props: GenericInputProps ) {
+	const {children, ...rest} = props;
+	return (
+		<>
+			<FormField {...rest} />
+				<div className="o-forms-input">
```
notlee, line 161: Oh! This needs to be _inside_ form field element 
https://registry.origami.ft.com/components/o-forms@9.6.0/readme?brand=core#markup

akomiqaia: in the end we used `span` instead of `div` to have a valid markup

notlee (approved):

PR #986, opened 2023-01-31
Ci 1460 upgrade coral to v 7

juanSanchezAlcala:
Prepare o-comments component for coral version 7.

This is a breaking change it wont work this version of o-comments with previous versions of coral
Instructions on how to migrate are inside the ticket.
[ticket](https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1653?modal=detail&selectedIssue=CI-1460)


juanSanchezAlcala:
> Thanks @juanSanchezAlcala .
> 
> What is the release plan with Coral V7? It will help me with the below question:
> 
> Is this change backward compatible? Looking at our code, it looks like we will pull in whatever version of Coral is hosted at the given url. If we release this component before the new version goes live, will it still work with versions pre v7?

Yes you are right it will be better to make it backwards compatibility. At first i did ,  but while development i thought that it wont be necessary but now, I'm realising that is needed. I'll change it thanks

components/o-comments/README.md
```diff
@@ -232,6 +232,10 @@ State | Major Version | Last Minor Release | Migration guide |
 ╳ deprecated | 2 | 2.5.0 | - |
 ╳ deprecated | 1 | 1.0.10 | - |
 
+## Test on local
+
+In order to test on local you should run from the main folder ```npm run watch -w components/o-comments``` and open the browser on http://local.ft.com:5000
+
```
notlee, line 238: How to test components [is documented](https://github.com/Financial-Times/origami#developing-components) in the root `README.md`. We don't want to include this in every component readme individually as it's repetative and may become outdated

```suggestion
```

components/o-comments/src/scss/coral-talk-iframe/main.scss
```diff
@@ -12,7 +12,7 @@
 
 // stylelint-disable selector-class-pattern, declaration-no-important
 @mixin _oCommentsTalkIframe {
-	.o-comments-coral-talk-container {
+	.o-comments-coral-talk-container, #coral {
```
notlee, line 15: This is causing lint errors. We don't normally use ids for styling but ignore that here by adding `selector-max-id` to the above stylelint-disable comment. Run `npm run lint -w components/o-comments` to confirm it's fixed

notlee:
As this is now backward compatible I don't think we need to relelase a major, or add a migration guide?

juanSanchezAlcala:
> As this is now backward compatible I don't think we need to relelase a major, or add a migration guide?

In my opinion y think we should do as it is using a new version of coral and we don't know 100% if there is a breaking change or not but it could potencially since the code injected is totally different , besides for be able to apply the styles on this version we need to update the coral admin urls getting the styles from the new release. Releasing as a breaking change we don't take the risk of introducing an error on a system that is installing with ^

notlee:
@juanSanchezAlcala won't the backend Coral changes break the current major anyway? A more simple non-major upgrade mitigates risk there

notlee:
Perhaps from a comms perspective you're right though

notlee (approved):

PR #992, opened 2023-02-02
fix: styling changes to deep landscape

phil-hunt-ft:
Makes multiple styling changes to deep landscape:
- Moves the text down to the bottom of the image with padding
- Changes the link text to black in the header tags
- Removes the line between the heading

See: https://financialtimes.atlassian.net/browse/CI-1517

### DIFF
--- a/components/o-topper/src/scss/themes/_deep-landscape.scss
+++ b/components/o-topper/src/scss/themes/_deep-landscape.scss
@@ -90,7 +90,7 @@
 		grid-template-rows: 66vw 1fr;
 
 		.o-topper__content {
-			grid-column: 1;
+			grid-column: 1 span/2;
 			margin: auto;
 		}
 		.o-topper__image{
--- a/components/o-topper/src/scss/themes/_deep-landscape.scss
+++ b/components/o-topper/src/scss/themes/_deep-landscape.scss
@@ -54,7 +54,7 @@
 	@include oGridRespondTo( $until:S) {
 		.o-topper__content {
 			width:100%;
-			padding:20px;
+			padding:20px 20px 0 20px;
 		}
 	}
 
--- a/components/o-topper/src/scss/themes/_deep-landscape.scss
+++ b/components/o-topper/src/scss/themes/_deep-landscape.scss
@@ -12,21 +12,12 @@
 		display:inline-block;
 		grid-row: 1;
 		box-sizing: border-box;
-		align-self: center;
+		align-self: end;
 		background: transparent;
-		position: absolute;
-		left: 10px;
-		bottom: 72px;
+	  	margin: 0 0 48px 12px;
 	}
-  	.o-topper__topic {
-	  	color: black !important;
-	}
-  	.n-myft-follow-button {
-		color: black !important;
-		border: 1px solid black !important;
-		&::before {
-		  	background-image:url(https://www.ft.com/__origami/service/image/v2/images/raw/fticon-v1:plus?format=svg&source=next-article&tint=%23000000%2C%23FFFFFF)
-		}
+	.o-topper--deep-landscape .o-topper__topic {
+	  color: black;
 	}
 	.o-topper__headline {
 		&::after {
@@ -91,7 +82,6 @@
 
 		.o-topper__content {
 			grid-column: 1 span/2;
-			margin: auto;
 		}
 		.o-topper__image{
 			height:66vw;
--- a/components/o-topper/src/scss/themes/_deep-landscape.scss
+++ b/components/o-topper/src/scss/themes/_deep-landscape.scss
@@ -54,7 +54,7 @@
 	@include oGridRespondTo( $until:S) {
 		.o-topper__content {
 			width:100%;
-			padding:20px 20px 0 20px;
+		  	margin: 0 0 36px 12px
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


components/o-topper/src/scss/themes/_deep-landscape.scss
```diff
@@ -14,6 +14,24 @@
 		box-sizing: border-box;
 		align-self: center;
 		background: transparent;
+		position: absolute;
+		left: 10px;
+		bottom: 72px;
+	}
+  	.o-topper__topic {
+	  	color: black !important;
+	}
```
notlee, line 23: For maintainabiliuty [we should avoid](https://developer.mozilla.org/en-US/docs/Web/CSS/important#best_practices) the `!important` declaration.

We actually don't need this anyhow because o-topper can already be themed with classes `o-topper--color-slate`, `o-topper--color-white`, etc. [Theme CSS is output last](https://github.com/Financial-Times/origami/blob/main/components/o-topper/main.scss#L50-L51) so our end CSS looks something like this:

```css
.o-topper--deep-landscape .o-topper__topic {
    color: black; // your new colour
}


.o-topper--color-white .o-topper__topic {
  // existing theme, css includes
    color: black;
}

.o-topper--color-slate .o-topper__topic {
  // existing theme, css includes
    color: white;
}
```

Notice they all have the same [CSS specifcity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity), with 2 classes. Since they have the same specifcity the last one overrides yours. Instead of `!important` we could have increase the specifcity of our selector e.g. `.o-topper--color-slate.o-topper--deep-landscape .o-topper__topic`. 

But we don't have to! Projects which use o-topper can choose to set either `o-topper--color-white` or `o-topper--color-black`. This demo uses `o-topper--color-black` (which assumes a dark topper, and therefore sets white text).

components/o-topper/src/scss/themes/_deep-landscape.scss
```diff
@@ -14,6 +14,24 @@
...
+  	.n-myft-follow-button {
+		color: black !important;
+		border: 1px solid black !important;
+		&::before {
+		  	background-image:url(https://www.ft.com/__origami/service/image/v2/images/raw/fticon-v1:plus?format=svg&source=next-article&tint=%23000000%2C%23FFFFFF)
+		}
+	}
```
notlee, line 30: There styles apply to to another component and don't belong here. It would make it hard to maintain `n-myft-follow-button` if its CSS lived all over the place. Any `n-myft-follow-button` styles should live in the `n-myft-ui` reposity. I may be mistake but at a glance it looks like [follow button themes are defined here](https://github.com/Financial-Times/n-myft-ui/blob/91d0a861c12311349cd679ba78555dd7a9ae84f5/mixins/lozenge/_themes.scss#L1)!

Related there's some [Origami docs here on customisation](https://origami.ft.com/documentation/components/customisation/) which explains why we don't support overridding components with CSS.

notlee (changes requested):

components/o-topper/src/scss/themes/_deep-landscape.scss
```diff
@@ -12,8 +12,17 @@
...
 		grid-row: 1;
 		box-sizing: border-box;
-		align-self: center;
+		align-self: end;
 		background: transparent;
+	  	margin: 0 0 48px 12px;
+	}
+	.o-topper--deep-landscape .o-topper__topic {
+	  color: black;
+	}
```
notlee, line 21:
```suggestion
```

I don't think we need this?

notlee (commented):
Looks good, just 1 peice of redundant css I think we can remove before merging

notlee (approved):

notlee (commented):
@phil-hunt-ft it looks like you've also changed the margin on mobile by mistake?

Screenshot 2023-02-08 at 10 56 57

components/o-topper/src/scss/themes/_deep-landscape.scss
```diff
@@ -43,38 +50,28 @@
...
-			grid-column: 1/span 2;
-			margin: auto auto auto 10%;
 		}
 	}
 
 	@include oGridRespondTo(L) {
 		grid-template-rows: 66vw 1fr;
 
-		.o-topper__content {
-			grid-column: 1;
```
notlee, line 75: Looks like this has adversly affected the layout on larger viewports, and the margin isn't quite right on mobile still. Have you been able to run locally to manually test? There're [instructions in the readme](https://github.com/Financial-Times/origami#running-origami-locally) – o-topper has no storybook demos yet, so you'll need to reference the registry demos
Screenshot 2023-02-08 at 12 39 19
Screenshot 2023-02-08 at 12 39 26

notlee (approved):

components/o-topper/src/scss/themes/_deep-landscape.scss
```diff
@@ -43,37 +50,32 @@
...
 	}
 
 	@include oGridRespondTo(L) {
 		grid-template-rows: 66vw 1fr;
 
 		.o-topper__content {
-			grid-column: 1;
-			margin: auto;
+		  	grid-column: 1;
+		  	margin: 0 0 72px 0;
```
notlee, line 78:
```suggestion
		  	margin: 0 0 72px;
```

The linter is erroring here because the last `0` is superfluous:
https://stylelint.io/user-guide/rules/shorthand-property-no-redundant-values/

I actually disagree with this rule, and think we should look to remove it, because it doesn't really matter and I believe what you wrote is more clear.

PR #999, opened 2023-02-10
feat: Add autoplayWhenInViewport, loop, and showControls properties

boyewale:
Under CI-1519, this introduces the autoplay when in viewport functionality. This uses the intersection observer on the viewport provided when initialising o-video.


umbobabo:
@notlee This is just a draft for a quick prototype to run some tests. You are welcome to give feedback, but you can avoid having an in-depth review at this stage. Our team will iterate on this PR till it's ready. As you mentioned, we may realise that o-video is not the right way we should go for.

umbobabo (commented):
After the mentioned changes is probably good enough to run some local tests with Design and Editorial via a local running instance of next-article

notlee:
Closing as I believe this functionality ended up in n-clip? Thanks @boyewale!

PR #1000, opened 2023-02-10
feat: style topic button for deep landscape [WIP]

phil-hunt-ft:


notlee:
Hey @phil-hunt-ft, could you provide more context as to why you're setting the colour of the topic tag to black? 

As you can see in this screenshot it becomes inaccessible on a dark backgorund:
Screenshot 2023-02-13 at 09 54 10

[o-topper supports themes](https://registry.origami.ft.com/components/o-topper@5.7.4/readme?brand=core#colors) already which will modifier forground colours. For example setting `o-topper--color-white` on o-topper's root element will set a white background (behind the image) and set contrasting black text:

Screenshot 2023-02-13 at 09 56 49

notlee (changes requested):
see: https://github.com/Financial-Times/origami/pull/1000#issuecomment-1427649998

PR #1006, opened 2023-02-15
docs: add oxford comma

notlee:
Testing a Jira integration. Let's add an oxford comma. Both me and Thérèse Coffey are a fan, but otherwise we probably have little in common


KamranMuniree (approved):
LGTM

PR #929, opened 2022-12-15
Ci 1409 deep landscape full bleed left

juanSanchezAlcala:
Add new topper Deep Landscape.

[Design](https://www.figma.com/file/oT8c4GLbGgtg8EsyndZgGe/Design-File-(for-product)?node-id=1506%3A2471&t=fQMlf1j0hN7As4Kk-0) 

[Ticket](https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1316?modal=detail&selectedIssue=CI-1409)

PR #1008, opened 2023-02-16
feat: Background Box support in deep landscape

umbobabo:
We are introducing a Background Box variation for the deep landscape layout in toppers.
The  Background Box option is meant to improve readability when the text shadow is not good enough depending on the used background image chosen by Editorial. More information in the [ticket](https://financialtimes.atlassian.net/browse/CI-1543)

Desktop/Tablet (White and black permutations) 

image

image

Mobile (White and black permutations) 

image
 
image

In this PR:

* Add new background box modifier
* Fix the credit not appearing in the Deep Landscape demo
* Extending the existing demo to cover more permutation of the components
* Reducing to 16px the margin-bottom for all the headlines (Design request)

### DIFF
--- a/components/o-topper/src/scss/themes/_deep-landscape.scss
+++ b/components/o-topper/src/scss/themes/_deep-landscape.scss
@@ -32,11 +32,18 @@
 					text-shadow: 2px 3px 4px rgba(18,18,18,0.4);
 				}
 			}
-			.o-topper__content--box-shadow {
-					padding: oSpacingByName('s4');
-					margin-bottom: oSpacingByName('s6');
-					background-color: if($foreground == white, rgb(18,18,18,0.4), rgb(255,255,255,0.4));
-			}
+		}
+	}
+
+	$background-box-colors: (
+		'light',
+		'dark'
+	);
+	@each $background-box-color in $background-box-colors {
+		.o-topper__content.o-topper__content--background-box-#{$background-box-color} {
+			padding: oSpacingByName('s4');
+			margin-bottom: oSpacingByName('s6');
+			background-color: if($background-box-color == 'dark', rgb(18,18,18,0.4), rgb(255,255,255,0.4));
 		}
 	}
 
--- a/components/o-topper/src/scss/themes/_deep-landscape.scss
+++ b/components/o-topper/src/scss/themes/_deep-landscape.scss
@@ -32,6 +32,11 @@
 					text-shadow: 2px 3px 4px rgba(18,18,18,0.4);
 				}
 			}
+			.o-topper__content--box-shadow {
+					padding: oSpacingByName('s4');
+					margin-bottom: oSpacingByName('s6');
+					background-color: if($foreground == white, rgb(18,18,18,0.4), rgb(255,255,255,0.4));
+			}
 		}
 	}
 
--- a/components/o-topper/src/scss/themes/_deep-landscape.scss
+++ b/components/o-topper/src/scss/themes/_deep-landscape.scss
@@ -32,18 +32,11 @@
 					text-shadow: 2px 3px 4px rgba(18,18,18,0.4);
 				}
 			}
-		}
-	}
-
-	$background-box-colors: (
-		'light',
-		'dark'
-	);
-	@each $background-box-color in $background-box-colors {
-		.o-topper__content.o-topper__content--background-box-#{$background-box-color} {
-			padding: oSpacingByName('s4');
-			margin-bottom: oSpacingByName('s6');
-			background-color: if($background-box-color == 'dark', rgb(18,18,18,0.4), rgb(255,255,255,0.4));
+			.o-topper__content--background-box {
+				padding: oSpacingByName('s4');
+				margin-bottom: oSpacingByName('s6');
+				background-color: if($foreground == white, rgb(18,18,18,0.4), rgb(255,255,255,0.4));
+			}
 		}
 	}
 
--- a/components/o-topper/README.md
+++ b/components/o-topper/README.md
@@ -88,6 +88,13 @@ These colors affect the background of the `.o-topper__background` and `.o-topper
 .o-topper--color-matisse
 ```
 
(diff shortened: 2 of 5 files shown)


notlee (approved):
@umbobabo Approving the CSS changes 👍 However do you think we could instead release `o-topper__content--text-shadow`? 

Picking between nothing, `text-shadow`, and `box-shadow` feels like an onerous decision to place on editorial. They'd need to pick from 3 options keeping in mind contrast requirements at multiple viewport sizes, and for screens that provide less good contrast to out work machines. 

Reducing to 2 options (nothing or the box shadow style) is an easier and less nuanced decision to have to make.

umbobabo:
> Picking between nothing, `text-shadow`, and `box-shadow` feels like an onerous decision to place on editorial. They'd need to pick from 3 options keeping in mind contrast requirements at multiple viewport sizes, and screens that provide less good contrast than our work machines.
> 
> Reducing to 2 options (nothing or the box shadow style) is an easier and less nuanced decision to have to make.

Thank you for the feedback.
I appreciate what you are saying, but at the same time, we see a push in the composition of very different pages with different looks and feels, like HTSI that are leading towards integrating more capabilities in Spark UI.
The requested changes went through the review process in Spark with Editorial, and they have been discussed and approved, so I would rather not challenge them at this stage.

If you have concerns about the direction this is taking (and that is understandable) let's have a chat and see if we should be more careful about accepting these kinds of changes.

I pushed an additional change that just enforces the bottom margin on mobile too for this new modifier via higher specific of the selector.

image

juanSanchezAlcala (commented):
Shouldn't we set the color of the text based on the background-box-colors in case it has ? Or it is posible to have backround dark and text black ? If not i would add also 
`color: if($background-box-color == 'dark', rgb(255,255,255), rgb(18,18,18,0.4));`

umbobabo:
> Shouldn't we set the color of the text based on the background-box-colors in case it has ? Or it is posible to have backround dark and text black ? If not i would add also `color: if($background-box-color == 'dark', rgb(255,255,255), rgb(18,18,18,0.4));`

The text color logic is quite [tangled with the background colour](https://github.com/Financial-Times/origami/blob/3eb60451d48f5f323f9b97c228f71e0b4b3c7161/components/o-topper/src/scss/_mixins.scss#L254), to change this logic only for deep landscape would mean change quite a lot of stuff given the way [all the topper are created from the same mixing](https://github.com/Financial-Times/origami/blob/3eb60451d48f5f323f9b97c228f71e0b4b3c7161/components/o-topper/main.scss#L50) that split colours to theme part.

I can have a better look, but my solution was to [force the background colour when there is a background box](https://github.com/Financial-Times/n-map-content-to-topper/pull/41).

juanSanchezAlcala (approved):

components/o-topper/src/scss/themes/_deep-landscape.scss
```diff
@@ -32,11 +32,18 @@
...
-			}
+		}
+	}
+
+	$background-box-colors: (
+		'light',
+		'dark'
+	);
+	@each $background-box-color in $background-box-colors {
+		.o-topper__content.o-topper__content--background-box-#{$background-box-color} {
```
notlee, line 43: The demo is broken as this new class hasn't been applied. Although the commit mentions "add margin on mobile resolution increasing specificity" but I'm not seeing any difference from before, there didn't seem to be a specificity issue.

Screenshot 2023-03-03 at 08 40 38

The new approach makes the markup / templating more complex and introduces scope for errors as the right `background-box` class has to be set for the right colour. I prefer your previous approach which aligns with how `text-shadow` is applied. Could we stick with your previous approach unless there's a good reason for the new classes I'm not seeing?

notlee (changes requested):
@umbobabo I would like us to ask Editorial if they intend to use both `text-shadow` and `background-box` styles, or if `background-box` means we can deprecate `text-shadow`.

Originally Deep Portrait had neither style. [Origami suggested design changes to improve readability](https://github.com/Financial-Times/origami/pull/929#pullrequestreview-1222438509). We worked with Jazmin to propose 2 options (text-shadow and background-box). DAC recommended background-box as did Origami, but editorial chose to go with text-shadow. Now box-shadow is being added. I'd like to know if they have changed their mind (as text-shadow isn't very effective) or if they actually would like to use both. 

I'm happy either way but want to make sure we're removing options which aren't intended to be used and associated tech debt.

umbobabo:
> @umbobabo I'm totally not opposed to this new style (it's better). I would like us to ask Editorial if they intend to use both `text-shadow` and `background-box` styles, or if `background-box` means we can deprecate `text-shadow`.
> 
> Originally Deep Portrait had neither style. [Origami suggested design changes to improve readability](https://github.com/Financial-Times/origami/pull/929#pullrequestreview-1222438509). We worked with Jazmin to propose 2 options (text-shadow and background-box). DAC recommended background-box as did Origami, but editorial chose to go with text-shadow. Now box-shadow is being added. I'd like to know if they have changed their mind (as text-shadow isn't very effective) or if they actually would like to use both.
> 
> I'm happy either way but want to make sure we're removing options which aren't intended to be used, along with tech debt associated with maintaining many permutations of a design in both code and design tooling.

It totally makes sense to me. I have raised this discussion in StoryTelling Stand Up this morning and I will keep you in the loop.

components/o-topper/src/scss/themes/_deep-landscape.scss
umbobabo: I decided to revert to the previous approach as you mentioned, which also fixed the demo, let me know what you think.

This is how things look now after a rebase on the latest version of Origami

image

image

image

image

notlee:
Cool, I've approved that @umbobabo from a code perspective. I've just updated [the Jira ticket](https://financialtimes.atlassian.net/browse/CI-1543?focusedCommentId=679680) with the context I shared here too regarding potentially deprecating the text shadow

PR #1009, opened 2023-02-17
feat: ft-concept-button, add inverse monochrome variant

notlee:
This will be used first within o-topper on the article page, on white backgrounds. Typically Origami components use "inverse" themes to indicate a white foreground, for use on a dark background. However in this case ft-concept-button has an "inverse" theme already which is intended for use on claret backgrounds, and uses a claret text colour when clicked. We want an inverse theme that is also monochrome...

In storybook, when not pressed and pressed:
Screenshot 2023-02-17 at 09 38 04

Screenshot 2023-02-17 at 09 37 03

Design example in context, not pressed:
Screenshot 2023-02-16 at 10 15 39

### DIFF
--- a/components/ft-concept-button/README.md
+++ b/components/ft-concept-button/README.md
@@ -105,7 +105,7 @@ export interface ConceptButtonProps {
 	// descriptive label for button
 	ariaLabel?: string;
 	// button theme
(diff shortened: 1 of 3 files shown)


akomiqaia (approved):

umbobabo (approved):
Thank you both!

PR #1010, opened 2023-02-17
feat: add option to set an additional class on shadow dom for o-comments

juanSanchezAlcala:
**Why ?** 
On the app the rem font size is different than in ft.com so we need to set a variant for the app to keep the same style.
**What ?** 
We set a new data-attribute called  data-o-comments-add-class which adds an additional class to coral component. On the app  we set the value data-o-comments-add-class="o-comments-app" and then it will take the styles that we have created for this class in this PR.
[ticket](https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1653?modal=detail&selectedIssue=CI-1460)

### DIFF
--- a/components/o-comments/src/js/stream.js
+++ b/components/o-comments/src/js/stream.js
@@ -74,7 +74,10 @@ class Stream {
 					: 'https://ft.coral.coralproject.net';
 				const scriptElement = document.createElement('script');
 				scriptElement.src = `${rootUrl}/assets/js/embed.js?${cacheBuster}`;
-
+				const containerClassName = ['o-comments-coral-talk-container'];
+				if(this.options.addClass){
+					containerClassName.push(this.options.addClass);
+				}
 				scriptElement.onload = () => {
 					this.embed = Coral.createStreamEmbed(
 						{
@@ -84,7 +87,7 @@ class Stream {
 							rootURL: rootUrl,
 							autoRender: true,
 							bodyClassName: 'o-comments-coral-talk-container',
-							containerClassName: 'o-comments-coral-talk-container',
+							containerClassName,
 							events: (events) => {
 								events.onAny((name, data) => {
 									this.publishEvent({name, data});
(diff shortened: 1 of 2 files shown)


notlee:
Hey @juanSanchezAlcala, couple of questions.

1. Can you think of a more generic name so this class could be used outside the app in the future? I think ft.com are the outliers by customising the root font-size value.
2. How will the new class be applied within the app? I believe `#coral` and `.o-comments-coral-talk-container` are set within the Coral iframe?

juanSanchezAlcala:
> Hey @juanSanchezAlcala, couple of questions.
> 
> 1. Can you think of a more generic name so this class could be used outside the app in the future? I think ft.com are the outliers by customising the root font-size value.
> 2. How will the new class be applied within the app? I believe `#coral` and `.o-comments-coral-talk-container` are set within the Coral iframe?

1. How about "o-comments-for-rem-1"
2.#coral match the new version of Coral 7,   .o-comments-coral-talk-container match coral version 6. #coral.o-comments-for-rem-1 would match the new version and add the overrides to Coral 7 to be ok. #coral is not set on the iframe only exists on version 7 which is with the shadow DOM

notlee:
@juanSanchezAlcala 

1. Good idea naming on font size, maybe `o-comments-root-font-size` would be more descriptive as 1rem could apply to lots of things.
2. If `#coral` exists in the shadow DOM I'm not sure I understand how you can set `#coral.o-comments-[the new class name]` only in the app?

juanSanchezAlcala:
> @juanSanchezAlcala
> 
> 1. Good idea naming on font size, maybe `o-comments-root-font-size` would be more descriptive as 1rem could apply to lots of things.
> 2. If `#coral` exists in the shadow DOM I'm not sure I understand how you can set `#coral.o-comments-[the new class name]` only in the app?

ok i think we should add the 1 at the end of the class to clarify that is for root font size 1rem because by default is for 0.8 which is set on ft.com , so it would be 'o-comments-root-font-size-1' , it is ok? .
About how we are going to set that class is throught javascript with the property "containerClassName" where we can pass an array of clasess to append to the main container of the coral component, we take that additional classes to add from data attribute data-o-comments-add-class.

notlee:
>with the property "containerClassName"

@juanSanchezAlcala Sorry Juan, I totally didn't spot the JavaScript changes!

notlee (approved):

PR #1018, opened 2023-02-21
feat: o-forms, allow submit events to be intercepted

notlee:
By default o-forms listens to submit events. It validates a form and, if it is valid, will then call submit on the form element. This means projects which use o-forms cannot prevent a valid form from submitting.

This may be desirable to add extra steps prior to submission. For example to load a captcha (~apparently this should happen on submit?~ which we want to run when the form is filled out correctly) or to display a final confirmation.

To achieve this option:

This PR introduces the `oForms.submit` event. This is fired by o-forms when its submit event handler has finished form validation, just before submitting the form. A user may listen for this event and prevent propagation to stop the form from submitting. As this would rely on the project's event listener running first, however, the there is now also an option (`preventSubmit`) to ensure the form is not submitted by `o-forms`.

E.g:
```js
import oForms from '@financial-times/o-forms';

// Initialise a form. Note `oForms.init()` could also be used.
const formElement = document.forms[0];
new oForms(formElement, {
	preventSubmit: true
});

// Listen to the form submit event to handle submission manually.
formElement.addEventListener('oForms.submit', (e) => {
	console.log(`A user would like to submit a form, it is ${e.detail.valid ? 'valid' : 'invalid'}.`);
	if(e.detail.valid) {
		e.detail.instance.form.submit();
	}
});
```

### DIFF
--- a/components/o-forms/README.md
+++ b/components/o-forms/README.md
@@ -788,19 +788,42 @@ The main `o-forms` JavaScript has been written to identify and run on a `<form>`
 </form>
 ```
 
-By default, `o-forms` is initialised _without_ native browser validation, and with an error summary for invalid elements when the form is submitted. In order to use the default behaviour, you'll need to do the following:
+By default, `o-forms` is initialised with validation logic. It will generate an error summary for invalid elements when the form is submitted:
 ```js
 import oForms from '@financial-times/o-forms';
 oForms.init()
 ```
-The default behaviour can be changed by configuring the options object:
+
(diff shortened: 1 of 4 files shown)


akomiqaia (approved):
LGTM. Jus one thing is that i thought we wanted to deprecate proclaim and use Jest assertions for tests?

notlee:
@akomiqaia Good shout, but I'd rather use one assertion library consistently in a file and keep refactoring to another in a separate PR. Do you mean [Sinon assertions](https://github.com/Financial-Times/origami/blob/893ede80660ca89dc52a7b407efb9e94172fafe8/components/o-video/test/playlist.test.js#L62)? I'm not aware we used Jest to test any components, but may have forgotten or missed something

JuanTincho (approved):
Thank you! 🙌

PR #1022, opened 2023-02-22
feat: add scrollContainer option on o-comments

juanSanchezAlcala:
Why?
We have an issue on the app that provoques bad rendering on comments when they are more than 12 . With this option we can fix it and stablished what will be the parent scroll container of the comments which in some cases like in the app now or if they would be in a modal element ,would be needed to work properly 
What ?
We add an option data-o-comments-scroll-container to de component that contains a selector to get the element which is going to be the custom scroll container

### DIFF
--- a/components/o-comments/src/js/stream.js
+++ b/components/o-comments/src/js/stream.js
@@ -78,6 +78,7 @@ class Stream {
 				if(this.options.addClass){
 					containerClassName.push(this.options.addClass);
 				}
+				const customScrollContainer = document.querySelector(this.options.scrollContainer);
 				scriptElement.onload = () => {
 					this.embed = Coral.createStreamEmbed(
 						{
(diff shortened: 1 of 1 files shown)


notlee (approved):

components/o-comments/src/js/stream.js
```diff
@@ -78,6 +78,7 @@ class Stream {
 				if(this.options.addClass){
 					containerClassName.push(this.options.addClass);
 				}
+				const customScrollContainer = document.querySelector(this.options.scrollContainer);
```
umbobabo, line 81: If `this.options.scrollContainer` is undefined, `customScrollContainer` will be _null_.

Did you test this case?

We should be fine [given Coral will fallback to undefined](https://github.com/coralproject/talk/pull/4050/files#diff-69ce3c7bd00078e329e4827a8aa3d391a0d31f7981f05b87d2cb28f56e022faeR80)

umbobabo (approved):

components/o-comments/src/js/stream.js
juanSanchezAlcala: yes I've done it and remains it as if you don't set that property

PR #1024, opened 2023-02-23
Add any icon support to o buttons TSX template

akomiqaia:
This PR does not change the way o-buttons work.  It adds support for passing any string for the `icon` prop while still keeping the default icon sets as suggestions in IDE. This should improve TS users' experience with icons and also stop showing errors when custom icons are used.

There are a lot of eslint changes in the PR but the line that actually changed is line 23 in `button.tsx`

### DIFF
--- a/components/o-buttons/src/tsx/button.tsx
+++ b/components/o-buttons/src/tsx/button.tsx
@@ -20,7 +20,7 @@ export interface ButtonProps {
 		| "refresh"
 		| "cross"
 		| ""
-		| (string & Record<never, never>);
+		| (string & Record<never, never>); // Support IDE autocomplete whilst allowing any string https://github.com/microsoft/TypeScript/issues/29729#issuecomment-1331857805
 	iconOnly?: boolean;
 	attributes?: {
 		[attribute: string]: string | boolean;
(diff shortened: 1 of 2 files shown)


components/o-buttons/src/tsx/button.tsx
```diff
@@ -1,25 +1,26 @@
...
+		| "arrow-up"
+		| "grid"
+		| "list"
+		| "edit"
+		| "download"
+		| "search"
+		| "refresh"
+		| "cross"
+		| ""
+		| (string & Record<never, never>);
```
notlee, line 23: This feels a little esoteric, it would be a great addition to have a little comment to explain what this is for / how it works, and or a link to [a reference](https://github.com/microsoft/TypeScript/issues/29729#issuecomment-1331857805).

notlee (approved):
👍

components/o-buttons/src/tsx/button.tsx
notlee, line 23:
```suggestion
		| (string & Record<never, never>); // Support IDE autocomplete whilst allowing any string https://github.com/microsoft/TypeScript/issues/29729#issuecomment-1331857805
```

j-mes:
Looks great @akomiqaia 😄

PR #1028, opened 2023-02-28
feat: introduce o-table typescript templates

frshwtr:


components/o-table/src/tsx/table.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+	className?: string;
+	sortable?: boolean;
+}
+
+export const Table = ({children, compact, horizontalLines = true, className, sortable, ...args}: TableProps) => {
+	const classNames = ['o-table', className && className.split(' ')];
+	const attributes: Record<string, any> = {};
+
+	if(!sortable) {
```
frshwtr, line 14: Interesting behaviour here. When toggling the control for this in Storybook, the sort button will remain. It has something to do with `addSortButtons` adding [this clas](https://github.com/Financial-Times/origami/blob/main/components/o-table/src/js/Tables/BaseTable.js#L485) but the DOM not resetting this upon re-rendering.

components/o-table/src/tsx/table.tsx
```diff
@@ -0,0 +1,72 @@
...
+		classNames.push('o-table--row-stripes');
+	}
+	return <table className={classNames.join(' ')}
+				  data-o-component="o-table" {...attributes} {...args}>{children}</table>;
+};
+
+interface ResponsiveTableProps extends TableProps {
+	behaviour: 'scroll' | 'overflow' | 'flat';
+	expanded?: boolean;
+	minimumRows?: number;
```
notlee, line 41: Only the `overflow` response table may be expandable. When `scroll` or `flat` behaviour is set the `expanded` and `minimumRows` properties have no effect. This isn't self-documenting and could be confusing for users who haven't read through our documentation carefully. Have you considered an interface for each type of responsive table?

components/o-table/src/tsx/components.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const TableFoot = args => {
+	return <tfoot {...args} />
+}
+export const TableRow = args => {
+	return <tr {...args}/>
+}
+
+interface TableHeadingProps {
+	children?: string | JSX.Element | JSX.Element[];
+	className?: string;
```
notlee, line 16: Is there a reason to support `className` as part of the public interface? 

We should make an effort to encourage teams not to set [unsupported CSS overrides](https://origami.ft.com/documentation/components/customisation/#avoid-css-overrides), which I think custom class names implies is allowed. 

There are times we need to provide a mechanism for custom classes though, for example where an Origami Sass API for customisation legitimately outputs a custom class name.

components/o-table/src/tsx/components.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export const TableHeading = (args: TableHeadingProps) => {
+	return <th scope="col" role="columnheader" {...args}/>
```
notlee, line 20: o-table supports custom sort formats, we need to provide a way to specify that with `data-o-table-data-type`
https://registry.origami.ft.com/components/o-table@9.2.7/readme?brand=core#custom-sort-imperative

components/o-table/src/tsx/table.tsx
```diff
@@ -0,0 +1,72 @@
+interface TableProps {
+	children: string | JSX.Element | JSX.Element[];
+	compact?: boolean;
+	horizontalLines?: boolean;
+	stripes?: boolean;
+	className?: string;
```
notlee, line 6: Same as above on custom `className`'s

notlee (commented):
Just had a quick glance through to give some ~early~ less late feedback ahead of a proper review :)

components/o-table/src/tsx/components.tsx
```diff
@@ -0,0 +1,92 @@
+import {Select} from "@financial-times/o-forms/src/tsx/o-forms";
```
notlee, line 1: @akomiqaia @frshwtr Ah, this relates to our o-multi-select conversation! 😄 Do we want to include o-forms tsx as part of o-table tsx, or do we prefer to leave composition of o-forms and o-table tsx to the user (and the storybook story for demo purposes).

*Including o-forms tsx in o-table tsx*

Pros:
- We move toward forcing o-table users to use o-forms. Origami is more opinioned. An opinioned Origami is easier to use and more consistent.
- It's (arguably) simpler to use o-table because the user doesn't have to think about how they would use with o-forms and there are fewer params to customise the select element  o-forms. (Though in this case though the user still has to manage setting, for example, the same id across `SelectFilter` and `Table` anyway).

Cons:
- As a user I still need to include o-forms CSS separately, kinda confusing?
- If I want to set slightly different form field options for my filter I can't use o-table's TSX, and may be tempted to just write markup instead.
- o-table would become directly dependent on o-forms. It should be listed in the package.json as a peer dependency. If we release a major of o-forms, we would have to release a major of o-table, which uses the upgraded version of o-forms. Makes a rollout plan more difficult and impacts more users.

*Rely on the user to compose o-forms tsx with o-table tsx*

Pros:
- Fits with how o-table and other components currently work. Which is easier to maintain and use – when you know that.
- No direct dependency between o-table and o-forms, users can upgrade both separately and a major release to one doesn't impact the other.

 Cons:
 - We have to document how to compose the templates of both components for Origami users to follow.
 
 
Any other thoughts? I feel like we should stick with keeping the composition of component tsx with the Origami user in this case

components/o-table/stories/tableWithFilter.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+							  filterColumn='2'>
+					<option value="">All</option>
+					<option value="Juicy">Juicy</option>
+					<option value="Smelly">Smelly</option>
+					<option value="Chewy">Chewy</option>
+					<option value="Sweet">Sweet</option>
+					<option value="Crunchy">Crunchy</option>
+				</SelectFilter>
+			</div>
+			<div className='o-table-container'>
```
notlee, line 33: Seeing o-table markup in the Storybook tells us it's not encapsulated in the TSX template. Do you think we could refactor this so `o-table-container` is part of the TSX template?

components/o-table/stories/table.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+	useEffect(() => {
+		javascript.init();
+	});
+
+	return <Table {...args}>
+		<TableCaption>
+			<h2 className="o-typography-heading-level-2">Origami Team's Favourite Fruit</h2>
+		</TableCaption>
+		{baseTableContents}
+		<TableFoot>
```
akomiqaia, line 38: It seems like the footnote is limited to the width of the first column. In [registry demos](https://www.ft.com/__origami/service/build/v3/demo?component=o-table%409.2.7&demo=captions&system_code=origami-repo-data&brand=core)  footnote does not get wrapped if the first column is small.
Screenshot 2023-03-22 at 11 19 11

 Also on SB our table stretches and takes up all the space(maybe desired behaviour) while in the registry demo, it seems like the table take only the space that it needs. Again I might be wrong about it's actual behaviour :D

components/o-table/stories/table.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+	useEffect(() => {
+		javascript.init();
+	});
+
+	return <Table {...args}>
+		{baseTableContents}
+	</Table>;
+}
+export const BasicTable = Template.bind({});
+
```
akomiqaia, line 27: what do you think about adding default args for controls? It might be a better experience for users if it's already set to something.

akomiqaia, line 27: Screenshot 2023-03-22 at 11 19 30

when I set sortable to true the sort icons appear but when I set it back to false they remain in the UI (and functionality as well).

components/o-table/stories/baseTableContents.tsx
```diff
@@ -0,0 +1,61 @@
...
+		<TableBody>
+			<TableRow>
+				<TableData>Dragonfruit</TableData>
+				<TableData>Stenocereus</TableData>
+				<TableData>Juicy</TableData>
+				<TableData dataType='numeric'>3</TableData>
+				<TableData dataType='numeric'>2.72</TableData>
+			</TableRow>
+			<TableRow>
+				<TableData>Durian</TableData>
```
akomiqaia, line 33: Not important at all abut but [some of our demos have](https://registry.origami.ft.com/components/o-table@9.2.7?brand=core#demo-sorting-disabled) links in rows and it might be good idea to show that in our demos as well? but I don't think this is a blocker

components/o-table/src/tsx/components.tsx
frshwtr: I think going with composition is the right choice here. It complements our general Origami decision around "users should use their own components where design/behaviour of a given element differs from Origami's implementation, and should not overide css" (paraphrased).

components/o-table/stories/table.stories.tsx
frshwtr: Aah yes I noticed this as well. It's something to do with the classes not resetting to an initial state after JS has attached classes to an element.

I was a little stumped on how to fix this, might be good for one of @akomiqaia or @notlee to take a look with me.

notlee (approved):

PR #1029, opened 2023-02-28
fix: remove unnecessary changes moved into another component

umbobabo:
I'm creating a wrapper component to handle all the observations of the different videos on the page so to make the changes in o-video a simple as possible.


notlee:
Closing as I believe we created n-clip instead

PR #1031, opened 2023-03-03
Major o-topper release: remove JS including `n-map-content-to-topper` 

notlee:
`o-topper` no longer includes JavaScript to select the correct topper for JSON-formatted FT articles and flags. This helper was deeply tied to the FT.com content store, and included hardcoded UUIDs and business logic beyond the scope of Origami. Origami components focus on providing reusable user interfaces – without business logic assumptions which could limit their use outside specific groups or use-cases.

Not to mention, `n-map-content-to-topper` does not set its version number in `package.json` causing Origami contributors to have to `npm i` regularly for some reason – e.g. when checking out a new branch. Super annoying and time consuming.

**Review but do not merge yet:** Let's wait for [this topper change to be released](https://github.com/Financial-Times/origami/pull/1008), so we don't block the feature release behind a major release which (may) involve co-ordination between multiple teams.

### DIFF
--- a/components/o-topper/MIGRATION.md
+++ b/components/o-topper/MIGRATION.md
@@ -1,5 +1,15 @@
 # Migration guide
+## Migrating from v5 to v6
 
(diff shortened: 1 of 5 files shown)


frshwtr:
LGTM

akomiqaia (approved):

PR #1035, opened 2023-03-07
Ci 1586 rule above deep portrait captions

juanSanchezAlcala:
We need to:
- remove the line or border above captions on deep-portrait toppers since all other toppers doesn't have it.
- remove opacity from standfirst on deep-landscape
Also we leverage to change the filename portrait to deep-portrait so we can identify better that is part of a deep topper.
[ticket](https://financialtimes.atlassian.net/browse/CI-1586)
[ticket](https://financialtimes.atlassian.net/browse/CI-1590)
 
BEFORE
image
image

AFTER
image
image


components/o-topper/src/scss/themes/_deep-portrait.scss
```diff
@@ -91,7 +91,6 @@
 		.o-topper__image-caption {
 			margin-bottom: 0;
 			margin-top: 10px;
```
umbobabo, line 93: Thought: This was perhaps added to separate the line from the image. We should probably get rid of this one too, what do you think?

juanSanchezAlcala: I shouldn't do now since editorial or designer are not complaining about this margin. I would move this question to the designer and in case she see it right with editorial we can do this.

umbobabo (approved):

notlee (approved):

PR #1040, opened 2023-03-09
O multi select tweaks

frshwtr:


components/o-multi-select/src/js/multi-select.js
```diff
@@ -97,14 +99,63 @@ class MultiSelect {
...
+            id="o-multi-select-selected"
+    ></ul>
+    <div class="o-multi-select__combobox-wrapper">
+        <div
+                class="o-multi-select__combobox"
+                id="${selectId}"
+                name=${selectName}
+                role="combobox"
+                aria-activedescendant=""
+                aria-labelledby="o-multi-select-label o-multi-select-selected"
```
notlee, line 120: When enhanced we need to set the `aria-labelledby` attribute with the id of the `label` provided by the user and the id of the list of selected options. 

1. We don't know the id of the label get by the user, or even if they have set a label. Could we use something like `this.coreWrapper.labels` to get hold of that id, or error if no label has been setup? This enforces some level of accessibility. https://html.spec.whatwg.org/multipage/form-elements.html#the-select-element 
2. The `o-multi-select-selected` references are entirely controlled by the component. To support multiple multi-select components, do we generate a random id?

frshwtr:
The readout of selectable options seems odd in Voiceover, it seems to exist without the changes we made 
image

The list gets read out with each individual character instead of words.

frshwtr:
> The readout of selectable options seems odd in Voiceover, it seems to exist without the changes we made image
> 
> The list gets read out with each individual character instead of words.

This is now fixed. There was an issue with how labels were linked

akomiqaia (approved):

PR #1062, opened 2023-03-30
o-tracking/add trace to custom events

maxcbc:
# o-tracking/add trace to custom events
## What
This PR adds context to custom events. Including:
- context.url
- context.domPathTokens
- support for `data-trackable-context-*` attributes

## Why
Currently custom events are the only events that can be emitted on elements which are missing this contextual information.
This means, if an engineer wishes to have the same contextual information as when an event is viewed/clicked then they must effectively re-implement get-trace in their own app.

## Impact
This change will mean existing usage of custom events:
- will have extra data added to its context object, however no properties will be overridden, with the exception of `url` and `domPath`  (mirroring behaviour on other event types).
- more data will be sent into spoor.

### DIFF
--- a/libraries/o-tracking/src/javascript/events/click.js
+++ b/libraries/o-tracking/src/javascript/events/click.js
@@ -1,9 +1,9 @@
 import Delegate from 'ftdomdelegate';
 import core from '../core.js';
-import {sanitise, assignIfUndefined, merge } from '../utils.js';
+import {assignIfUndefined, merge, sanitise} from '../utils.js';
 import {get as getSetting} from '../core/settings.js';
-import {getTrace} from '../../libs/get-trace.js';
-import { Queue } from '../core/queue.js';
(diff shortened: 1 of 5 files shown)


FranciscoRuiz-FT (approved):
Hi @maxcbc, sorry for the delay on checking this and thank you for the added feature and the nice refactorings. It all looks great :+1:

notlee:
Hey @FranciscoRuiz-FT. Looks like Max has left the FT. Would you still like to release this?

notlee:
>Currently in the Data Platform we are still trying to hire someone who can support the O-tracking. We currently don't have any dedicated person to support this project. So for now we are not planning any code changes in the O-tracking until something is urgent.

https://github.com/Financial-Times/origami/pull/1063

Given the above, and that Max has left the FT, I'm going to close this PR for now.

PR #1063, opened 2023-03-30
feat: [o-tracking] add mutation observer to view tracking

maxcbc:
# feat: [o-tracking] add mutation observer to view tracking

TLDR; adds view events for matching elements added to DOM after `oTracking.view.init()` method called.

## What
This PR adds a mutation observer to the view tracking initialisation method (where supported), which adds/removes intersection observation of elements matching the given selector as they are added/removed from the DOM.

## Why
In most places on the FT, the site is largely static after the dom loads. 
However in some places (ft.com/myft) the DOM is updated after JS initialisation. This means that view tracking would need to be re-initialised on every DOM change, potentially causing duplicate view events.

### DIFF
--- a/libraries/o-tracking/src/javascript/events/component-view.js
+++ b/libraries/o-tracking/src/javascript/events/component-view.js
@@ -53,15 +53,30 @@ const init = (opts = {}) => {
 	}
 
 	const selector = opts.selector || '[data-o-tracking-view]';
-	const elementsToTrack = [...document.querySelectorAll(selector)];
 
-	if (!elementsToTrack.length) {
-		return;
+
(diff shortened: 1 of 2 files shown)


notlee:
Hey @FranciscoRuiz-FT. Looks like Max has left the FT, do you think this is something we should review and release?

hamidrazaft:
> Hey @FranciscoRuiz-FT. Looks like Max has left the FT, do you think this is something we should review and release?

> Hi @notlee : Currently in the Data Platform we are still trying to hire someone who can support the O-tracking. We currently don't have any dedicated person to support this project. So for now we are not planning any code changes in the O-tracking until something is urgent.

notlee:
Given the above, I'll close this PR for now

PR #1076, opened 2023-04-03
fix: o-forms, remove valid state [OR-63]

notlee:
We agreed to make this change ages ago. Let's confirm with the design team again and wait for this change to make it into our figma ui kit before merging. [Conversation on Slack](https://financialtimes.slack.com/archives/C01481FKWA2/p1680536391077469).

Figma branch: https://www.figma.com/file/MyHQ1qdwYyek5IBdhEEaND/branch/jSgH0Zz7WqPsB3E3XpXQty/FT-UI-Library?t=VAoTu9WeEBzX7ls6-1
Jira: https://financialtimes.atlassian.net/browse/OR-63

We decided to remove the valid state because:
- It's not great at what it does, there is no indication for screen reader users and it's a poor indication for some people with colour blindness.
- It is only applied when there has been a previous error, so some valid things don't show as valid – confusing!
- It does not even guarantee it is correct, we have server side validation to go yet – misleading!
- Also gives the input low contrast against paper.


akomiqaia (changes requested):
LGTM but we need to update test as well.

notlee:
This was approved by CPs design manager before he left, and given a tentative 👍 by Keiran

notlee:
thanks @akomiqaia, I updated those test expectations

akomiqaia (approved):

akomiqaia:
@notlee approved but expander and tooltip percy screenshots are quite different. not sure why

Issue #535, opened 2022-01-10
o-forms: add radio or button role to pseudo-radio-link


JakeChampion:
This was mentioned in #532

> Additionally, emulating links with radio buttons maybe confusing for users of assistive technologies and should be given an appropriate role of radio or button as well as a shared label using a fieldset and legend. Further information can be found in the Radios section of the GOV.UK Design System.

notlee:
This one will need some thought and thorough testing. Will a link still work as a link with a new role. Would it be more confusing to present as a form element and work as a link? Do we update products to use the standard square radio buttons instead and deprecate this? Or do we need to change the design?

Context, this was originally added for this myft ui: https://github.com/Financial-Times/o-forms/pull/225

notlee:
Looks like this is pretty much just used on the myft feed page and, especially given its complexity, I would be keen to deprecate `pseudo-radio-link`.

![Screenshot 2023-01-05 at 09 35 20](https://user-images.githubusercontent.com/10405691/210748188-124711d8-0e9c-42a6-bd52-f996d73700c8.png)

On the myft page there is nothing linking the "View" label with the topic/timeline `pseudo-radio-link` elements. Though they look like a form input they behave entirely like links and have `aria-current="page"` attributes accordingly:
```
<div class="feed-toolbar__toggle">
    <div class="feed-toolbar__toggle-label">
        View
    </div>
    <div class="o-forms-input--pseudo-radio-link o-forms-input--feed-toolbar">
        <a class="o-forms-input__link o-forms-input__link--grid-icon o-forms-input__link--current feed-toolbar__toggle-label--link" href="#" aria-current="page" data-trackable="topic">Topic</a>
        <a class="o-forms-input__link o-forms-input__link--list-icon  feed-toolbar__toggle-label--link" href="#" data-trackable="time">Timeline</a>
    </div>
</div>
```

notlee:
☝️ the above is a silly idea, as the buttons don't present as an exclusive option (it would appear as though both can be pressed at the same time, which they cannot). In this case a standard radio box which submits on change would be best, with submit button available for the core experience.

PR #1080, opened 2023-04-05
fix: o-forms, deprecate pseudo-radio-link [OR-62]

notlee:
Links which look like form inputs may be confusing for some users of assistive technologies, as they may be accessed or behave differently than expected. Further, unlike true o-forms radio inputs, there is no fieldset with legend to group the links.

Instead use a standard box style radio button:
https://github.com/Financial-Times/origami/issues/535

### DIFF
--- a/components/o-forms/README.md
+++ b/components/o-forms/README.md
@@ -322,7 +322,9 @@ To show a state label with no text set the modifier class `o-forms-input__state-
 
 ##### Pseudo box radio inputs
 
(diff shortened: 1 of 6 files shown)


notlee:
Let's give the Content Discovery team some time to comment here before merging:
https://financialtimes.slack.com/archives/C04RUBL9G/p1680710152563999

fenglish:
@notlee We are going to remove `pseudo-radio-link` from next-myft-page by this PR https://github.com/Financial-Times/next-myft-page/pull/2657

akomiqaia (approved):

PR #1083, opened 2023-04-11
fix: change outline color for focus links under matisse theme

juanSanchezAlcala:
change outline color to white when focus links on alphaville/matisse template
[ticket](https://financialtimes.atlassian.net/browse/CI-1645)

BEFORE
image

AFTER
image

### DIFF
--- a/components/o-topper/src/scss/_mixins.scss
+++ b/components/o-topper/src/scss/_mixins.scss
@@ -345,6 +345,9 @@
 				color: oColorsMix($foreground, $color-name, 73);
 			}
 		}
+		:focus-visible {
+			outline: 2px solid oColorsByName('white');
+		}
 	}
 }
 


components/o-topper/src/scss/_mixins.scss
```diff
@@ -345,6 +345,9 @@
 				color: oColorsMix($foreground, $color-name, 73);
 			}
 		}
+		:focus-visible {
+			outline: 2px solid oColorsByName('white');
```
notlee, line 349: It would be better to be more specific so we can control other properties of the focus style independently.

```suggestion
			outline-color: oColorsByName('white');
```

Even better, in the future we'll be able to control this by setting a css custom property :)

notlee (approved):

PR #1089, opened 2023-04-13
feat: add professional theme for checkboxes

akomiqaia:


components/o-forms/README.md
```diff
@@ -377,6 +377,9 @@ To align the checkbox to the **right** of its `label`, add the `o-forms-input__r
 	</span>
 ...
 ```
+Square checkbox also supports additional theme:
```
notlee, line 380:
```suggestion
Square checkboxes also support an additional theme:
```

components/o-forms/demos/src/professional.mustache
```diff
@@ -22,5 +22,25 @@
 				</span>
 			</span>
 		</div>
+		<div class="o-forms-field o-forms-field--optional o-forms-field--professional" role="group" aria-labelledby="checkbox-title">
+			<span class="o-forms-title" aria-hidden="true">
+				<span class="o-forms-title__main" id="checkbox-title">Checkboxes for proffessional theme</span>
```
notlee, line 27:
```suggestion
				<span class="o-forms-title__main" id="checkbox-title">Checkboxes for professional theme</span>
```

components/o-forms/src/scss/_mixins.scss
```diff
@@ -28,13 +28,7 @@
...
 		}
 		.o-forms-input__label {
-			&:before {
-				border: 1px solid _oFormsGet("default-border", $theme);
-			}
-
-			&:after {
-				background-color: _oFormsGet("controls-base", $theme);
-			}
+			@include _updateColorThemes($theme);
```
notlee, line 31: `_updateColorThemes` should be namespaced with `_oForms`, `_oFormsUpdateColorThemes` as it could clash with mixins defined in other components (though that's unlikely).

However it's hard to tell what `_updateColorThemes` does based on its name. It sounds quite generic but actually can only apply to specific elements. Are you able to think of a more descriptive name? Alternatively, in this case, I would argue it would be easier to read and maintain if we copy-pasted instead of using a mixin. What do you think?

notlee (commented):
Nice work :) I left a couple of minor comments

components/o-forms/src/scss/_mixins.scss
akomiqaia: @notlee in the end I removed the mixin and just copy pasted. If we have more repetition in the future we can rewrite it into a mixin afterwards. [82bab13](https://github.com/Financial-Times/origami/pull/1089/commits/82bab131107ed7fb0cf6f271f8cbdd41b415a167) reflects this change

notlee (approved):

PR #1091, opened 2023-04-19
feat: add professional inverse theme for chekcboxes

akomiqaia:


notlee (approved):
👍 Looks good. But I don't think there's a way to preview in Storybook? It would be good to follow up with a PR that adds `professional` / `professional-inverse` themes to the TSX and storybook

akomiqaia:
> 👍 Looks good. But I don't think there's a way to preview in Storybook? It would be good to follow up with a PR that adds `professional` / `professional-inverse` themes to the TSX and storybook

[1ce452b](https://github.com/Financial-Times/origami/pull/1091/commits/1ce452b8877f880db67390bf8456419407f142de) should add SB (storybook) demos and also fixed a little bug as well

notlee (approved):

PR #1094, opened 2023-04-25
feat: options can be rendered as already selected OR-184

akomiqaia:


components/o-multi-select/src/tsx/multi-select.tsx
```diff
@@ -1,16 +1,21 @@
...
+	label: string;
+	selected: boolean;
+}
+
 export function MultiSelect({
 	multiSelectOptions,
 	id,
 }: {
-	multiSelectOptions: string[];
+	multiSelectOptions: MultiSelectType[];
```
notlee, line 10: This is a breaking change. We need to release a new major or support both `string[]` and `MultiSelectType[]`

components/o-multi-select/src/js/multi-select-options.js
```diff
@@ -92,14 +92,15 @@ function createOptionButton(option, index) {
...
  * @returns {HTMLElement} The newly created option element.
  */
-export function createOption(idBase, option, index) {
+export function createOption(idBase, option, index, selected = false) {
 	const optionEl = document.createElement('div');
 	optionEl.setAttribute('role', 'option');
 	optionEl.id = `${idBase}-${index}`;
 	optionEl.className = 'o-multi-select-option';
-	optionEl.setAttribute('aria-selected', 'false');
+	optionEl.setAttribute('aria-selected', selected);
```
notlee, line 103: I think this should be true/false?
https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaSelected#value

notlee (changes requested):
Nice work fixing this quickly for Marlon :) It would be good to add tests for this also

notlee (changes requested):
You also need to write a Migration Guide and update the table in the README to release a major. Both work, but probably it would be easier to support both TSX interfaces 😀

frshwtr:
Note from pairing: The enhanced experience doesn't update the core values when selected. Ideally it should do this if a form gets submitted off the back of this.

notlee (approved):

PR #1095, opened 2023-04-27
fix: o-header, subnav, improve scroll button behaviour.

notlee:
1. Viewport resize: Previously, o-header would only respond to viewport resize if the user had initialised `o-viewport`. This was undocumented. Let us always initialise `o-viewport` and include it as a dependency.
2. Viewport resize part 2: The wrapper width was calculated and stored on initialisation. This value is then wrong after resize, breaking the calculation to decide whether to show scroll buttons.
3. DOM mutation: When subnav items change, recalculate subnav scrollability. This supports single page application (or similar), where navigation items are updated dynamically.

### DIFF
--- a/components/o-header/package.json
+++ b/components/o-header/package.json
@@ -34,7 +34,8 @@
     "@financial-times/o-toggle": "^3.0.0",
     "@financial-times/o-typography": "^7.0.1",
     "@financial-times/o-utils": "^2.0.0",
-    "@financial-times/o-visual-effects": "^4.0.1"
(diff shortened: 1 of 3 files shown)


wheresrhys (commented):
just a thought - does o-header-services need the same fix?

notlee:
@wheresrhys Doesn't look like it, a different implementation directly listening to the browser's resize event. Arguably could benefit from enhancement 3) but we haven't seen a request raised around that usecase.

akomiqaia (approved):

PR #1101, opened 2023-05-02
OR-197-omit-events-from-multi-select

Marlon-Buckley:


components/o-multi-select/README.md
```diff
@@ -106,6 +106,18 @@ const oMultiSelectElement = document.getElementById(
 oMultiSelect.init(oMultiSelectElement);
 ```
 
+`o-multi-select` also dispatches custom event `oMultiSelect.OptionChange` that bubbles and gets triggered when option gets selected or removed. The event data has following interface:
```
notlee, line 109: Hey, looks good, just some little suggestions here which might improve readability 😄 Mainly using a heading and short sentences.

```suggestion
### Events 

`o-multi-select` dispatches a custom event `oMultiSelect.OptionChange` that bubbles. It is triggered when an option gets selected or removed. The event data has following interface:
```

notlee (approved):

PR #1103, opened 2023-05-04
Or 180 fix dac issues

akomiqaia:


components/o-multi-select/src/js/state.js
```diff
@@ -16,12 +16,26 @@ export function updateState() {
...
+				element.children[0].setAttribute('tabindex', '-1');
+			});
 		} else {
 			this._selectedOptions.classList.remove('o-multi-select__visually-hidden');
+
+			// Remove tabindex from all children of the selectedOptions element
+			// to make them focusable when the selectedOptions element is visible.
+			[...this._selectedOptions.children].forEach(element => {
+				element.children[0].removeAttribute('tabindex');
+			});
```
notlee, line 38: Instead of visually hiding have you considered setting `display: none` instead? Then it wouldn't be necessary to set the tabindex, simplifying this function.

components/o-multi-select/src/js/utils.js
```diff
@@ -85,6 +85,12 @@ export function onComboBoxKeyDown(event) {
 		addOptionToList.call(this);
 	}
 	this._updateCurrentElement();
+	if (this._open && isScrollable(this._listboxEl)) {
+		const currentOption = this._listboxEl.querySelector(
+			'.o-multi-select-option__current'
+		);
+		maintainScrollVisibility(currentOption, this._listboxEl);
+	}
```
notlee, line 93: this appears to work really well 👍

notlee (approved):
I have one suggestion which I think could simplify the codebase but no blockers, looks good 😄

components/o-multi-select/src/js/state.js
akomiqaia: I need the children's width to calculate whether I need to hide them so I can not set `display: none`. Maybe you have suggestions on how else this can be done?

PR #1108, opened 2023-05-09
fix: way obtain scrollContainer element

juanSanchezAlcala:
change the way to obtain scrollContainer element. 
Why ? Previously we were using document.querySelector which on the app was getting a wrong element since there were several elements/containers with that class. This way we ensure that we are getting the container which is parent of the element where the comments are in

### DIFF
--- a/components/o-comments/src/js/stream.js
+++ b/components/o-comments/src/js/stream.js
@@ -78,7 +78,7 @@ class Stream {
 				if(this.options.addClass){
 					containerClassName.push(this.options.addClass);
 				}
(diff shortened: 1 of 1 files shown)


AliceArmstrong (commented):
This change looks good to me, and looking at the code I think I can infer why we are doing this change, but it would be good to have the reasoning behind the change written in the PR description in case people want to look up this change in the future 🙂

AliceArmstrong (approved):

notlee (approved):

PR #1113, opened 2023-05-18
Box radio professional [OR-163]

notlee:
Add box radio buttons for FT Professional

Screenshot 2023-05-18 at 08 49 48
Screenshot 2023-05-18 at 10 25 56

(This PR also removes a demo for Custom Pseudo Radio Links, as it it deprecated – see commit messages)


components/o-forms/src/scss/_checkbox.scss
```diff
@@ -87,4 +88,47 @@
 			}
 		}
 	}
+
+	@each $theme in $themes {
+		@if _oFormsSupports('#{$theme}') {
+		.o-forms-field--#{$theme} {
+			.o-forms-input--checkbox {
+				input[type=checkbox] {// stylelint-disable-line selector-no-qualifying-type
```
frshwtr, line 96: This disable rule seems to be used a lot. Is it worth disabling this particular rule in config if we disagree with it?

frshwtr (approved):
LGTM

components/o-forms/src/scss/_checkbox.scss
notlee: I don't disagree with the rule I disagree with the code 😄 We can't change the approach without new markup though, a breaking change.

notlee:
Percy looks good

PR #1116, opened 2023-05-18
fix: change style shadow dom container

juanSanchezAlcala:
We are having an issue on safari where the comments overflow the parent container.
We can fix that and setting 
**display: inline-block!important
width:100%.**
We need to set the **!important** due the Coral script introduce an inline style on that element setting display:block

[ticket](https://financialtimes.atlassian.net/browse/CI-1725)

### DIFF
--- a/components/o-comments/src/scss/coral-talk-iframe/main.scss
+++ b/components/o-comments/src/scss/coral-talk-iframe/main.scss
@@ -21,7 +21,8 @@
 	}
 	#coral {
 		--ft-magnify: 1.31282;
-		display: inline-block!important;
+                 // See https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1653?view=detail&selectedIssue=CI-1725
+		display: inline-block !important;
 		width:100%;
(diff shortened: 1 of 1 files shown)


notlee (approved):
It would be great to add your `!important` explanation as a comment. 👍 Has someone given Coral feedback so they can look to fix this their end?

components/o-comments/src/scss/coral-talk-iframe/main.scss
```diff
@@ -21,6 +21,8 @@
 	}
 	#coral {
 		--ft-magnify: 1.31282;
+		display: inline-block!important;
```
umbobabo, line 24: Suggestion: let's fix this for now but report it to Coral, we can remove this code when they have fixed it their side

```suggestion
                 // See https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1653?view=detail&selectedIssue=CI-1725
		display: inline-block !important;
```

umbobabo (approved):

PR #1120, opened 2023-05-24
Professional links [OR-196]

notlee:
This is a bit wild. I'm looking forward to dropping the complexity of some of our Sass in the future. For now I wanted to add FT Professional support to existing mixins `oTypographyBody` and `oTypographyLink` but also introduce new wrapper classes so that all child links/body copy can have the professional style applied without calling the mixins / applying a class on every single link. This works in a more similar way to how o3 will work:
- `o-typography-inverse` - Make body copy and links white, for use on a dark background.
- `o-typography-professional` - Change link colours for the FT Professional sub-brand (core brand only)
- `o-typography-professional o-typography-inverse` - A version of the inverse theme for use with the FT Professional sub-brand (core brand only)

I have left fallback declarations for now, for browsers which don't support CSS Customer Properties, though we don't really need to going forward.

professional:
Screenshot 2023-05-24 at 11 40 01

professional + inverse:
Screenshot 2023-05-24 at 11 40 27

inverse
Screenshot 2023-05-24 at 11 40 31


frshwtr (approved):
It's not too wild, I think what you're aiming to do here is sensible and helps get things into position to reduce sass dependency in future.

PR #1122, opened 2023-05-24
feat: Add :read-only styles to match the disabled styles for inputs

chee:
Apparently HTML has a(nother) [way of making inputs immutable](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly)

We use this in Spark for readonly text areas, and are currently overriding the o-forms styles to make it look more disabled.

In this PR i've made it always match the :disabled styles, but i'm not sure if that is always what people would want(?). I can also see an argument for this being a feature, a breaking change or a patch... I'll leave it in your wise hands

### DIFF
--- a/components/o-forms/src/scss/shared/_single-inputs.scss
+++ b/components/o-forms/src/scss/shared/_single-inputs.scss
@@ -19,7 +19,8 @@
 /// @access private
 /// @output Shared disabled state styles for inputs.
 @mixin _oFormsInputDisabled {
-	&:disabled {
+	&:disabled,
+	&:read-only {
 		cursor: default;
 		color: _oFormsGet('disabled-text');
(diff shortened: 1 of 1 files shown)


notlee (approved):
We can use selector list arguments!
I learnt 2 things today.

https://docs.google.com/document/d/1z6kecy_o9qHYIznTmqQ-IJqre72jhfd0nVa4JMsS7Q4/edit#heading=h.er7no6jwfim1
https://caniuse.com/css-not-sel-list

chee:
hedgehog nonsense

notlee (changes requested):
As in Slack:
Oh this didn't work because select and some other form inputs are always read-only 🙈 
I guess we can update both selectors to be `input` specific.
Screenshot 2023-05-24 at 14 39 18

PR #1124, opened 2023-05-24
fix: o-typography, restore default link colour

notlee:


notlee:
We missed this despite Percy flagging because I forgot to check... 
https://github.com/Financial-Times/origami/pull/1120

As we're out of hours and this is relatively simple I'm going to merge this without review, so we can mitigate against this visual bug making it to production.

PR #1126, opened 2023-05-25
feat: o-banner, add FT Professional theme [OR-164]

notlee:
Screenshot 2023-05-25 at 09 53 30


notlee:
want to rename this professional-inverse

akomiqaia (approved):
LGTM. And I think you are right to rename to `professional-inverse`. I will still approve it since it just the theme name change

PR #1127, opened 2023-05-25
feat: autocomplete, add a suggestionTemplate override option

adgad:
### Problem

I'm trying to use o-autocomplete in our TagMe rewrite. However, TagMe (which doesn't currently use origami) has some custom rendering of it's suggestions, with additional context to help editorial decide which tag to choose.

Screenshot 2023-05-25 at 16 03 16

### Solution

Don't know if this is the best way - but allow the user to override the `suggestionTemplate` function. Open to better ideas, or if I should even be using this component!

Screenshot 2023-05-25 at 17 09 07

This seems to work well enough. The only situation it doesn't work in is with the defaultValue (which is a string), but this can be compensated for in the override function itself.

### DIFF
--- a/components/o-autocomplete/README.md
+++ b/components/o-autocomplete/README.md
@@ -157,6 +157,7 @@ If the `source` function is returning an array of strings which are already suit
 
 The most common scenario which requires having to define a `mapOptionToSuggestedValue` function is when the `source` function is returning an array of objects, where one of the properties in the object should be used as the suggestion.
 
+
 #### Example
 
 ```js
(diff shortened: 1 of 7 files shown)


notlee:
Thanks for the excellent PR 😄 Demos and docs and everything.

Happy to accept this. Could you add some validation (maybe an allow list of elements?) and guidelines to help ensure only presentation elements are used within each option?
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/option_role#all_descendants_are_presentational

adgad:
Hmm, validation might be tricky since the output is going to be a string 🤔 I can put a caution in the README, like they've got on the [accessible-autocomplete](https://github.com/Financial-Times/accessible-autocomplete#templates-default-undefined) docs.

Or perhaps if the `suggestionTemplate` function returns a DOM element instead of a string, within o-autocomplete we can validate that, and then send the `outerHTML` to accessible-autocomplete? whaddya reckon?

notlee:
Sorry @adgad I only just noticed your reply. If we want to keep the current interface maybe we could create a temporary [DocumentFragment](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) within o-autocomplete and iterate over child nodes to validate? Guidelines would definitely be good too

adgad:
[discussed offline]

Adding validation will be tricky, because the `suggestionTemplate` function gets called for every suggestion at every keystroke. Parsing the HTML string this frequently will be expensive. Additionally, we wouldn't be able to validate it once at the beginning, as we won't know the shape of the `option` argument.

As a result, we've decided to add guidelines to the readme for now, and trust our engineers and testing processes

PR #1137, opened 2023-06-02
backstage: update o-tracking events

leannecornish-ft:
### What
Updating the tracking event types in o-overlay, from `overlay` to `component`.

### Why
The analytics team have switched over to components, as it offers greater reporting insight.

### Impact
This change will mean existing usage of `overlay` for analysis will no longer receive events.


leannecornish-ft:
Closing as there are some apps still using the `overlay` event type.

PR #1144, opened 2023-06-15
backstage: dont run percy unless correct label is applied on the PR

akomiqaia:
Creates PR template (which is very basic and need more love)
And introduces a git label to run workflows. If the git label is not `Run Percy` Percy jobs should not run.

### DIFF
--- a/.github/workflows/percy.yml
+++ b/.github/workflows/percy.yml
@@ -25,7 +25,7 @@ on:
 
 jobs:
   percy-components-ft-concept-button:
-    if: ${{ github.ref = 'refs/heads/main' || github.event.label.name == 'percy' }}
+    if: ${{ github.ref == 'refs/heads/main' || github.event.label.name == 'percy' }}
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
@@ -46,7 +46,7 @@ jobs:
         PERCY_TOKEN: ${{ secrets.PERCY_TOKEN_COMPONENTS_FT_CONCEPT_BUTTON }}
         WORKSPACE: components/ft-concept-button
   percy-components-g-audio:
-    if: ${{ github.ref = 'refs/heads/main' || github.event.label.name == 'percy' }}
+    if: ${{ github.ref == 'refs/heads/main' || github.event.label.name == 'percy' }}
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
+## Describe your changes
```
frshwtr, line 1: Nice!

frshwtr (approved):

notlee (dismissed):
I don't think these changes meet this part of the ticket?

>Maintain existing behaviour for chore and main Percy runs. This will ensure tests are always run before a release and screenshots are always up to date.

If we don't always run on merge to main, then the baseline screenshots will become out of date, and future PR runs may compare against old screenshots.

frshwtr:
> I don't think these changes meet this part of the ticket?
> 
> > Maintain existing behaviour for chore and main Percy runs. This will ensure tests are always run before a release and screenshots are always up to date.
> 
> If we don't always run on merge to main, then the baseline screenshots will become out of date, and future PR runs may compare against old screenshots.

@notlee If I'm reading correctly, the workflows will run on main, but will not execute the percy scripts due to the `if` statements? And your suggestion is that percy should still run on main every time?

ka-tsu-ra:
Looks like you've added a condition to run the percy tasks if main branch is building. (I've not used github actions before so I'm trusting that you know `github.ref` and `refs/heads/main` work)

But should the condition be using a `==` i.e. `github.ref == 'refs/heads/main'` rather than `github.ref = 'refs/heads/main'`?

frshwtr:
> Looks like you've added a condition to run the percy tasks if main branch is building. (I've not used github actions before so I'm trusting that you know `github.ref` and `refs/heads/main` work)
> 
> But should the condition be using a `==` i.e. `github.ref == 'refs/heads/main'` rather than `github.ref = 'refs/heads/main'`?

Yes you're right. Good spot, thanks :)

akomiqaia:
This would trigger anytime we merge with the main right? And this might cause us to rerun Percy jobs even when we don't have a tag on it right? My assumption was that since we merge PR with Percy label then if statement should be true and Percy screenshots should get updated. But if we don't apply Percy label that means we don't have to run Percy jobs.

frshwtr:
> This would trigger anytime we merge with the main right? And this might cause us to rerun Percy jobs even when we don't have a tag on it right? My assumption was that since we merge PR with Percy label then if statement should be true and Percy screenshots should get updated. But if we don't apply Percy label that means we don't have to run Percy jobs.

`ref` on this context object references the branch which triggered the Github Action. We would want this behaviour as we still need to run on main once merged so that screenshots are up to date. See [here](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)

ka-tsu-ra:
From my non-origami pov this looks ok :)

ka-tsu-ra (approved):

PR #1151, opened 2023-06-26
fix: o-colors, always mix with FT colours, not web colours

notlee:
So, Sass automatically makes any web colour a hex value. https://www.w3.org/wiki/CSS/Properties/color/keywords

If you set backgorund: wheat in the browser, the browser has a hex code for that and you get a background. Sass runs with this and treats wheat as a colour, `type-of(wheat)` is color. Meanwhile `type-of(paper)` is a string.

```scss
.example {
    // https://www.w3.org/wiki/CSS/Properties/color/keywords
	content: type-of(paper);
	content: type-of(not-a-web-color);
	content: type-of('wheat');
        content: type-of(wheat);
	content: type-of(hotpink);
}
```

```css
.example {
  content: string;
  content: string;
  content: string;
  content: color;
  content: color;
}
```

So, if you don't pass a string explicitly to `oColorsMix`, we may intentionally mix a standard web colour instead of the correct FT colour.


frshwtr (changes requested):
There are some broken Percy tests on this. Will investigate whilst you're on hols.

frshwtr (approved):

PR #1155, opened 2023-06-30
feat: update g-audio interface with teal/pink themes

joannaskao:
This does not currently work for multiline inline players yet


notlee:
Hey @joannaskao! As discussed, I just happened to spot your draft PR and had some early feedback to share. Hope it's helpful!

- The primary Sass mixin of Origami components shouldn't be used for configuration, just for outputting styles of selected features. It could be confusing to make an exception. Checkout our tutorial on creating a new Origami component, it includes [a section on supporting multiple themes](https://origami.ft.com/documentation/tutorials/create-a-new-component-part-3/#component-themes) you might find helpful. We don't require themes are managed this way though, if you'd prefer to stay with your current approach but output modifier classes `.gaudio--pink` to support multiple themes.
- Mixins also need to be namespaced as we we still rely on `@include` rather than dart-sass' module system e.g. `teal-theme` would become `gAudioTealTheme`.
- I'd recommend prepending an underscore to your new variables and mixins to indicate they are private to the component. That'll give you much more flexibility to refactor without a breaking change :) Though if you wanted to follow the tutorial above to manage themes in a similar way to other Origami components, that would remove the need to define lots of variables.
- Lots of hex codes here! 😱 It's hard to know what colour they are, and whether they are part of the FT's palette or unique to this component. Could we use [o-colors Sass functions](https://registry.origami.ft.com/components/o-colors@6.6.1/sassdoc?brand=core) here instead of direct hexes?

notlee:
Closing this PR as it's been a while, and I believe you handled this as a one off. Happy to open again if you'd like to revisit.

PR #1169, opened 2023-07-12
feat: Allow JSX Elements to be passed in as relatedContent

cebirmingham:
## Describe your changes

Added the ability to pass JSX Elements into relatedContent for o-header-services. Also fixed the bug where the relatedContent `href` was set to be the label.

### DIFF
--- a/components/o-header-services/stories/fixtures.tsx
+++ b/components/o-header-services/stories/fixtures.tsx
@@ -33,7 +33,7 @@ export const relatedContent: ListItem[] = [
 	{
 		label: 'Sign in',
 		url: '/login',
-	}
+	},
 ];
 export const primaryNavData: NavItem[] = [
 	{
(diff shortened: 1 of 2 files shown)


components/o-header-services/stories/fixtures.tsx
```diff
@@ -34,6 +34,7 @@ export const relatedContent: ListItem[] = [
 		label: 'Sign in',
 		url: '/login',
 	},
+		<button onClick={() => window.alert('Hello')}>Say Hello</button>,
```
notlee, line 37: We now try to avoid customisation in our demos, so people know what Origami provides directly, and can copy html to get the same result. So I think we should revert the changes in `/stories/fixtures.tsx`. Good to see your changes work for review purposes though! :) In the future we'll introduce tsx documentation and testing for custom cases like this (sorry we don't have that now!)

Screenshot 2023-07-12 at 13 22 46

notlee (changes requested):
Think we should revert the storybook demo, but the functionality looks good, thanks! :)

notlee (approved):

PR #1178, opened 2023-07-13
fix: re order lists so that the contents are correctly removed when d…

frshwtr:
…isplayed on the registry

## Describe your changes

With the previous release, the first list is omitted when rendered on the Registry.

image

This is because the Registry's readme builder will [remove the first list in the rendered HTML](https://github.com/Financial-Times/origami-registry-ui/blob/e0198010cdb31b2cff1f49ba4cb7ad039f856adf/lib/code-docs/readme/index.js#L49). The MD parser seems to have merged this first list with the contents, and as the above code is identifying the whole block as the Nav, it removes it.

Re-ordering this so that the copy list comes after the content list should fix this.

### DIFF
--- a/components/o-grid/README.md
+++ b/components/o-grid/README.md
@@ -22,13 +22,14 @@ A 12 column responsive, flexbox-based grid system for laying out documents, temp
 -   [Contact](#contact)
 -   [Licence](#licence)
 
+This component is a collection of Sass styles to build a 12 column grid system, with a few JavaScript helpers.
+
 Our grid consists of:
 -	12 columns, of equal size
 -	5 layouts (breakpoints) to support devices of various sizes
 -   gutters of variable width, that adjust according to the current layout
 -   a maximum page width of 1220px, by default
 
-This component is a collection of Sass styles to build a 12 column grid system, with a few JavaScript helpers.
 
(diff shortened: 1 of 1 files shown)


components/o-grid/README.md
```diff
@@ -28,6 +22,12 @@ Our grid consists of:
 -   [Contact](#contact)
 -   [Licence](#licence)
 
+Our grid consists of:
+-	12 columns, of equal size
+-	5 layouts (breakpoints) to support devices of various sizes
+-   gutters of variable width, that adjust according to the current layout
+-   a maximum page width of 1220px, by default
+
 This component is a collection of Sass styles to build a 12 column grid system, with a few JavaScript helpers.
```
notlee, line 31: This reads a little weirdly, how about moving below this line?

- What the component is
- What the grid consists of
- A diagram illustrating the former point

components/o-grid/README.md
```diff
@@ -28,6 +22,12 @@ Our grid consists of:
 -   [Contact](#contact)
 -   [Licence](#licence)
 
+Our grid consists of:
+-	12 columns, of equal size
+-	5 layouts (breakpoints) to support devices of various sizes
+-   gutters of variable width, that adjust according to the current layout
```
notlee, line 28: Some funky indentation above here and it would be great to capitalise the start of the sentence

frshwtr:
@notlee thanks, resolved both

notlee (approved):

PR #1183, opened 2023-07-13
Use build service on dev

akomiqaia:
## Describe your changes
Use build serviece instead of sass


notlee:
What's the benefits of `getStyles` vs. putting a single Origami Build Service request in `Layout.astro`? This seems quite complex and I suspect will result in slightly worse performance for users.

apps/astro-website/src/helpers/getStyles.js
```diff
@@ -0,0 +1,27 @@
...
+		if (!dependenciesList.includes(pkg)) {
+			throw new Error(`Package ${pkg} needs to be installed`)
+		}
+		const packageVersion = allDeps[pkg]
+		componentsWithVersions.push(`${component}@${packageVersion}`)
+	})
+
+	const componentsQuery = componentsWithVersions.join(",")
+
+	return `<head>
```
frshwtr, line 21: I think the `<head>` tags should be moved out and moved to surround the fragment in `Header.astro:38`, so that other sources can be included if needed.

akomiqaia:
> 

When Astro builds the project it generates `html` files for each page and at the moment all the css build links can go inside the  `Layout.astro` file and that would be fine. But I guess `getStyles` can help us if we only need to get styles only for one page and also this function figures out what version of components it should get from the build service based on `package.json`. what performance issues can we encounter with this approach?

apps/astro-website/src/helpers/getStyles.js
akomiqaia: This is a good shout. I wander is Astro has some sort of function that can just add links in `head`.

notlee:
That's pretty awesome @akomiqaia, I love that it's relying on package.json which makes package management across the app easier + ensures a flat dependency tree. But I think this might be a preemptive performance optimization which might not lead to benefits, but does add complexity. One thing to consider in Origami's case is that some of our components are pretty large, including low level components like o-buttons. If they're included on separate pages with other components they'll need to be downloaded multiple times. If all components are loaded up front in one Origami Build Service service request the initial request may be slower but: 1) it will be cached by the browser, so subsequent page load is much faster 2) it will benefit from minification and compression significantly, as we have _a lot_ of repeated CSS declarations in CSS – o-buttons CSS is a whopping 342.09kB, but compressed is only 11.96 kB.

akomiqaia:
cool updated the code. So everything is initialised at layout level

notlee (approved):
👍

PR #1200, opened 2023-07-18
backstage: update md frontmatter and render blogs correctly

akomiqaia:
## Describe your changes
Update newsletters frontmatter and use `tldr` as one of the variables. This pr changes how routing works compared to the old website. ~~In old website pathname for a blog post is determined by publish date and tags (for example `blog/2023/06/21/sass/`) but in Astro, these paths are updated to `blog/2023-06-21-sass`. if we want to keep the original routing we have to create folders in the Astro pages folder for each of them. We are kind of forced to use a new way of routing. Hopefully, this won't be too problematic. and possibly we might want to redirect from old routes to this.~~ this was reverted by last commit so not relevant anymore 

### DIFF
--- a/apps/astro-website/README.md
+++ b/apps/astro-website/README.md
@@ -28,9 +28,7 @@ The formal Origami specification lives in the [`_specification-v1` folder](_spec
 
 ### Blog posts
 
-[TODO]
-
-Blog posts live in the [`_posts` folder](_posts) and the file names are prefixed with the post date. Posts can include an `author` frontmatter value to signify who wrote the post, and an array of `tags` which are displayed alongside the post. The `description` frontmatter is particularly important for blog posts as it is displayed as a preview on the blog listing.
+Blog posts live in the [`posts` folder](src/content/posts) and the file names are prefixed with the post date. Posts can include an `author` frontmatter value to signify who wrote the post, and an array of `tags` which are displayed alongside the post. The `description` frontmatter is particularly important for blog posts as it is displayed as a preview on the blog listing. The `publishDate` frontmatter for a date that newsletter was published. And `tldr` frontmatter is used to display a summary of the post before the main content. If it is omitted the description will be used instead.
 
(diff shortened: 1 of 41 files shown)


notlee:
If we change the routes we would need to add redirects for existing posts. We could do that, but we could avoid this by renaming `pages/blog/[slug].astro` to `pages/blog/[...slug].astro` – this would match the full path `blog/2023/06/21/sass/`. Then we'd also need to update `getStaticPaths` to map to the md files. A regex was the first solution which came to my mind but I'm open to more readable alternatives:

https://regexper.com/#%2F%28%5B%5Cd%5D%7B4%7D%29-%28%5B%5Cd%5D%7B2%7D%29-%28%5B%5Cd%5D%7B2%7D%29-%28.*%29%2Fg
```
	return postEntries.map(entry => ({
		params: {
			slug: entry.slug.replace(
				/([\d]{4})-([\d]{2})-([\d]{2})-(.*)/g,
				"$1/$2/$3/$4"
			),
		},
		props: {entry},
	}))
```

akomiqaia:
> If we change the routes we would need to add redirects for existing posts. We could do that, but we could avoid this by renaming `pages/blog/[slug].astro` to `pages/blog/[...slug].astro` – this would match the full path `blog/2023/06/21/sass/`. Then we'd also need to update `getStaticPaths` to map to the md files. A regex was the first solution which came to my mind but I'm open to more readable alternatives:
> 
> 
> https://regexper.com/#%2F%28%5B%5Cd%5D%7B4%7D%29-%28%5B%5Cd%5D%7B2%7D%29-%28%5B%5Cd%5D%7B2%7D%29-%28.*%29%2Fg
> 
> ```
> 	return postEntries.map(entry => ({
> 		params: {
> 			slug: entry.slug.replace(
> 				/([\d]{4})-([\d]{2})-([\d]{2})-(.*)/g,
> 				"$1/$2/$3/$4"
> 			),
> 		},
> 		props: {entry},
> 	}))
> ```

@notlee updated and this should get us old routing

notlee (approved):

PR #1202, opened 2023-07-19
Document Principles part moved to content

Behnazz:
## Describe your changes
- moved principle .md files to content under the folder documents
- update the principle .astro files to work with content
- add meta tag to the <head> to fix the uft-8 bug


apps/astro-website/src/layouts/Head.astro
```diff
@@ -1,7 +1,7 @@
 <head>
+	<meta charset="utf-8" />
```
Behnazz, line 2: this helps with some weird letter we may see when rendering an .md file

akomiqaia (approved):
LGTM

PR #1208, opened 2023-07-21
Astro emails

akomiqaia:
## Describe your changes
Update the way we used to send-newsletters. 


apps/astro-website/README.md
```diff
@@ -34,26 +34,23 @@ Newsletter blog posts must have an accompanying email, this is outlined in the s
...
+2. Write the newsletter. This is best done in the blog post, as this is standard Markdown.
 
 3. Open a pull-request on this repo, and get it approved by another member of the team. Once approved, merge into master. This must be done _on_ or _after_ the publish date indicated by the post file name. The blog post is now published, check it on the live site.
 
-4. Now you'll need to review and publish the email, this is done in a locally cloned copy of the repo (this step may be automated later when we have confidence in the email format). You'll need to have npm dependencies installed: `npm install`
+4. Now you'll need send the newsletter as an email. This is done from your machines CLI. The email newsletter relies on some pretty gross Liquid templating to ensure that we avoid copy/paste errors in the HTML we created [email page template](src/pages/emails/[slug].astro) in Astro. The email also has required front-matter: a `title` which becomes the subject of the email.
```
notlee, line 49: Thanks for updating the documentation as you go. You might want to double check though, for example this section mentions Liquid templates which is no longer correct.

apps/astro-website/src/helpers/email-newsletter.js
```diff
@@ -0,0 +1,131 @@
...
+		console.error("")
+		console.error("Status code:", error.statusCode)
+		console.error("")
+		console.error("Specify the newsletter HTML with an `EMAIL_SOURCE_HTML`")
+		console.error("environment variable and make sure the email is live.")
+		console.error("")
+		process.exit(1)
+	}
+
+	// Add a max width to all images
```
notlee, line 44: Should this now happen as part of the astro page, with the other dom manipulation?

apps/astro-website/src/helpers/email-newsletter.js
```diff
@@ -0,0 +1,131 @@
...
+		path.join(__dirname, "public/styles/email.css"),
+		"utf8"
+	)
+	htmlContent = htmlContent.replace(
+		'<link rel="stylesheet" href="/styles/email.css">',
+		`<style type="text/css">${emailStyles}</style>`
+	)
+
+	// Generate the plain text content
+	const plainTextContent = htmlToText.convert(htmlContent, {
```
notlee, line 60: We used to be able to preview the plain text version to make sure it was okay. Have we lost that ability now?

apps/astro-website/src/layouts/DocLayout.astro
```diff
@@ -2,30 +2,37 @@
...
+		oSyntaxHighlight.init(null)
+	</script>
 
-<script>
-	import oLayout from "@financial-times/o-layout/main"
-	import oSyntaxHighlight from "@financial-times/o-syntax-highlight/main"
-	oLayout.init(null)
-	oSyntaxHighlight.init(null)
-</script>
+	<div class="o-layout o-layout--docs" data-o-component="o-layout">
```
notlee, line 22: We should add the `body` tag here and otherwise confirm the html is valid

apps/astro-website/src/layouts/Layout.astro
```diff
@@ -2,15 +2,24 @@
...
-	<div class="o-layout__footer">
-		<Footer />
+<!DOCTYPE html>
+<html
+	lang="en-GB"
+	class="core o-typography--loading-sans o-typography--loading-sans-bold o-typography--loading-display o-typography--loading-display-bold"
+	data-o-component="o-typography"
+>
+	<Head title={title} description={description} />
+	<div class="o-layout o-layout--landing">
```
notlee, line 16: `body`

apps/astro-website/src/pages/blog/[...slug].astro
```diff
@@ -54,3 +56,17 @@ const tags = entry.data.tags.length > 0 && entry.data.tags.join(",")
 	</p>
 	<Content />
 </DocLayout>
+
+{
+	hasCharts && (
```
Comment (notlee) [inline apps/astro-website/src/pages/blog/[...slug].astro:61]: What is `hasCharts` for? If this is for only 1 blog post could this live in that blog posts md file?

apps/astro-website/src/pages/emails/[slug].astro
```diff
@@ -0,0 +1,353 @@
+---
+import * as cheerio from "cheerio"
+import {getCollection} from "astro:content"
+
+import {EMAIL, SLACK} from "@/helpers/variables"
+
+export async function getStaticPaths() {
+	const postEntries = await getCollection("posts")
```
Comment (notlee) [inline apps/astro-website/src/pages/emails/[slug].astro:8]: Will this render every post as an email? We should probably have a mechanism to limit these only to posts we intend to send as emails. We may want to email something that isn't a newsletter. Should we store meta data perhaps in the frontmatter to indicate a post should be emailed?

apps/astro-website/src/pages/emails/[slug].astro
```diff
@@ -0,0 +1,353 @@
...
+	return postEntries.map(entry => {
+		const splitSlug = entry.slug.split("-")
```
Comment (notlee) [inline apps/astro-website/src/pages/emails/[slug].astro:10]: We require a specific naming convention here for any post which is emailed, that should be documented

apps/astro-website/src/pages/emails/[slug].astro
```diff
@@ -0,0 +1,353 @@
...
+const companionPostUrl = `https://origami.ft.com/blog/${dateSlug}/${stringName}/`
+
+const MarkdownFiles = await Astro.glob("@/content/posts/*.md")
+const currentMarkdown = MarkdownFiles.filter(md =>
+	md.file?.includes(entry.id)
+)[0]
+const currentMdContent = currentMarkdown.compiledContent()
+const $ = cheerio.load(currentMdContent)
+const $p = $("p")
+$p.prop(
```
Comment (notlee) [inline apps/astro-website/src/pages/emails/[slug].astro:40]: This is disgusting. I love it. Actually think this is a better setup to before. No more awkwardly copy-pasting content and forgetting to update the copied version after making changes!

apps/astro-website/src/pages/emails/[slug].astro
Comment (notlee) [reply apps/astro-website/src/pages/emails/[slug].astro:10]: I also get a different newsletter than production.

http://localhost:3000/emails/newsletter-2020-06/ 👉  May 2020
https://origami.ft.com/emails/newsletter-2020-06/ 👉 June 2020

This post to email slug mapping feels a bit shaky. Do you know why we host these newsletters, is it only to send them? I'm thinking perhaps we should update their url structure to be easier to maintain now they're derived directly from posts. And implement a redirect only if it's not too hard, or we expect people to actually visit them because of a reason I've missed. Thoughts?

apps/astro-website/README.md
akomiqaia: fair enough. I think I misread this and read like Email templates were gross to write :D Will arrange wording

apps/astro-website/src/layouts/DocLayout.astro
akomiqaia: in HTML5 `body` is not a mandatory tag to apply on an element but it can be much better for clarity.

apps/astro-website/src/pages/blog/[...slug].astro
Comment (akomiqaia) [reply apps/astro-website/src/pages/blog/[...slug].astro:61]: yeah. It is only one md file that has that and I thought bringing MDX is part of another branch and I should do it there but maybe we can do it here.

apps/astro-website/src/helpers/email-newsletter.js
```diff
@@ -63,10 +60,51 @@ async function sendNewsletter(options) {
...
+
+		// rd.question("Does plain text look as expected? (yes/no)", answer => {
+		// 	reviewReplies.plainText = answer
+		// 	rd.close()
+		// })
+
+		// rd.question("Does HTML look as expected? (yes/no)", answer => {
+		// 	reviewReplies.html = answer
+		// 	rd.close()
+		// })
```
notlee, line 90: Can these be deleted?

notlee (approved):

PR #1209, opened 2023-07-21
backstage: run install deps before matrix executes

frshwtr:


.github/workflows/test.yml
```diff
@@ -179,14 +179,8 @@ jobs:
             - 'apps/storybook/**'
           apps/website:
             - 'apps/website/**'
-  test:
+  check-out-and-install:
```
frshwtr, line 182: This isn't going to work, you need to restore a cache of the cached files before using it in `test`

PR #1212, opened 2023-07-31
fix: navigation bar

akomiqaia:
- rename content from documents to documentation to math the name of a content to a page path


apps/astro-website/src/components/header/SecondaryNav.astro
```diff
@@ -1,35 +1,42 @@
...
-		if (breadCrumbsArray.length != 1) {
-			return
-		}
-		const url = `${breadCrumb}`
-		children.push({
-			label: capitaliseFirstLetter(breadCrumbsArray[0].replace(/[-_]/g, " ")),
+let secondaryNavChildren = []
+
+if (subNavPages?.length > 0) {
+	secondaryNavChildren = createSecondaryNavBarChildren(subNavPages,currentPage)
```
notlee, line 17: It looks like `createSecondaryNavBarChildren` assumes the label is used in the path. The `nav_label` front-matter has no affect. How can we ensure we can set a `nav_label` and still configure which to show in the sub-nav, and in what order?

apps/astro-website/src/components/header/SecondaryNav.astro
```diff
@@ -1,35 +1,42 @@
...
+} else if (lastChildrenDepth) {
+	if (collectionName) {
+		const documentCollections = await getCollection(collectionName, ({id}) => id.startsWith(subDirectory) )
```
notlee, line 20: In this branch of code we use `subDirectory` to filter a collection. In the other branch we use `collectionFilterCallback`. Is there a good reason to have both options or could these be one?

apps/astro-website/src/components/header/SecondaryNav.astro
```diff
@@ -1,35 +1,42 @@
 ---
-import {capitaliseFirstLetter} from "@/helpers/utils"
+import {getCollection} from "astro:content"
+import {createSecondaryNavBarChildren, filterAndSortNavbars} from "@/helpers/utils"
 import NavItem from "./NavItem.astro"
 
 const currentPage = Astro.url.pathname
 
 const paths = currentPage.split("/").filter(x => x)
 const ancestors = generateAncestors(paths)
```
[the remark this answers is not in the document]
akomiqaia: The short answer is NO. We are generating nav_labels. It seems like we might want to have `ancestors` as part of the API. or find a way to figure out how to get them without passing them down from a parent.

apps/astro-website/src/components/header/SecondaryNav.astro
```diff
@@ -1,51 +1,67 @@
...
+
+async function generateAncestors(paths) {
+	// This assumes that for secondary nav first item will be always the collection name
+	const collectionName = paths[0]
+
+	const urls = [{
+		label: capitaliseFirstLetter(collectionName),
+		url: `/${collectionName}`,
+		"aria-current": currentPage === `/${collectionName}` ? "page" : undefined,
+	}]
```
akomiqaia, line 48: For ancestors unfortunately we are still forced to rely on the Path. The `Documentation` label comes from Primary Nav and the content for it is inside the main header.

akomiqaia: the new update should borrow the label from the primary Navigation definition.

notlee (approved):
Looks good for now, thanks 👍 Let's aim to design away the problem 😄

PR #1213, opened 2023-07-31
backstage: remove ruby related code

akomiqaia:
## Describe your changes
Remove ruby related code


notlee:
Looks like that's pretty much everything! You might also need to update `package-lock.json` to remove it as a workspace and also update `origami.code-workspace` – is that generated?

akomiqaia:
> Looks like that's pretty much everything! You might also need to update `package-lock.json` to remove it as a workspace and also update `origami.code-workspace` – is that generated?

[c6b3fa1](https://github.com/Financial-Times/origami/pull/1213/commits/c6b3fa1ee497bb68b003f4f15ea96ce454672320) updated here. And we do generate it but once we create a new component. otherwise, I don't think there are other generations. BTW this is only for VScode and it seems like none of us are using it. Maybe we don't have to have this file in our code base.

notlee (approved):

PR #1215, opened 2023-08-02
feat: option to keep related content visible in header-services

adgad:
## Describe your changes

By default, any related content in o-header-services shifts into the drawer on small screens.

In spark, we use this space for the New Content button, which we would like to retain visible in the nav on small screens, as it is a critical feature.

This change introduces a new classname on the related content element, which prevents it being shifted into the drawer and keeps it visible at all breakpoints.

It also adds a `keepRelatedContentVisible` prop to the TSX template.

N/A

## Link to Figma designs

N/A


notlee:
Hey, thanks for opening a PR. Looks pretty good and I think it's something I'd be happy for us to merge. Just wondering though how you handle the height change of the nav, or whether you need to? I could see needing to match the media query used by o-header-services to respond to smaller screens – that media query isn't guaranteed to remain the same currently

adgad:
Ahh I see what you mean! I think we need to retain the setting of margin/padding which happens here on large break points - i've added those styles to the keep-visible class also.

I don't think we've had the need for any further styling so far in spark - we have a small o-button and it pretty much just fits. If we did though, presumably the breakpoint would either be `M` (default), or one that we control in the sass options - so I _think_ that should be fine to keep in sync...

components/o-header-services/src/scss/_top.scss
```diff
@@ -92,6 +92,12 @@
 	.o-header-services__related-content {
 		display: none;
 
+		&.o-header-services__related-content--keep-visible {
```
wheresrhys, line 95: is 'keep-visible' too vague a modifier. Keep visible when, how, ...? 

would always-visible be better?

adgad: ooh, yes that's a much better name!

adgad:
The build failures are from `snyk protect` - am assuming this is entirely unrelated to this change 🙈

notlee (approved):

PR #1226, opened 2023-08-10
fix: fatal error on publish

akomiqaia:
## Describe your changes
fixes memory issues when building using Astro by bumping heap

- [x] I have applied `percy` label on my PR before merging and after review.

### DIFF
--- a/scripts/build-website.bash
+++ b/scripts/build-website.bash
@@ -4,6 +4,7 @@
 
 set -e
 cd apps/astro-website/
+export NODE_OPTIONS=--max-old-space-size=10248
 npm run astro build
 cd ../storybook/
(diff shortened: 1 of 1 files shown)


scripts/build-website.bash
```diff
@@ -4,6 +4,7 @@
 
 set -e
 cd apps/astro-website/
+export NODE_OPTIONS=--max-old-space-size=10248
 npm run astro build
```
notlee, line 8: That's a lot of memory! Let's inline so it's only going to apply to astro (?)

```suggestion
NODE_OPTIONS=--max-old-space-size=10248 npm run astro build
```

notlee (approved):

PR #1228, opened 2023-08-10
fix: broken urls for blogposts

akomiqaia:
## Describe your changes
FIx broken links for blog posts

- [x] I have applied `percy` label on my PR before merging and after review.

### DIFF
--- a/apps/astro-website/src/pages/blog/index.astro
+++ b/apps/astro-website/src/pages/blog/index.astro
@@ -29,7 +29,7 @@ const sortedPosts = allPosts.sort(
 					return (
 						<li class="o-layout__listing__item">
 							<h2 class="o-layout__listing__item__title">
-								<a href={"blog/" + generateBlogPostURL(post.slug)}>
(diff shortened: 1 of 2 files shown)


apps/astro-website/src/pages/blog/index.astro
```diff
@@ -29,7 +29,7 @@ const sortedPosts = allPosts.sort(
 					return (
 						<li class="o-layout__listing__item">
 							<h2 class="o-layout__listing__item__title">
-								<a href={"blog/" + generateBlogPostURL(post.slug)}>
```
notlee, line 32: I think this will break heroku and local, but let's roll with that for now

notlee: I'm being silly. What we want here is an absolute path @akomiqaia `"/blog/" +`

notlee: Or do we? Nah we're on the index page, that should be fine

PR #1231, opened 2023-08-14
docs: introduce getting started guide for origami dictionary

frshwtr:
## Describe your changes

Preliminary readme for Origami Dictionary documenting the steps to get set up in order to work with our design tokens.

### DIFF
--- a/apps/dictionary/README.md
+++ b/apps/dictionary/README.md
@@ -21,7 +21,7 @@ The dictionary uses [Tokens Studio](https://tokens.studio/) and [Style Dictionar
 * Ensure you have been added to the Origami team on GitHub.
 * Clone the repository
 ```bash
-git@github.com:Financial-Times/origami.git
+git clone git@github.com:Financial-Times/origami.git
 ```
 * Install the dependencies
 
--- a/apps/dictionary/README.md
+++ b/apps/dictionary/README.md
@@ -8,7 +8,7 @@ The dictionary uses [Tokens Studio](https://tokens.studio/) and [Style Dictionar
 
 ### Developers and Desingers
 
-* Ensure you have been given a Tokens Studio account, the Origami team lead can provide this.
+* Ensure you have been given a Tokens Studio account, the #origami-support Slack channel can help with this.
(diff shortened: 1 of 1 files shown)


apps/dictionary/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+The Origami Dictionary is the repository for storing, managing and publishing Origami Design Tokens.
+
+## Getting Started
+
+The dictionary uses [Tokens Studio](https://tokens.studio/) and [Style Dictionary](https://amzn.github.io/style-dictionary/#/) to manage and publish design tokens.
+
+### Developers and Desingers
+
+* Ensure you have been given a Tokens Studio account, the Origami team lead can provide this.
```
frshwtr, line 11: I think this is actually incorrect 😅, but the lead should have permission to do so, and at least two people should be able to manage this to avoid situations where the only maintainer of the TS account is away.

notlee: I'd recommend asking the Origami Team in the `#origami-support` channel since it requires finding out who the lead is. Someone will be able to help.

notlee:
It might be worth splitting this by task rather than discipline, since there's maybe some overlap? I'm thinking it's possible to create tokens via Github without Token Studio. Or did we find that was maybe not what we want to do because of handling meta data?

notlee (approved):

apps/dictionary/README.md
```diff
@@ -0,0 +1,35 @@
...
+    * Add your dictionary branch `poc/multi-brand`.
+    * Specify a file path where your tokens should be stored `apps/dictionary/tokens`.
+
+### Developers
+
+* Follow the steps for [Developers and Designers](#developers-and-desingers).
+* Ensure you have been added to the Origami team on GitHub.
+* Clone the repository
+```bash
+git@github.com:Financial-Times/origami.git
```
akomiqaia, line 24:
```suggestion
git clone git@github.com:Financial-Times/origami.git
```

akomiqaia (approved):

PR #1246, opened 2023-08-30
fix: o-social-follow, update Twitter branding to X [OR-296]

notlee:
I decided to make this a minor release, without removing Twitter from the API entirely, as SVG is not currently inlined for this component. It's already using the `x` logo from the Image Service update and we only need to replace the twitter blue with black.

### DIFF
--- a/components/o-social-follow/README.md
+++ b/components/o-social-follow/README.md
@@ -47,7 +47,7 @@ The next example shows multiple social media links.
 ```html
 <section class="o-social-follow" aria-label="Follow on social media">
 	<a href="#" class="o-social-follow-icon o-social-follow-icon--twitter">
-		<span class="o-social-follow-icon__label">on twitter</span>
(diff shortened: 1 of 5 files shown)


components/o-social-follow/README.md
```diff
@@ -47,7 +47,7 @@ The next example shows multiple social media links.
 ```html
 <section class="o-social-follow" aria-label="Follow on social media">
 	<a href="#" class="o-social-follow-icon o-social-follow-icon--twitter">
-		<span class="o-social-follow-icon__label">on twitter</span>
+		<span class="o-social-follow-icon__label">on X, formally know as Twitter</span>
```
frshwtr, line 50: Spelling differs to other demo. Think this should be changed to `formerly`

notlee, line 50:
```suggestion
		<span class="o-social-follow-icon__label">on X, formerly know as Twitter</span>
```

PR #1256, opened 2023-08-31
feat: make a "change" to o-buttons

frshwtr:
## Describe your changes

This is a test PR to ensure Percy is triggering as expected. Do not merge.

I will close this shortly once the team has seen this working as expected.


notlee:
Looks like lots of components have changes detected there that I wouldn't expect 🤔 We've hit 75% of our Percy runs until 23rd September 🙈

notlee:
I think perhaps because we're considering devDependencies, and the dependents of devDependencies.

PR #1257, opened 2023-09-04
High fidelity demo

akomiqaia:
## Describe your changes
Create initial High-fidelity demo for o3-button. 

[OR-255](https://financialtimes.atlassian.net/browse/OR-255)

[OR-255]: https://financialtimes.atlassian.net/browse/OR-255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frshwtr:
Percy's failing trying to find `o-something`, not sure why it thinks it's a directory though 🤔

components/o-buttons/src/scss/_brand.scss
```diff
@@ -35,6 +35,7 @@ $_o-buttons-shared-brand-config: (
 );
 
 @if oBrandIs('core') {
+	@debug 'o-buttons: Using core brand config';
```
frshwtr, line 38: These can be removed now. I think these were what you used to generate the token values?

akomiqaia: ups. Seems like I forgot to push my last changes :D And `o-something` was actually a directory but not sure where it came from but removed it now. It has some  `_variables` files in it. most likely we generated at some point and forgot about it.

frshwtr: Aah yeah. The way percy scripts run mean that it's going to still pick up the change. We will need to remove the commit, I think it's committed on `poc/multi-brand` and as we're pushing into it, we can fix it on that branch instead.

apps/dictionary/scripts/build-config/build-component.js
```diff
@@ -16,128 +13,59 @@ StyleDictionaryPackage.registerTransform({
...
-						excludePrefix: ["color-base", "usecase"],
-					},
-				},
-			],
-		},
-	},
-})
+function buildComponentTokens() {
+	const componentName = process.argv[2]
+	// const sources = [`tokens/components/${componentName}/tokens.json`]
```
frshwtr, line 24: This can be remove

components/o3-button/package.json
```diff
@@ -1,39 +1,35 @@
 {
-  "name": "@financial-times/o-buttons-experimental",
+  "name": "@financial-times/o3-button",
   "version": "0.0.0",
-  "description": "o-buttons with design tokens",
+  "description": "Origami buttons with design tokens",
```
frshwtr, line 4: This is true right now, but when we come to introducing more o3 stuff (webcomponents etc..) our description will be different.

components/o3-button/package.json
```diff
@@ -1,39 +1,35 @@
...
   "keywords": [
     "buttons",
     "design-tokens"
   ],
-  "homepage": "https://registry.origami.ft.com/components/o-buttons-experimental",
+  "homepage": "https://registry.origami.ft.com/components/o3-button",
```
frshwtr, line 9: Is this going to exist when we publish? This should point to the storybook docs if not.

frshwtr (approved):

components/o3-button/package.json
akomiqaia: yes definately.

components/o3-button/package.json
akomiqaia: I left it on this since we don't know where it will go. most likely it will be a chromatic URL which we can update later as well. And this kind of thing should be generated by our `create-new-component` CLI tool which will need a serious update once we know how we are writing our components. There is [a ticket](https://financialtimes.atlassian.net/browse/OR-307) to track it as well.

PR #1261, opened 2023-09-06
feat: add show search bar prop to header

ibrahim-mamdouh-ft:
## Describe your changes
We in professional-life-cycle team are conducting an new test for enhanced search and we need the header to have the ability to be shown by default via a prop

In this PR I have added a config object to the Header component in the config I have added a prop **searchBarOpen** to control if the search-bar would be opened by default or not the default behavior for the header is still the same the new feature will only work when **searchBarOpen=true**

I have also added a new prop to o-toggle **isOpened** to determine if the target is open by default or not

[ELES-701](https://financialtimes.atlassian.net/browse/ELES-701)

## Link to Figma designs
https://www.figma.com/file/Hjz28uiQqs8pxUm6X7fCG5/FT.com_Search_Test?node-id=46%3A4083&mode=dev

- [X] I have updated the relevant docs

[ELES-701]: https://financialtimes.atlassian.net/browse/ELES-701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/o-header/README.md
+++ b/components/o-header/README.md
@@ -42,6 +42,14 @@ const headerEl = document.querySelector('.o-header');
 const header = new oHeader(headerEl);
 ```
 
+you can have the search-bar open by default 
+
+```js
+import Header from '@financial-times/o-header';
+const headerEl = document.querySelector('.o-header');
+const header = new oHeader(headerEl, { searchBarOpen: true });
+```
+
 Alternatively, a `o.DOMContentLoaded` event can be dispatched on the document to auto-construct an o-header object for each element with a `data-o-component="o-header"` attribute:
 
 ```js
--- a/components/o-header/README.md
+++ b/components/o-header/README.md
@@ -37,15 +37,17 @@ Some elements inside the header require specific data attributes so the JavaScri
 An o-header object must be constructed for every `<header>` you have on your page that uses this component.
 
 ```js
-import Header from '@financial-times/o-header';
+import oHeader from '@financial-times/o-header';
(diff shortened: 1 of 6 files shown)


components/o-header/README.md
```diff
@@ -42,6 +42,14 @@ const headerEl = document.querySelector('.o-header');
...
 ```
 
+you can have the search-bar open by default 
+
+```js
+import Header from '@financial-times/o-header';
+const headerEl = document.querySelector('.o-header');
+const header = new oHeader(headerEl, { searchBarOpen: true });
+```
+
```
egargan-ft, line 52: I would put this under an `### Options` heading, and word it like this --

> The `oHeader` constructor accepts an optional `options` object, used to control certain behaviours:
> * `searchBarOpen` determines whether the header's search bar is immediately visible. Defaults to `false`.

Then, include your code snippet. This is more in line with their README style.

components/o-header/src/js/header.js
```diff
@@ -6,7 +6,7 @@ import sticky from './sticky.js';
 
 class Header {
 
-	constructor (headerEl) {
+	constructor (headerEl, config = {}) {
```
egargan-ft, line 9: I'd rename this `options`, so it's more obvious that these are 'opt-in' features

egargan-ft (changes requested):

PR #1264, opened 2023-09-06
chore: add hypo demo OR-255

akomiqaia:
## Describe your changes
- update token studio transformers version
- Demonstrate that unique tokens work

We don't have to merge this but same time it should be fine to merge it and then remove the demo code for unconventional button


akomiqaia:
If we are happy with this and we prove that a hypothetical brand can be easily created we could remove code related to the hypothetical brand and merge the rest. This also updates some important bits around token studio and style dictionary.

apps/dictionary/scripts/build-config/build-component.js
```diff
@@ -18,18 +18,34 @@ StyleDictionaryPackage.registerFilter({
...
 
 	const transformers = [
+		"ts/size/px",
 		"ts/descriptionToComment",
 		"ts/typography/css/shorthand",
 		"ts/border/css/shorthand",
 		"ts/shadow/css/shorthand",
 		"ts/color/css/hexrgba",
 		"ts/color/modifiers",
+		"name/cti/kebab",
```
notlee, line 47: This demonstrates how we'd add unique variants well. Let's not merge these tokens, but let's merge the support you've added for assets and composite typography tokens? My only feedback on those aspects is that we're still getting camel case tokens. These should be kebab case to match our other tokens. E.g. `--_o3-button-primary-hypo-typography-fontSize` should be `--_o3-button-primary-hypo-typography-font-size`, inline with naming such as `--o3-font-size-2`

frshwtr: I think we got this working before, I think the ordering of these transformers matters and is preventing the last part from being transformed.

components/o3-button/src/css/core/o3-button/_variables.css
```diff
@@ -1,6 +1,6 @@
 /**
  * Do not edit directly
- * Generated on Mon, 11 Sep 2023 10:00:38 GMT
+ * Generated on Tue, 12 Sep 2023 10:41:23 GMT
```
frshwtr, line 3: This is a note for future work. I'm wondering if this comment needs regenerating every time if there are no changes to a stylesheet. It adds extra changes to a PR which may not be necessary.

frshwtr:
Happy for this to be approved once Lee's suggestion is commited.

apps/dictionary/scripts/build-config/build-component.js
akomiqaia: yeah but then our prefixer was reverting that because in the original code, we are using a token path to join it and build a token name but now I switched to using the token name instead of a path (paths are camelCased while the token names are already separated by '-').

akomiqaia:
> Happy for this to be approved once Lee's suggestion is commited.

Lee's suggestion has been updated with the latest commits.

frshwtr (approved):

PR #1265, opened 2023-09-06
o-header, add option to reveal the search bar by default [OR-311]

notlee:


akomiqaia (approved):
LGTM. and `searchState` sounds good but do you think it might be a bit confusing in the sense that it might mean whole state of a search bar that can include more things than state of openness. Maybe `isSearchBarOpen` better option? but will trust you judgment on this.

notlee:
Thanks for the review @akomiqaia. I think that's a fair comment. I would avoid a question like `isSearchBarOpen`, it's an option that the user is setting not a query of the current state, but it could definitely be more specific – somehow. As we've very unlikely to add other search states, I'm going to go ahead and release this

PR #1267, opened 2023-09-07
O3 tokens library

akomiqaia:
## Describe your changes
Create a prototype of `o3-web-tokens` component


components/o3-button/main.css
```diff
@@ -1,4 +1,4 @@
-@import '../../apps/dictionary/build/css/icons/_variables.css';
+@import '~@financial-times/o3-web-token/icons/_variables.css';
```
notlee, line 1: I wouldn't be surprised if we are asked to brand icons. I _think_ we could add brands later without a breaking change?

components/o3-button/package.json
```diff
@@ -22,7 +22,8 @@
     "watch": "bash ../../scripts/component/watch.bash"
   },
   "peerDependencies": {
-    "@financial-times/o-normalise": "^3.3.0"
+    "@financial-times/o-normalise": "^3.3.0",
+    "@financial-times/o3-web-token": "file:../o3-web-token"
```
notlee, line 26: We'll want a package.json to publish o3-web-token, then we can depend on it here as normal.

components/o3-web-token/internal/_variables.css
```diff
@@ -0,0 +1,46 @@
...
+  --o3-color-palette-oxford-40: #0A3866;
+  --o3-color-palette-oxford-50: #0D4680;
+  --o3-color-palette-oxford-60: #0F5499;
+  --o3-color-palette-oxford-70: #1262B3;
+  --o3-color-palette-oxford-80: #1470CC;
+  --o3-color-palette-oxford-90: #177EE6;
+  --o3-color-palette-oxford-100: #198CFF;
+  --o3-color-palette-teal-20: #052F33;
+  --o3-color-palette-teal-30: #08474D;
+  --o3-color-palette-teal-40: #0A5E66;
```
notlee, line 39: We're missing base and usecase tokens for the internal brand here.

components/o3-web-token/origami.json
```diff
@@ -0,0 +1,27 @@
+{
+	"origamiType": "component",
+	"origamiVersion": "2.0",
```
notlee, line 3: Not as part of this PR: we need to define a 3.0 spec and update the version here

components/o3-web-token/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "scripts": {
+    "build-tokens": "node ../../apps/dictionary/scripts/build-config/build.js"
+  },
+  "engines": {
+    "npm": "^8 || ^9"
+  },
+  "volta": {
+    "node": "20.6.0"
+  },
+  "percy": false,
```
notlee, line 25: !! I didn't know this was a thing?

components/o3-button/package.json
notlee: Oh! We have one. Does it need a non-zero version number perhaps? 0.0.1? `components/o3-web-token/package.json`

components/o3-web-token/whitelabel/_variables.css
```diff
@@ -16,4 +16,7 @@
   --o3-color-palette-black-40: #999;
   --o3-color-palette-black-50: #808080;
   --o3-color-palette-black-60: #666;
```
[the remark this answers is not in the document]
notlee: On usecase tokens we should aim to have the same available across brands.

notlee:
I see a bunch of my comments are handled over here :) https://github.com/Financial-Times/origami/pull/1264/files

components/o3-button/main.css
akomiqaia: yeah. We could always create branded icons inside the icons folder. So you get `@import '~@financial-times/o3-web-token/icons/PROFESSIONAL/_variables.css';` and keep non-branded separate. or we could define icon tokens for brands under brand sets and that should output those tokens as CSS properties in `brand/_bariabled.css` file.

notlee (approved):

PR #1279, opened 2023-09-12
Chromatic setup - OR-289

akomiqaia:
## Describe your changes
- Setup up chromatic and storybook for o3 components
- unrelated to this PR but it removes the file header from style dictionary-generated files.

**NOTE:** At the moment we get chromatic deployments if we push `chromatic-setup` branch once approved workflow file will need to change to `poc-multi-brand`. and once it is ready to go to live we change it to `main`


apps/dictionary/scripts/build-config/utils.js
```diff
@@ -58,9 +66,15 @@ export class ConfigBuilder {
 	 * @param {object} config[].options - options for token file
 	 * @param {boolean} config[].options.outputReferences - output references for token file
 	 * @param {string[]} [config[].options.classNames] - class names for token file
+	 * @param {string} [config[].options.fileHeader] - class names for token file
 	 * */
 	setFilesConfig(config) {
-		this.config.platforms.css.files = config
+		this.config.platforms.css.files = config.map(item => {
```
notlee, line 72: It could be useful to see what a file has rebuilt whilst developing locally, are you sure we want to remove the header completely?

components/o3-button/stories/button.css
```diff
@@ -1,16 +1,9 @@
...
 @import '../internal.css';
 @import '../whitelabel.css';
 @import '../professional.css';
 
-
-
-
 .sub-container {
-  margin: 16px;
+	margin: 16px;
```
notlee, line 7: We should use our own spacing variables! :D

apps/dictionary/scripts/build-config/utils.js
akomiqaia: I just removed the timestamps from file headers since that usually creates the only difference in tokens. I guess we don't want to have file changes detected when tokens actually did not change.

package.json
```diff
@@ -7,7 +7,7 @@
 	"scripts": {
 		"regenerate": "node ./scripts/regenerate-ci-config-files.js",
 		"install": "./scripts/install.bash",
-		"storybook": "cd apps/storybook && npm run storybook",
+		"storybook": "cd apps/o3-storybook && npm run storybook",
```
notlee, line 10: If this ends up in `main` we won't be able to easily run our existing storybook.

components/o3-button/stories/button.css
akomiqaia: This class is not even in use, to be honest. at the moment. I was thinking refactoring the o3-button would be its own thing. While adding all the demos and things related to the button this will get cleared up. Just wanted not to make everything part of this PR and get chromatic out of the way.

package.json
akomiqaia: would a new script `storybook-o3` work better? or maybe the opposite is better (storybook-o2) and make o3 sb default

components/o3-button/stories/button.css
notlee: If this isn't in use we should delete it. It's a small change and is being modified as part of the PR, so I think we're as well to delete now.

package.json
notlee: It would be confusing in the meantime though.

`npm run storybook` 👈 run o3 storybook
`npm run build-storybook` 👈 build o2 storybook

akomiqaia: sure. are you suggesting to `storybook-o3` for now and then switch to default later on?

frshwtr:
There's a lot of `package-lock` changes here, do you know what caused this?

apps/dictionary/scripts/build-config/utils.js
```diff
@@ -2,6 +2,14 @@ import fs from "fs"
 
 export class ConfigBuilder {
 	constructor(StyleDictionaryPackage) {
+		StyleDictionaryPackage.registerFileHeader({
```
frshwtr, line 5: Excellent!

apps/dictionary/scripts/build-config/utils.js
```diff
@@ -2,6 +2,14 @@ import fs from "fs"
...
+			name: "customFileHeader",
+			fileHeader: defaultMessage => {
```
frshwtr, line 7: Small suggestion, we can store this anonymous function in a `const` with a name describing the output of the function. This will help reduce the comments we have throughout our code.

```js
const headerWithoutComments = (defaultMessage) => {
   ...function code...
}

StyleDictionaryPackage.registerFileHeader({
    ...
    fileHeader: headerWithoutComments,
    ...
}

```

components/o3-button/package.json
```diff
@@ -21,16 +21,15 @@
...
     "npm": "^8 || ^9"
   },
   "volta": {
     "node": "18.17.1"
   },
   "percy": false,
-  "private": false
+  "private": false,
+  "peerDependencies": {
+    "@financial-times/o3-web-token": "file:../o3-web-token"
```
frshwtr, line 33: This is ok for now but let's make sure we change this before we publish the package.

package.json
frshwtr: I also think it would make sense to keep the existing scripts pointing to o2 for now. Until o3 becomes mainstream this should stay the case.

> sure. are you suggesting to `storybook-o3` for now and then switch to default later on?

This works. But I'm also happy for O3 storybook's scripts to be ran from the working directory it exists in. Unless there's a reason why the Chromatic deployment would require a full-repo scope.

akomiqaia:
> There's a lot of `package-lock` changes here, do you know what caused this?

@frshwtr All changes are related to the new storybook. and updated some packages. Some of them are updating integrity SHAs.

components/o3-button/package.json
akomiqaia: Yeap we will need to do that as well once published.

package.json
akomiqaia: I reverted it [9fae68a](https://github.com/Financial-Times/origami/pull/1279/commits/9fae68ac63e593435f48b96ff93c70f96303b8dd) so now we will need to run o3 storybook from a directory

notlee (approved):

PR #1284, opened 2023-09-20
backstage: a number of small o3-button fixes

notlee:
This should have be a bunch of smaller commits, but, this commit grew organically and required some exploration of ideas so... sue me.

- Fixes a number of small divergences from our existing o-buttons. Notably fixes the height of "small" buttons, the size of "big" icons, and brings visual font size and weight inline since introducing Metric2 and not re-introducing font smoothing styles (https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/).
- Removes some remaining px values for rem values.


components/o3-button/main.css
```diff
@@ -3,24 +3,32 @@
...
+	--_o3-button-block-padding: var(--o3-spacing-3xs);
+
 	--_o3-button-border-size: 1px;
-	--_o3-icon-size: 20px;
+	--_o3-icon-size: 1.25rem;
+
 	min-width: var(--_o3-button-min-width, 60px);
 	min-height: var(--_o3-button-min-height, 28px);
-	padding: 6px 8px;
+	padding: var(--_o3-button-block-padding) var(--_o3-button-inline-padding);
```
akomiqaia, line 23: what are these variable names trying to convey? padding is applied top/bottom right/left and not sure why the names of vars include block and inline.

components/o3-button/src/tsx/pagination.tsx
```diff
@@ -20,11 +20,8 @@ export interface ButtonPaginationPager {
 	onClick?: any;
 }
 
-const Ellipsis = ({size}: Pick<ButtonProps, 'size'>) => {
+const Ellipsis = ({}: Pick<ButtonProps, 'size'>) => {
```
akomiqaia, line 23: we don't need type definition if we are not passing down any props here

components/o3-button/main.css
notlee: Yeah that's right, I'm deriving the name from [CSS' logical properties](https://css-tricks.com/css-logical-properties-and-values/).

components/o3-button/src/tsx/pagination.tsx
notlee, line 23:
```suggestion
const Ellipsis = () => {
```

components/o3-button/main.css
akomiqaia: I see. Okay that does make more sense now :D

akomiqaia (approved):

PR #1292, opened 2023-10-04
React 18 Node 18 & Storybook 7

benwallman:
## Describe your changes

Quite big update of versions. This was necessary as it's challenging to use newer versions of software with origami. By updating to React 18, we get to take advantage of a bunch of new features.

Used the [storybook codemod](https://www.npmjs.com/package/@storybook/codemod) to update a bunch of the stories into the new format. Would recommend highly for people attempting similar updates.

With this PR, I've removed all our storybook addons, as I can't tell if they're still needed. It would be fantastic to get clarification on their purpose and whether they're still relevant.

https://financialtimes.atlassian.net/jira/software/c/projects/CPREL/boards/1426?selectedIssue=CPREL-772


notlee:
We paired on this and found the issue seemed to be caused by npm generating a bad package-lock file within the PageKit mono repo. Clearing the package-lock file and all node_modules directories (including those in workspaces) seemed to fix the issue. 

Though it would be good to update React / Node / Storybook for the sake of our demos, we need to keep out plugins working, and we don't want to prioritise more work on this right now. This is because we're focused on delivering experimental versions of modernised components, which we will present in a separate Storybook.

PR #1293, opened 2023-10-04
chore: add log on coral stream raising an error

juanSanchezAlcala:
If the initialzation of the client object raise and exception we are not caching it and not logging it. With these changes we track the error in that case.

Why ? 
We do this because a lot of users are reporting coral initialisation errors , we need to collect data to support this reports


notlee:
Hey @juanSanchezAlcala, you may also be interested in [o-errors](https://github.com/Financial-Times/origami/tree/main/libraries/o-errors) to implement this. It's an old library and I have no idea if/how it's used by Customer Products 🙈 So, just a suggestion to take a quick look. I have no problem with a custom tracking event

notlee:
Closing this as it's been open for a while. Please let me know if you would like to revisit this PR.

PR #1302, opened 2023-10-16
chore: release o3-tokkens and o3-buttons

akomiqaia:
## Describe your changes
Release o3-tokens and check ci/cd works for o3 components


components/o3-web-token/package.json
```diff
@@ -14,7 +14,8 @@
   "license": "MIT",
   "type": "module",
   "scripts": {
-    "build-tokens": "node ../../apps/dictionary/scripts/build-config/build.js"
+    "build-tokens": "node ../../apps/dictionary/scripts/build-config/build.js",
```
notlee, line 17: Could this be `build`? Then we don't need to specify this in our Github Actions.

components/o3-web-token/package.json
```diff
@@ -14,7 +14,8 @@
...
+		"lint": "bash ../../scripts/component/lint.bash"
```
notlee, line 18: As discussed with o3-button, let's come back to linting

components/o3-web-token/package.json
notlee: We don't generate a build directory for o3-web-token. How does that affect my import statement as a user? Consistency would be good, to make components easier to use

notlee, line 17: Like you did for 03-button (👏 ) perhaps we could also include `files`, to be explicit about what files should be installed

https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files

components/o3-web-token/.gitignore
```diff
@@ -3,7 +3,6 @@
 /.sass-cache/
 /bower_components/
 /node_modules/
-/build/
```
notlee, line 6: Files are built on deploy, but never committed. Therefore the build files that are committed in this PR will become out of date. I think we should continue to `.gitignore` build files. Is there are reason you removed this?

components/o3-web-token/README.md
```diff
@@ -1,56 +1,35 @@
...
-	- [Contributing](#contributing)
 	- [Migration](#migration)
 	- [Contact](#contact)
 	- [Licence](#licence)
 
-_Instructions for the component. We've broken this down by Markup, JavaScript, Sass, etc. You may add more sections as needed._
 
 ## Usage
 
+Check out [how to include Origami components in your project](https://origami.ft.com/documentation/components/#including-origami-components-in-your-project) to get started with `o3-web-token`.
```
notlee, line 12: This line doesn't apply to o3 – we don't have global o3 tutorials/documentation yet. Let's remove this line (I wrote this comment before but maybe forgot to post ❓ )

components/o3-web-token/.gitignore
akomiqaia: yeah, but these are tokens for brands right? and when we commit tokens from Figma we want to know what changed in the code as well. I was a bit hesitant to put brands in build but for consistency made sense

notlee: Hmmmmmm... good point, I think probably so 👍 We can add automation later to build and commit when token changes come in from Token Studio

notlee (approved):

PR #1303, opened 2023-10-17
backstage: Remove "percy" from component package.json

notlee:
It does not appear to be used by our percy script. We could instead not run the action given o3 component changes.

If there are o2 and o3 changes in place the action will run but as o3 components have no traditional demos, Percy will not actually do anything.


akomiqaia (approved):
LGTM

PR #1308, opened 2023-10-19
chore: impement pnpm

akomiqaia:
## Describe your changes
switch to pnpm


package.json
```diff
@@ -34,16 +34,8 @@
 	},
 	"engines": {
 		"node": "16.13.0",
-		"npm": "8.5.4"
```
frshwtr, line 37: How does this affect Volta, any changes needed on its config?

akomiqaia: https://docs.volta.sh/advanced/pnpm it seems that it is experimental at the moment

frshwtr:
Note pnpm 8 [does not support node 14](https://pnpm.io/installation#compatibility), some of our components will need upgrading

PR #1318, opened 2023-10-25
fix: update release please and origami json for o3-components

akomiqaia:
## Describe your changes
- create release please template
- only build o3 components before pre release
- replace matrix setup for release-please since it did not work


.github/workflows/release-please.yml
```diff
@@ -29,7 +27,7 @@ jobs:
         if: ${{ steps.release.outputs.releases_created }}
       - run: npm ci
         if: ${{ steps.release.outputs.releases_created }}
-      - run: npm run -w ${{ matrix.component }} build --if-present
+      - run: for i in "${o3-COMPONENTS[@]}"; do npm run build -w $i --if-present; done
```
notlee, line 30: This will build all o3 components, even if none are being published. It's not a big deal now but will slow down our releases as the number of o3 components grows. Did you consider using `steps.release.outputs`?

akomiqaia: just replaced some code. I am not sure if there is a way of testing this locally without pushing to NPM.

notlee (approved):

PR #1335, opened 2023-11-03
chore: export o-meter's TSX

frshwtr:
## Describe your changes
Exports o-meters TypeScript templates so that they can be used in external projects.


notlee:
Thanks for the PR @frshwtr! This will need to be a `fix:` commit as we need to trigger a new release, to include the tsx in the npm bundle

components/o-meter/.npmignore
```diff
@@ -8,7 +8,7 @@
 .eslintignore
 .stylelintignore
 stories/
-src/tsx/
+src/
```
akomiqaia, line 11: I believe this will ignore `src` directory for npm and for o2 components we do want to keep `src` to be part of npm package. Since we import tsx from there. (js and css too :D )

akomiqaia (changes requested):

components/o-meter/.npmignore
notlee:
```suggestion
```
Oh yeah, 100%. We need to remove this line entirely

notlee (approved):

components/o-meter/.npmignore
frshwtr: oops

akomiqaia (approved):

PR #1341, opened 2023-11-09
chore: release main

origamiserviceuser:
:robot: I have created a release *beep* *boop*
---

<details><summary>o-teaser: 6.2.10</summary>

### Bug Fixes

* update o-teaser tsx templates to use date-fns version 2 ([#1331](https://github.com/Financial-Times/origami/issues/1331)) ([c53316b](https://github.com/Financial-Times/origami/commit/c53316ba964378d4f9077e2250fc73b40e9d2ab7))
</details>

<details><summary>o3-web-token: 0.2.0</summary>

### Features

* add banker icons ([6fa2ff0](https://github.com/Financial-Times/origami/commit/6fa2ff00e966667dcd4b6c8d0e528c7f9fee3018))
</details>

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #1343, opened 2023-11-13
fix: set o3-button peer dependencies as range

frshwtr:
## Describe your changes

`npm ci` rejects changes on #1341 due to the peer dependency rules of o3-button, these have been update to accept any version greater than and including the current peer dep and below `1.0.0`.

### DIFF
--- a/components/o3-button/package.json
+++ b/components/o3-button/package.json
@@ -34,6 +34,6 @@
 	},
 	"private": false,
 	"peerDependencies": {
-		"@financial-times/o3-web-token": "^0.1.2"
+		"@financial-times/o3-web-token": ">=0.1.2 < 1"
 	}
 }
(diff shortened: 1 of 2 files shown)


notlee:
As there may breaking changes between 0.x.x releases, we need o3-button to depend on the correct version. o3-button is the only package which depends on o3-web-token as far as I know, do you know why we'd see a conflict? Something something mono repo?

frshwtr:
> As there may breaking changes between 0.x.x releases, we need o3-button to depend on the correct version. o3-button is the only package which depends on o3-web-token as far as I know, do you know why we'd see a conflict? Something something mono repo?

I was thinking it was related to [this](https://michaelsoolee.com/npm-pre-1-caret-rules/), it used to be documented in older versions of npm docs but doesn't appear to be any more. But we've bumped a patch here which leaves me stumped.

PR #1344, opened 2023-11-13
chore: added base components for SV

MaryGodservant:


components/o3-button/stories/sustainable-views/button.stories.tsx
```diff
@@ -0,0 +1,37 @@
+import type {Meta} from '@storybook/react';
+import {Button as ButtonTsx} from '../../src/tsx/button';
+import * as ButtonStories from '../story-templates';
+import {Pagination as PaginationTemplate} from '../pagination-template';
+
+import '../../src/css/brands/sustainable-views.css';
+
+export default {
+	title: 'Sustainable Views/o3-button',
+	component: ButtonTsx,
```
frshwtr, line 10: As we use the standard Button typing here, we will end up with options in storybook for themes not supported by Sustainable Views.

frshwtr: We can control the above by using `argTypes` to override. However it seems the SB story template we use assumes all Button stories will support all button variations.

Best action would be to open up the template to some user control via args.

components/o3-web-token/package.json
```diff
@@ -18,7 +18,8 @@
 		"./professional.css": "./build/core/professional/_variables.css",
 		"./icons.css": "./build/icons/_variables.css",
 		"./internal.css": "./build/internal/_variables.css",
-		"./whitelabel.css": "./build/whitelabel/_variables.css"
+		"./whitelabel.css": "./build/whitelabel/_variables.css",
+	    "./sustainable-views.css": "./build/sustainable-views/_variables.css"
```
akomiqaia, line 22: indentation seems to be one tab off? we need to get our prettier fixed 😆

akomiqaia (approved):
LGTM

PR #1346, opened 2023-11-15
backstage: add o2 stories to o3 storybook

notlee:
A mess, to try some things out

- o3 stories within o2, should be separate to avoid building all the sass when working on o2 
- Doesn't build for multiple brands. Hard coded core brand. o-meter story Sass commented out as it doesn't support core. 
- Should we use origami.json and multiple builds, or can we show all o2 in one storybook using a similar approach to o3? May require some story reworking 
- `withDesign` needs removing. It breaks things. We have a design tab now from Chromatic!


notlee:
Closing this now we have https://github.com/Financial-Times/origami/pull/1347

PR #1347, opened 2023-11-16
backstage: migrate o2 storybook to latest version

frshwtr:
## Describe your changes
I know this looks big.. hopefully the commits below explain concisely.

### Outstanding tasks remaining

[ ] check if o-meter is behaving as expected with storybook
Choose between one of the following:
[ ] consolidate all brands into one storybook instance
[ ] ensure storybook is built once for each brand

Without either of the above being comitted, this PR will regress `main`.
 
[or-407](https://financialtimes.atlassian.net/browse/OR-407)


components/o-meter/stories/meter.scss
```diff
@@ -4,5 +4,5 @@
 @include oNormalise();
 @import '@financial-times/o-typography/main';
 @include oTypography();
-@import '@financial-times/o-meter/main';
-@include oMeter();
+//@import '@financial-times/o-meter/main';
+//@include oMeter();
```
notlee, line 8: Did you mean to comment this out? I had this commented out in my proof of concept branch as I hadn't implemented support for the branded o2 storybooks, and o-meter doesn't support the core brand. With this commented out, the internal brand story will have no styles

notlee: Sorry! Just realised this is still a draft 😅

.github/workflows/o2-chromatic.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # 👈 Required to retrieve git history
+      - name: Install dependencies
+        run: npm ci
+        # 👇 Adds Chromatic as a step in the workflow
+      - name: Publish Origami to Chromatic
+        uses: chromaui/action@v1
```
notlee, line 34: this is very nice

apps/o2-storybook/readme.md
```diff
@@ -0,0 +1,51 @@
...
+```bash
+ORIGAMI_STORYBOOK_BRAND=internal npm run storybook -w apps/o2-storybook
+```
+
+Origami 2 supports these brands:
+* `core`
+* `internal`
+* `whitelabel`
+
+## Creating stories and structure
```
notlee, line 30: 🎉 great to see these docs improve

notlee: Oh actually! The first bit is great, but I think the docs copied from o3-storybook are a lie. We don't structure stories for brands in the same way. Ideally we would, but it would be an annoying amount of work. Can we update the docs here to describe how we handle o2 branded components?

notlee (approved):

PR #1348, opened 2023-11-16
feat: release o-testing-library

akomiqaia:
## Describe your changes

We had some issues with testing after installing new modules for o3 components. We tried to have a different approach to fix issues with the old component setup but this solution, compiling problematic modules for web-test-runner, seemed easiest so far and it seems like it resolves problems with testing.

### DIFF
--- a/libraries/o-testing-library/README.md
+++ b/libraries/o-testing-library/README.md
@@ -1,0 +1,38 @@
+# o-testing-library [![MIT licensed](https://img.shields.io/badge/license-MIT-blue.svg)](#licence)
+
+Origami testing helper.
+
+- [o-testing-library ](#o-testing-library-)
+	- [Usage](#usage)
+	- [Migration Guide](#migration-guide)
+	- [Contact](#contact)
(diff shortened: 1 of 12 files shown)


libraries/o-testing-library/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+- [o-testing-library ](#o-testing-library-)
+	- [Usage](#usage)
+	- [Migration Guide](#migration-guide)
+	- [Contact](#contact)
+	- [Licence](#licence)
+
+## Usage
+
+This module is bundling the [DOM testing library](https://testing-library.com/docs/dom-testing-library) and [user-event](https://testing-library.com/docs/ecosystem-user-event/) modules. It is intended to be used in Origami components and applications to provide a consistent way of testing. Now you can import both modules from the same place with the following import:
```
notlee, line 13: It would be helpful to give more context here. That this is to avoid compilation complexity whilst using web-test-runner to test o2 components. This information may provide helpful context for people in the future (maybe us, maybe not)

PR #1350, opened 2023-11-20
chore: update last release sha

akomiqaia:
## Describe your changes
We accidentally merge bad PR and now it's trying to publish minor changes for some components. Not to do this we are manually setting good PRs SHA.


notlee:
This won't work, as release please will look at any commit after this commit, which includes your mistaken commit. I think setting this to the bad commit sha should do it though, assuming it isn't inclusive

notlee (approved):

PR #1362, opened 2023-11-30
fix: o3-button, enforce set of allowed icons in tsx [OR-406]

notlee:
This enables Storybook to apply icons via query param, allowing stories to be shared or embedded with an icon selected.

It also avoids custom class names starting with `o3-button-` which confuses where the class is defined.

For custom icons, users may apply a css custom property via a parent element or through the `attributes` TSX prop.

https://financialtimes.atlassian.net/browse/OR-406


components/o3-button/src/tsx/button.tsx
```diff
@@ -67,10 +66,14 @@ export function Button({
 	attributes = {},
 	onClick,
 }: ButtonProps) {
+	// Combine custom classes with first party o3-button classes,
+	// rather than override them.
+	const customClasses = attributes.className;
+	delete attributes.className;
```
akomiqaia, line 72: Why don't we directly pass `attributes.className` to the function? and could be part of another PR :D

components/o3-button/src/tsx/button.tsx
```diff
@@ -16,8 +16,6 @@ export interface ButtonProps {
 		| "search"
 		| "refresh"
 		| "cross"
-		| ""
-		| (string & Record<never, never>); // Support IDE autocomplete whilst allowing any string https://github.com/microsoft/TypeScript/issues/29729#issuecomment-1331857805
 	iconOnly?: boolean;
 	visuallyHideDisabled?: boolean;
```
akomiqaia, line 20: We have a letter missing here :D

akomiqaia (commented):
I will be happy to approve this but few things to discuss maybe:
-  we should really start using TS properly and define types for all the functions as well - - Also are we happy how this shows up as controls in Storybook? I like the idea of it being part of attributes object but maybe not that useful to have in controls?
- and i know we don't have this yet but should we document what attributes do we expect for o3-button? or document that this is the way we add new icon classes?

frshwtr:
I have tested this locally with a query string from our figma/storybook links and can confirm this works.

LGTM, but I'll let Ako add the approval once his comments are resolved.

components/o3-button/src/tsx/button.tsx
notlee: The argument we pass to `makeClassNames` is an object so that would need to be written something like `{customClasses: attributes.className}`, which is much the same.

I think as a user I would expect any `attributes` I define to be passed straight through to the html element. Setting `[data-custom-classes]` then having that apply to `className` could be quite confusing, when `[data-some-other-thing]` would apply as a data attribute.

akomiqaia (approved):
LGTM

PR #1378, opened 2023-12-19
backstage: introduce prettier write pre-commit

frshwtr:
## Describe your changes
We already have prettier installed, but it appears it runs optionally depending on local developer configuration.

This PR proposes an approach where prettier formats and writes files staged for commit before comitting happens, ensuring any changed files now follow our prettier rules.


akomiqaia (approved):
Something we needed for long time :D It would be good to figure out how to do this locally as well. I think I tried to configure this for my local but did not seem to work.

frshwtr:
> Something we needed for long time :D It would be good to figure out how to do this locally as well. I think I tried to configure this for my local but did not seem to work.

PR #1384, opened 2024-01-03
fix: show pre-selected options in the select box and use option label…

kitkat119:
… for aria-label

Now that a value can be provided, that should be used when formatting the selectedOptionsValues rather than the label, otherwise the pre-selected options are not displayed correctly.

## Describe your changes
Following on from [this PR](https://github.com/Financial-Times/origami/pull/1379), we found that any pre-selected options that were providing a value as well as a label, were not being displayed correctly in the select box. 

This change fixes that by using the option's value, with the label as a fallback, when formatting the selectedOptionsValues before they are passed to the select element. 

It also fixes the aria-label which was showing `[object object]`.


akomiqaia (approved):
LGTM

PR #1388, opened 2024-01-04
fix: Use textContent rather than innerText

kitkat119:
## Describe your changes

I've been writing tests using react-testing-library for my usecase of the multiselect but have struggled because the option's innerText was undefined. Apparently innerText is not implemented in jsdom so this PR changes to use `textContent` instead.


notlee (approved):
Hey 👋 I'm happy to approve this as it shouldn't change any functionality in this case. However, they're not equivalent, and it would be reasonable to use `innerText` or other web API which depends on CSS layout. I'd definitely recommend taking a look at how you test components, to test them in a real browser!

PR #1396, opened 2024-01-16
Simplify brand selector logic in for-everyone site

egargan-ft:
## Describe your changes

This PR simplifies the logic involved in the 'brand selector' in the new for-everyone site. It also removes some unneeded and unused files.

In this approach, we simply override Starlight's `ContentPanel`  component (the container component for the central area of the page), and add some logic that applies `o3-brand-professional`, `o3-brand-core`, etc to the container element based on `Astro.props.lang` when the page is rendered.

`Astro.props.lang` will be one of `en-GB-x-prof`, `en-GB-x-core` etc according to the current URL path (e.g. `/professional` etc `/`). This is handled by Starlight's i18n features, which have already been set up in the site.

The `LanguageSelector` component has also been simplified. The logic that dynamically manages brand classes has been removed, and the path-setting logic has been tidied.

https://financialtimes.atlassian.net/browse/OR-482

- [x] I have updated relevant env variables in Doppler.


apps/for-everyone-website/astro.config.mjs
```diff
@@ -30,9 +29,8 @@ export default defineConfig({
 			},
 			components: {
 				Hero: './src/components/Hero.astro',
-				ThemeProvider: './src/components/ThemeProvider.astro',
 				LanguageSelect: './src/components/LanguageSelect.astro',
-				ThemeSelect: './src/components/ThemeSelect.astro',
+				ContentPanel: './src/components/ContentPanel.astro',
```
notlee, line 33: Could we hide the theme select? We're using Origami's internal brand styling more and more for the site, but we don't support dark mode styles, so the site looks a bit borked with dark mode enabled. 

(Design tokens will make dark mode support easier in the future if we want to!)

notlee (commented):
This is much much better, and actually consistently shows the correct brand :D One little thing

apps/for-everyone-website/src/components/ContentPanel.astro
```diff
@@ -0,0 +1,47 @@
+---
+import type {Props} from "../props"
```
frshwtr, line 2: These props imports give errors in my IDE. `../props` does not exist, I think this code was in the Starlight default components which these custom components were inspired from. I don't know if there's any Starlight magic thats somehow correctly importing the props.

Could we remove it if it's not being used?

notlee (approved):

PR #1400, opened 2024-01-18
chore: search and build all o3 libraries

frshwtr:
## Describe your changes

O3 Dependencies for the for-everyone website are linked locally due to how npm workspaces resolves packages installed that come from the same monorepo. As we don't build and commit the exported elements in O3, they require building so that for-everyone can find them.

Currently we do this in the `build` script in the `package.json`, however this list needs maintaining as new O3 components come into place, and could get very long.

This PR introduces a suggested approach, where Bash is used to find any `o3-` prefixed components and run their build script before running astro's build step in for-everyone.

This means we don't need to define them. However has some caveats.

### Caveats

* It assumes every package prefixed with `o3-` has a build script, and that all will be used in for-everyone, resulting in unecessary packages being build.
* In current draft state, there is no error handling, e.g should trying to run build on a package where the script doesn't exist just ignore this particular package and continue?


notlee:
>It assumes every package prefixed with o3- has a build script, and that all will be used in for-everyone, resulting in necessary packages being build.

That's probably okay, all our components should be documented on the site. We might want to reintroduce `origami.json` or set custom keys in `package.json` with Origami information but I think this is pragmatic right now.

>In current draft state, there is no error handling, e.g should trying to run build on a package where the script doesn't exist just ignore this particular package and continue?

We can use the `--if-present` to ignore where there is no build script. If the script errors I think bailing out is reasonable, we should know about that and fix it.

–

Another issue to consider is that running `dev` may also catch us out. We'll always need to remember to run `build` first. An alternate approach might be to commit a component's build? However we might want to automate that, to ensure we're working with an up to date build – such as getting a bot to commit the build on release.

notlee:
@frshwtr Possibly, though long pre-commit tasks can get annoying, and building only changed components could get complex? I do like the idea of reviewing the built output though.

notlee (approved):
Happy to approve as an improvement, though the developer experience needs more work I think

PR #1401, opened 2024-01-18
Display o3 spacing tokens on the for everyone site. 

notlee:
Though the content is the same so far, the token values and descriptions for spacing could vary per brand. 

I had to create a spacing mdx file for each brand to correctly pass the current brand to `Spacing.astro`, which includes the branded tokens. This duplications is not really maintainable, as even creating the files and consistently updating the front-matter for every brand is a pain. 

It feels like we aught to be able to pass the brand as `Astro.props.locale` to mdx content, but I'm not sure how without further investigation – [here's where the magic happens](https://github.com/withastro/starlight/blob/ce01a51da775b968ed7227c81dc87e56bc090cac/packages/starlight/index.astro#L17). It might be further evidence that we will we need to move away from the Starlight/i18n approach?


frshwtr (approved):
LGTM!

PR #1403, opened 2024-01-22
Add 'internal' brand to for-everyone site

egargan-ft:
> [!NOTE]  
> This PR is a draft pending an exploration into a better and less-duplicate-y approach to supporting brands in the site

## Describe your changes

This PR adds the 'internal' brand to the for-everyone website, making it available as an option in the brand picker.

The steps for adding a new brand are as follows:

* Add a new 'locale' to `astro.config.mjs` for the new brand
* Tell the `ContentPanel` component which CSS file to load when this language/brand is set as active
* Copy all components over from an existing brand, switching out the brand-specific CSS file import for that component, and updating the Figma and Storybook links
* Copy other documentation mdx files that contain brand-specific content or styles that have to be manually set

👆 _Of these, that last step could absolutely be improved!_

### DIFF
--- a/apps/dictionary/tokens/internal/base/color.json
+++ b/apps/dictionary/tokens/internal/base/color.json
@@ -4,47 +4,91 @@
       "palette": {
         "black": {
           "type": "color",
-          "value": "#000"
+          "value": "#000",
+          "origamiKeys": [
+            "palette",
+            "primary"
+          ]
         },
         "white": {
           "value": "#fff",
-          "type": "color"
+          "type": "color",
+          "origamiKeys": [
+            "palette",
+            "primary"
+          ]
         },
         "oxford": {
           "value": "#0F5499",
-          "type": "color"
+          "type": "color",
+          "origamiKeys": [
+            "palette",
+            "primary"
+          ]
         },
         "teal": {
           "value": "#0D7680",
-          "type": "color"
+          "type": "color",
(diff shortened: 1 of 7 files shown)


frshwtr:
> Copy all components over from an existing brand, switching out the brand-specific CSS file import for that component, and updating the Figma and Storybook links

This looks like it's going to become a lot to maintain once we add more components. Is there a way to use the same copy when it's identical across brands? I recall we can default load some with i18n.

notlee (approved):

notlee:
> > Copy all components over from an existing brand, switching out the brand-specific CSS file import for that component, and updating the Figma and Storybook links
> 
> This looks like it's going to become a lot to maintain once we add more components. Is there a way to use the same copy when it's identical across brands? I recall we can default load some with i18n.

Core content is used as a fallback when there isn't brand specific content. This falls down when:
1. We want to include Figma/Storybook links to the current brand.
2. Brands include differing themes we want to document, like Professional doesn't have a mono theme since its default looks like mono.

We can, and will need to, improve on this. Abstracting or otherwise reworking the `Preview` component will help with that. Otherwise we'll likely need to take on more significant work to build on astro directly, without starlight. That's what Ed was exploring and left some notes for us https://financialtimes.slack.com/archives/CSW6B2VAN/p1706293983163469

PR #1411, opened 2024-01-26
Spacing support matrix

notlee:


frshwtr (approved):

apps/for-everyone-website/src/content-partial/docs/guides/spacing-implementers-note-origami.mdx
```diff
@@ -0,0 +1,33 @@
...
+<table>
+    <thead>
+        <tr>
+            <th width="50%">o2 (traditional)</th>
+            <th width="50%">o3 (new)</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td>Aligned: Our latest spacing scale has additional options, and different naming. See our [deprecated UI library](https://www.figma.com/file/MyHQ1qdwYyek5IBdhEEaND/FT-UI-Library?type=design&node-id=0-2169&mode=design&t=WXTU1v4lXmcRPKaO-0). </td>
```
wheresrhys, line 12: Given that aligned is a word that is used when discussing the layout of a design, and that it is also used here to mean 'compatible with current design system thinking'... perhaps a different choice of word would be better as the dual meaning could be confusing (I'm not the target audience, but I do find the terminology used in this PR a bit unclear)

notlee: How about "Compatible"?

notlee: I'm going to merge this as this language is used on other pages, but am happy to change across the board

PR #1416, opened 2024-01-30
docs: display o-autocomplete demos in O2 Storybook

frshwtr:
## Describe your changes

Surfaces o-autocomplete demos in Storybook.

Future registry/o2-storybook migrations may be automated. But this PR was done manually and discovered the following areas to pay particular attention to when migrating demos:
* Comments appear in HTML/templating syntax, and are not compatible with JSX, these will need transforming.
* Some tags that should be self-closing are not. Which throws up errors when compiling JSX.

### DIFF
--- a/components/o-autocomplete/stories/dynamic-complex.stories.tsx
+++ b/components/o-autocomplete/stories/dynamic-complex.stories.tsx
@@ -1,0 +1,121 @@
+import {StoryFn} from '@storybook/react';
+import '../demos/src/shared.scss';
+import oForms from '@financial-times/o-forms';
+import {useEffect} from 'react';
+
+import Autocomplete from '../main.js';
+import {data} from './data.js';
+oForms.init();
+
+export default {
+	title: 'Components/o-autocomplete',
+	parameters: {
+		html: {},
+	},
+	argTypes: {},
+};
+
+/**
+ * @typedef {object} CustomOption
+ * @property {string} Continent_Code - 2 letter continent code
+ * @property {string} Continent_Name - name of continent
+ * @property {string} Country_Name - name of country
+ * @property {number} Country_Number - id of country
+ * @property {string} Three_Letter_Country_Code - three letter country code
+ * @property {string} Two_Letter_Country_Code - two letter country code
+ */
+
+/**
+ * @param {CustomOption} option - The option to transform into a suggestion string
+ * @returns {string} The string to display in the suggestions dropdown for this option
+ */
+function mapOptionToSuggestedValue(option) {
+	if (typeof option !== 'object') {
+		throw new Error(
+			`Could not map option to suggested value, unexpected type: ${typeof option}.`
+		);
+	}
+
+	if (typeof option.Country_Name !== 'string') {
+		throw new Error(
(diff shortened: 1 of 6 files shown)


notlee:
@frshwtr Could Prettier help with both issues? Fun fact, we [run it on component demos](https://github.com/Financial-Times/origami-build-service/blob/a16de21fc6760b9717a74498c8c24f8c17882de6/lib/middleware/v3/outputDemo.js#L24) before returning via the Build Service.

components/o-autocomplete/stories/dynamic-complex.stories.tsx
```diff
@@ -0,0 +1,121 @@
+import {StoryFn} from '@storybook/react';
+import '../demos/src/shared.scss';
+import oForms from '@financial-times/o-forms';
+import {useEffect} from 'react';
+
+import Autocomplete from '../main.js';
+import {data} from './data.js';
+oForms.init();
```
notlee, line 8: Importing dependencies and re-creating demos is much nicer, if we can automate this that's great. Have you considered a less nice solution of committing already compiled JS for demos? We don't expect to develop o2 components much if at all. Perhaps it could save you time to focus on the transition to o3?

components/o-autocomplete/stories/dynamic-complex.stories.tsx
```diff
@@ -0,0 +1,121 @@
+import {StoryFn} from '@storybook/react';
+import '../demos/src/shared.scss';
+import oForms from '@financial-times/o-forms';
+import {useEffect} from 'react';
+
+import Autocomplete from '../main.js';
+import {data} from './data.js';
```
notlee, line 7: Looks like this file doesn't exist

notlee: At least when I run 

`npm run storybook -w apps/o2-storybook/`

I see

`Module not found: Error: Can't resolve './data.js'` 🤔

notlee (commented):
Working pretty nicely!

- The delayed demo doesn't appear to be delayed?
- Should these PRs include moving the README over, as we have for other components with stories?
- Some food for thought on the automation plan above.

components/o-autocomplete/stories/dynamic-complex.stories.tsx
frshwtr: @notlee Happy to try this on the other demos to migrate. Shall we keep this change for autocomplete?

frshwtr:
> Working pretty nicely!
> 
>     * The delayed demo doesn't appear to be delayed?

What is the expected delayed behaviour? I see a delay in suggestions with a spinner when typing.

>     * Should these PRs include moving the README over, as we have for other components with stories?

Yes, this should be the case. Pushed a commit for this.

components/o-autocomplete/stories/dynamic-complex.stories.tsx
frshwtr: @notlee This will involve us needing to either commit all built legacy demos or build them with storybook's build. Are these also acceptable given the legacy nature?

notlee (approved):

PR #1419, opened 2024-02-02
fix: pagination improvements

notlee:
- Fix pages displayed on mobile when there are between 5-7 pages.
- Remove small pagination option. We do not recommend this.
- Use legacy o-grid values for media queries, for now.
- Correct specificity of selected buttons.

before/after
Capture-2024-02-02-112634
Capture-2024-02-02-112233


akomiqaia (approved):
LGTM

PR #1422, opened 2024-02-02
docs: remove sass-css monitoring post

notlee:
focus more on steps we would like teams to take


notlee:
Oh, I removed `o-syntax-highlight` because it's clashing with Astro styles. This was the quickest solution to readable syntax highlighting – it is a problem on other pages too

akomiqaia (approved):
We can always use [astro's syntax highlighting](https://docs.astro.build/en/guides/markdown-content/#syntax-highlighting)

PR #1427, opened 2024-02-05
chore: updating ui typography tokens

MaryGodservant:


components/o3-web-token/build/core/_variables.css
```diff
@@ -80,13 +80,14 @@
...
   --o3-spacing-4xl: 6rem;
-  --o3-font-family-metric: 1;
-  --o3-font-family-financier-display: 2;
-  --o3-font-weight-light: Light;
-  --o3-font-weight-regular: Regular;
-  --o3-font-weight-medium: Medium;
-  --o3-font-weight-semibold: Semibold;
-  --o3-font-weight-bold: Bold;
+  --o3-font-family-metric: metric 2 VF;
+  --o3-font-family-financier-display: financier display VF;
```
frshwtr, line 84: These seem erroneous, I think these should be strings?

akomiqaia (approved):
We might need to update o3-button with new typography tokens? or otherwise we could have a new Jira ticket that deals with it

PR #1428, opened 2024-02-07
backstage: build tsx templates

akomiqaia:
- configure ts inside the component root that outputs necessary files for the o3-button
- define a more generic ts configuration at the root for tsx transpilation

### DIFF
--- a/components/o3-button/package.json
+++ b/components/o3-button/package.json
@@ -27,7 +27,8 @@
 	},
 	"types": "./build/types/index.d.ts",
 	"scripts": {
-		"build": "bash ../../scripts/component/build-o3.bash o3-button"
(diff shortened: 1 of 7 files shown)


components/o3-button/src/tsx/pagination.tsx
```diff
@@ -1,37 +1,19 @@
...
-	onClick?: any;
-}
+// @ts-nocheck
 
-type Ellipsis = Pick<ButtonProps,'theme'> & {
-	attributes?: {
-		[attribute: string]: string | boolean;
-	};
-}
+//  TODO: remove @ts-nocheck and fix types
```
frshwtr, line 3: Is this to be done as part of this PR?

frshwtr (approved):

PR #1448, opened 2024-02-20
chore: added editorial typography tokens to the core brand

MaryGodservant:


notlee:
I've asked Mary about moving these to the component layer. I think this is important because:
1. Editorial typography won't be supported for all brands. I guess we're used to selectively supporting components, but there's no way to know from the whitelabel brand which use-case tokens are required and which aren't when adding new brands.
1. Usecases are published in our o3-web-tokens package (o3-foundations). This package will be shared by all brands engineering side, and we don't want to force users to update when we change editorial type if it's not relevant for their brand (e..g internal)

frshwtr (approved):

PR #1453, opened 2024-02-22
Add `data-o3-brand` and `data-o3-theme` selector convention [OR-572]

notlee:
See commit messages! And [Slack](https://financialtimes.slack.com/archives/CSW6B2VAN/p1708532615067549).

### DIFF
--- a/apps/for-everyone-website/src/components/ContentPanel.astro
+++ b/apps/for-everyone-website/src/components/ContentPanel.astro
@@ -1,26 +1,23 @@
 ---
-function getBrandClassFromLocale(lang: string) {
+function getBrandFromLocale(lang: string) {
 	debugger
 	switch (lang) {
 		case "en-GB-x-prof":
-			return "o3-brand-professional"
+			return "professional"
 		case "en-GB-x-sv":
-			return "o3-brand-sustainable-views"
+			return "sustainable-views"
 		case "en-GB-x-internal":
-			return "o3-brand-internal"
+			return "internal"
 		case "en-GB-x-core":
 		default:
-			return "o3-brand-core"
+			return "core"
 	}
 }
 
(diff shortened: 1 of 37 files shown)


libraries/o3-tooling-token/build/sustainable-views/_variables.js
```diff
@@ -728,5 +728,485 @@ export default {
 		],
 		"css": "--o3-color-palette-claret",
 		"figma": "o3/color/palette/claret"
+},
```
notlee, line 731: All the extra lines appears to be because I built tokens, and we forgot to since adding new SV tokens

notlee: o3-button CSS is actually a few lines shorter now :)

apps/for-everyone-website/src/components/ContentPanel.astro
```diff
@@ -45,3 +42,10 @@ let brandClass = getBrandClassFromLocale(Astro.props.lang)
 		}
 	}
 </style>
+
+<script>
```
notlee, line 46: Moved this from [Hero.astro](https://github.com/Financial-Times/origami/pull/1453/files#diff-15961c9cd376c48b9c0a1db6055783de3dfc3e65cfc5d041880ef2384b092311) – it's still not a good place, but is better, and co-located with other brand logic

apps/for-everyone-website/src/components/PaginationAnatomy.astro
```diff
@@ -1,5 +1,5 @@
 ---
-import {ButtonPagination} from '@financial-times/o3-button';
+import {ButtonPagination} from '@financial-times/o3-button/jsx';
```
notlee, line 2: Oops. Missed these imports in my last PR, the for-everyone build is failing

PR #1470, opened 2024-03-01
docs: fetch iframed stories from build service

frshwtr:
## Describe your changes

Some stories that were migrated from registry would display an `iframe` in the HTML tag. For stories that do this, the HTML preview will fetch from the iframed source and display that instead.


apps/o2-storybook-composition/addons/html/src/withHTML.js
```diff
@@ -0,0 +1,50 @@
+import {makeDecorator, useChannel} from '@storybook/addons';
+import {EVENTS} from './constants';
+
+export const withHTML = makeDecorator({
+	name: 'withHTML',
+	parameterName: 'html',
+	skipIfNoParametersOrOptions: false,
+	wrapper: (storyFn, context, {parameters = {}}) => {
+		console.log('wrapped');
```
notlee, line 9:
```suggestion
```

apps/o2-storybook-composition/addons/html/src/Panel.js
```diff
@@ -0,0 +1,77 @@
...
+	}); // https://storybook.js.org/docs/react/addons/addons-api#usechannel
+
+	useChannel({
+		[EVENTS.CODE_UPDATE]: async ({code}) => {
+			const parser = new DOMParser();
+			const codeDoc = parser.parseFromString(code, 'text/html');
+
+			const foundIframe = codeDoc.getElementsByTagName('iframe');
+
+			console.log('a');
```
notlee, line 23:
```suggestion
```

apps/o2-storybook-composition/addons/html/src/Panel.js
```diff
@@ -0,0 +1,77 @@
...
+		code: null,
+		options: {},
+	}); // https://storybook.js.org/docs/react/addons/addons-api#usechannel
+
+	useChannel({
+		[EVENTS.CODE_UPDATE]: async ({code}) => {
+			const parser = new DOMParser();
+			const codeDoc = parser.parseFromString(code, 'text/html');
+
+			const foundIframe = codeDoc.getElementsByTagName('iframe');
```
notlee, line 21: This assumes there's no iframe within a demo right? I'm not 100% sure no demos include an iframe, could we confirm that or make this selector more specific in some way?

apps/o2-storybook-composition/addons/html/src/Panel.js
```diff
@@ -0,0 +1,77 @@
...
+						.src.replace('/demo', '/demo/html');
+					const html = await fetch(demoEndpoint);
+
+					if (html.status !== 200) {
+						throw new Error(`${html.status}: ${html.text()}`);
+					}
+					const json = await html.text();
+					code = json;
+				} catch (e) {
+					console.log('Failed to load load Iframe story from source: \n' + e);
```
notlee, line 42: Is this an opportunity to make the error visible to the user, perhaps via the html tab? With an invitation to contact the team for support?

notlee (approved):
Looks like it works well 😄 Some suggestions here. I also think we should apply this to `apps/o2-storybook` as well, as I think there are scenarios people will view the non-composed storybook. What do you think? Copy-pasting seems fine given the support status of o2

notlee (changes requested):
See previous review comments 😅

notlee (approved):

PR #1476, opened 2024-03-07
backstage: improve typing

akomiqaia:
## Describe your changes
- Remove sub pathing for JSX components and make it the default export
- Use the browser field for bundled mjs code since it only contains client code.

[OR-503](https://financialtimes.atlassian.net/browse/OR-503)

[OR-503]: https://financialtimes.atlassian.net/browse/OR-503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tsconfig.json
```diff
@@ -5,9 +5,9 @@
 		"jsx": "react-jsx",
 		"esModuleInterop": true,
 		"skipLibCheck": true,
-		"target": "ES2021",
+		"target": "ES2019",
```
notlee, line 8: Why the drop on our target?

apps/for-everyone-website/src/components/ToggletipAnatomy.astro
```diff
@@ -1,5 +1,5 @@
 ---
-import { TooltipToggle } from '@financial-times/o3-tooltip/jsx';
```
notlee, line 2: The site was broken locally so I updated the import. We'll need to bump the versions used within the for-everyone package.json within the release branch though as well. Else when it's deployed, it'll install the current version before this change and fail

components/o3-tooltip/package.json
```diff
@@ -16,30 +16,27 @@
...
+	"module": "./build/jsx/index.mjs",
 	"files": [
 		"build",
 		"origami.json"
 	],
 	"exports": {
 		".": {
-			"types": "./build/types/index.d.ts",
-			"import": "./build/js/index.js"
+			"browser": "./build/js/index.mjs",
```
notlee, line 29: I had to add a conditional export `browser` here for it to work within our Astro site. I'm a little nervous about how other bundlers will handle this – we'll see 🤞

tsconfig.json
notlee: Yeah looks like we're good with 2021 :) 
https://caniuse.com/?feats=mdn-javascript_builtins_string_replaceall,mdn-javascript_builtins_promise_any,mdn-javascript_builtins_weakref,mdn-javascript_operators_logical_or_assignment,mdn-javascript_operators_logical_and_assignment,mdn-javascript_operators_nullish_coalescing_assignment,mdn-javascript_grammar_numeric_separators,mdn-javascript_builtins_finalizationregistry#summary

notlee (approved):

tsconfig.json
akomiqaia: [e3784de](https://github.com/Financial-Times/origami/pull/1476/commits/e3784de58173a457d9f1e09fb427ae27897f9330) this udpates it. Also moved the browser after types in `package.json`. I remember that types should always be first. Or maybe that was an issue in some cases but don't remember the reason now.

PR #1510, opened 2024-03-26
feat: add post modifier to teaser

juanSanchezAlcala:
add post modifier , meant to be used in live blog posts.
At the moment affects the styles if the teaser has the layout small

This PR change the way to solve [CLS](https://web.dev/articles/cls) by using aspect ratio instead of padding property as it was using

[ticket](https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1653?selectedIssue=CI-2167)

image

- [x] I have updated relevant env variables in Doppler.

### DIFF
--- a/components/o-teaser/README.md
+++ b/components/o-teaser/README.md
@@ -130,6 +130,7 @@ Additional modifiers:
 - `stacked`: moves the image to the top of the teaser
 - `opinion`: changes the tag colour to blue
 - `live`: change background to red and position of elements to make the teaser stand out
+- `post`: set liveblog post styles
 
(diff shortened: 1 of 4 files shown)


AliceArmstrong (approved):
👍
Text looks slightly wrong, but I'm guessing Storybook just isn't using the production fonts?

akomiqaia (approved):
LGTM. it seems like linting is not correct but that should be an easy fix. Maybe applying Percy label is not a bad idea as well to check the visual differences between old and new version

umbobabo (approved):

PR #1515, opened 2024-03-27
Fix css exports

akomiqaia:
## Describe your changes
- remove exports keyword from package json
- instead generate cjs,esm folders
- fix imports
- update readmes


components/o3-button/README.md
```diff
@@ -251,16 +251,16 @@ To make a button disabled add the `disabled` attribute. To visually hide the dis
 
 ## JSX
 
-For React users, Origami components now include JSX templates. Import JSX templates along with component CSS.
+For React users, Origami components now include JSX templates. Import JSX templates along with component CSS. JSX templates are exported as `cjs` (common JS) and `esm` (ECMAScript) modules, so depending on your system configuration, you may need to import the correct module type.
```
frshwtr, line 254: This feels specific to all o3 components, which would end up being duplicated in all O3 readmes, we should look to extract the common documentation to a general readme for all o3 components. Individual components can then signpost to this readme.

I'm OK for this to go in though, I'll create a ticket to create a general readme.

PR #1516, opened 2024-03-27
feat: Add query parameter to custom suggestionTemplate function

andygout:
## Describe your changes
This PR makes changes which extend on those introduced by https://github.com/Financial-Times/origami/pull/1127 and enables highlighting of custom rendering of suggestions.

If the user does not provide a `suggestionTemplate` in the options when instantiating an instance of `oAutocomplete` then the `Autocomplete.suggestionTemplate` will be used:

https://github.com/Financial-Times/origami/blob/f11860c20b67a7584d8b04f1948ff54dcb90d68d/components/o-autocomplete/src/js/autocomplete.js#L387-L413

This has a direct reference to `autocompleteEl` (via `this.autocompleteEl`), which can be used to acquire the current text input value, which is then used to enable highlighting of the suggestion.

However, when providing a custom `suggestionTemplate` function, this direct reference to `autocompleteEl` is not an option. It is possible to obtain the text input via `const input = document.getElementById('autocomplete-input');`, though to do this prior to rendering every suggestion is inefficient.

The changes made in PR https://github.com/Financial-Times/accessible-autocomplete/pull/29 means that `query` (the text which was typed into the autocomplete text input field by the user) can be included in a custom `suggestionTemplate` callback function's parameters and so used to highlight the `option` value (or, in the scenario of `option` being an object, a value thereof).

This screengrab demonstrates the benefit of these changes via a scenario in which a custom `suggestionTemplate` function was used (in order to include the suffix that describes the type of thing the suggestion is: person, venue, company, etc.) and where the suggestion's name is also highlighted in accordance with the text the user has input:

autocomplete-custom-highlighted-suggestion

**Open question:** `Autocomplete.suggestionTemplate` can continue to acquire the `query` value via `const input = this.autocompleteEl.querySelector('input');` (and then acquire the value from `input.value`); to remove that line and provide the `query` in the same way as a custom `suggestionTemplate` function (as a second function parameter) results in linting error (see [here](https://github.com/Financial-Times/origami/actions/runs/8459515926/job/23175973753?pr=1516)): `Error:   396:21  error  Expected 'this' to be used by class method 'suggestionTemplate'  class-methods-use-this`. This implies the class method would need to be changed into a standalone function which in turn entails a fair amount of overhauling perhaps excessive for the intention of this work, though I'd be interested to hear opinions.

I've added @adgad as a reviewer on the basis that they may be familiar with this code from their work on https://github.com/Financial-Times/origami/pull/1127.

N/A

## Link to Figma designs
N/A

## TODO (pre-merge)
- [x] Upgrade @financial-times/accessible-autocomplete version to that which includes changes included in PR https://github.com/Financial-Times/accessible-autocomplete/pull/29
- [x] Identify if the [Storybook demos](https://main--655f72ec522e424302dc6201.chromatic.com/?path=/story/o2-core_components-o-autocomplete--dynamic-delayed&globals=backgrounds:!undefined) for `dynamic-custom-highlighted-suggestion` (and `dynamic-custom-suggestion`) should have their `hidden` value set to `true`
  - **Update:** The intention is to keep things in Storybook as what people get from Origami out of the box, so the demos whose `hidden` value is set to `true` should remain as such

## Dependency PRs
- https://github.com/Financial-Times/accessible-autocomplete/pull/29

## Release
This adds functionality in a backwards-compatible manner so I propose to release these changes as a **minor** version.

- [x] I have updated relevant env variables in Doppler. — N/A

### DIFF
--- a/components/o-autocomplete/README.md
+++ b/components/o-autocomplete/README.md
@@ -236,9 +236,13 @@ This function is used to override the default rendering of suggestion items, wit
 
 It is typically used when wanting to provide additional context or styling for each suggestion item.
 
+The `query` value (text which was typed into the autocomplete text input field by the user) is provided so that it can be used as a basis for highlighting the `option` value (or one of its values if it is an object).
+
 :warning: **Caution:** because this function allows you to output arbitrary HTML, you should [make sure it's trusted](https://en.wikipedia.org/wiki/Cross-site_scripting), and accessible. The HTML will be output within listbox options, so [ensure all descendants are presentational](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/option_role#all_descendants_are_presentational).
 
 #### Example
+
+Providing additional context by displaying multiple `option` properties:
 
 ```js
 import oAutocomplete from 'o-autocomplete';
@@ -250,6 +254,7 @@ async function customSuggestions(query, populateOptions) {
 
 /**
  * @param {{"name": string, "role": string}} option - The option to transform into a suggestion
+ * @param {string} [query] - Text which was typed into the autocomplete text input field by the user
  * @returns {string} The HTML to render in the suggestion list*/
 function suggestionTemplate(option) {
 	return `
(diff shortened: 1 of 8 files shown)


notlee (approved):
:shipit:

PR #1526, opened 2024-04-03
fix: fix token collision error

akomiqaia:
also implements integration tests for style dictionary (OR-616)[https://financialtimes.atlassian.net/browse/OR-616]


apps/dictionary/__test__/integration.test.js
```diff
@@ -0,0 +1,128 @@
...
+		const iconsOutput = readFileSync(
+			pathJoin(__dirname, buildPath, 'icons.css'),
+			{encoding: 'utf-8'}
+		);
+		it('should build CSS variables for icon tokens', () => {
+			expect(iconsOutput).toMatchSnapshot();
+		});
+	});
+
+	// foundation
```
frshwtr, line 34: What extra value do these comments add? They seem to repeat information in the `describe` message.

apps/dictionary/__test__/tokens/core/base/spacing.json
```diff
@@ -0,0 +1,54 @@
+{
```
frshwtr, line 1: We shouldn't need this many token sets for these tests. Could we reduce the example down to perhaps a few files demonstrating each of the different kinds of tokens we have? (e.g Base, use case, component)

apps/dictionary/__test__/integration.test.js
```diff
@@ -0,0 +1,128 @@
...
+import {
+	buildIconCSS,
+	buildBrandForCSS,
+	buildComponentTokens,
+	buildToolingMetaTokens,
+	buildToolingIconTokens,
+} from '../scripts/build-config/buildFunctions';
+
+describe('buildCss()', () => {
+	const destinationPrefix = '__test__/build/css';
```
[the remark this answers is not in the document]
akomiqaia: yeah but we have `it.each` and then we have to move the building script inside as well. I chose not to build for each brand and instead remove build folder at the beginning of the test.

apps/dictionary/scripts/build-config/utils.js
```diff
@@ -136,12 +124,27 @@ StyleDictionaryPackage.registerTransform({
...
+	'ts/typography/css/shorthand',
+	'ts/typography/css/fontFamily',
+	'ts/border/css/shorthand',
+	'ts/shadow/css/shorthand',
+	'ts/color/css/hexrgba',
+	'ts/color/modifiers',
+	'name/cti/kebab',
+	'name/origamiPrivatePrefix',
+];
+
```
frshwtr, line 147: This breaks the flow of the two JSDoc comments, reordering will keep the two JSDoc comments nearby for relevance.

```suggestion
const transformers = [
	'Origami/pxToRem',
	'ts/size/px',
	'ts/size/lineheight',
	'ts/descriptionToComment',
	'ts/typography/css/shorthand',
	'ts/typography/css/fontFamily',
	'ts/border/css/shorthand',
	'ts/shadow/css/shorthand',
	'ts/color/css/hexrgba',
	'ts/color/modifiers',
	'name/cti/kebab',
	'name/origamiPrivatePrefix',
];

 * @property {string} destination - The output file path.
 * @property {TokenFilter|undefined} [tokenFilter] - A function to filter tokens to include.
 * @property {string|undefined} [parentSelector] - A parent CSS selector for generated CSS.
 */
```

frshwtr, line 147: This array breaks the flow of the two JSDoc comments, reordering will keep the two JSDoc comments nearby for relevance.

apps/dictionary/scripts/build-config/utils.js
```diff
@@ -246,7 +227,7 @@ export function buildMeta({sources, destination}) {
 							) {
 								return false;
 							}
-							return isTokenStudioSource(token);
+							return true;
```
frshwtr, line 230: Why is this check removed?

apps/dictionary/scripts/build-config/utils.js
```diff
@@ -283,63 +262,45 @@ export function getBasePath() {
...
-					isSetBelongsToSubBrand(theme, tokenSet)
-				);
-			});
-
-			return componentTokenSets.map(tokenSet =>
-				path.join(basePath, `tokens/${tokenSet}.json`)
-			);
-		});
+/**
+ * Find all token sources for a specific brand.
```
frshwtr, line 274: This is a very useful comment! In the future I think it would be useful for us as a team to put information like the characteristics of sub brand naming in a doc for contributing to our O3 space.

apps/dictionary/scripts/build-config/utils.js
```diff
@@ -283,63 +262,45 @@ export function getBasePath() {
...
+ * If the brand name a `/` separated string, it is a sub-brand.
+ * Sub-brands will include the parent brand's tokens.
+ * Include tokens will be overridden by "source" attribute tokens.
+ *
+ * @param {string} brand
+ * @returns {{sources: string[], includes: string[]}} - The source tokens and include tokens for the brand.
+ */
+export function getBrandSourcesAndIncludes(brand) {
+	const theme = getTokenStudioThemes().filter(
```
frshwtr, line 283: `Array.find()` can be used here instead of `Array.filter()` to get the first item matching the testing function. We don't need to access `[0]` with this approach.

apps/dictionary/scripts/build-config/utils.js
akomiqaia: since we are building tokens for Astro Metadata here this filter will result in empty tokens. Astro metadata needs tokens that are mostly foundation tokens or non-source tokens. This logic was left behind accidentally when code got updated at some point I believe.

PR #1528, opened 2024-04-03
fix:  do not affect nested expander elements

juanSanchezAlcala:
Expander was affecting to toggler buttons inside nested o-expander elements.
With this fix we exclude elements that are nested in other expander element.

I've added an example in storybook so we can test that scenario. Let me know if this is not needed

[ticket](https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1653?selectedIssue=CI-1976)

### DIFF
--- a/components/o-expander/src/js/expander-utility.js
+++ b/components/o-expander/src/js/expander-utility.js
@@ -112,6 +112,8 @@ class ExpanderUtility {
 		// Elements.
 		this.oExpanderElement = oExpanderElement;
 		this.contentElement = this.oExpanderElement.querySelector(this.options.selectors.content);
+		//Add data-o-component=o-expander in case it doesn't have it 
+		this.oExpanderElement.setAttribute('data-o-component','o-expander')
 		this.toggles = [].slice.apply(this.oExpanderElement.querySelectorAll(this.options.selectors.toggle)).filter((item) => 
 			//Do not get nested elements
 			item.closest('[data-o-component="o-expander"]') === this.oExpanderElement
--- a/components/o-expander/src/js/expander-utility.js
+++ b/components/o-expander/src/js/expander-utility.js
@@ -112,7 +112,10 @@ class ExpanderUtility {
 		// Elements.
 		this.oExpanderElement = oExpanderElement;
 		this.contentElement = this.oExpanderElement.querySelector(this.options.selectors.content);
-		this.toggles = [].slice.apply(this.oExpanderElement.querySelectorAll(this.options.selectors.toggle));
+		this.toggles = [].slice.apply(this.oExpanderElement.querySelectorAll(this.options.selectors.toggle)).filter((item) => 
+			//Do not get nested elements
+			item.closest('[data-o-component="o-expander"]') === this.oExpanderElement
+		);
 		if (!this.toggles.length) {
 			throw new Error(
 				'o-expander needs a toggle link or button. ' +
--- a/components/o-expander/src/js/expander-utility.js
+++ b/components/o-expander/src/js/expander-utility.js
@@ -31,6 +31,11 @@ class ExpanderUtility {
 			throw new Error('Expected an expander Element.');
 		}
 
+		//Do not initialised if it was already initiated
+		if(oExpanderElement?.oExpander?.initialized){
+			return;
+		}
+
 		// Error if no options are given.
 		if (typeof opts !== 'object') {
 			throw new Error(`Expected an \`opts\` object, found type of "${typeof opts}".`);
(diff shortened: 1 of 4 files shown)


components/o-expander/stories/nested-expander.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+					<li>thirteen</li>
+				</ul>
+			</Expander>
+			</div>
+		</Expander>
+	);
+};
+
+export default {
+	title: 'Components/o-expander/nested',
```
notlee, line 34:
```suggestion
	title: 'Components/o-expander',
```

I don't think we need to group by `nested` here
Screenshot 2024-04-03 at 16 14 10

notlee (changes requested):
Looks reasonable! I'm happy to approve when tests are passing, are you happy to take a look at those?

juanSanchezAlcala:
> Looks reasonable! I'm happy to approve when tests are passing, are you happy to take a look at those?

Hi @notlee , maybe im wrong but i think [this test](https://github.com/Financial-Times/origami/blob/7391d124e3dd06385926451652e35fd989b50ce1/components/o-expander/test/custom.test.js#L83) isn't ok, since the title says that is testing shrinkTo with a number but the test sets "height" instead.

juanSanchezAlcala:
@notlee I've found the reason of the tests failing. In the tests some fixtures doesn't include data-o-component=o-expander . Im adding it on the code in case the expander element doesn't have it when initialising.

components/o-expander/src/js/expander-utility.js
```diff
@@ -112,7 +112,12 @@ class ExpanderUtility {
 		// Elements.
 		this.oExpanderElement = oExpanderElement;
 		this.contentElement = this.oExpanderElement.querySelector(this.options.selectors.content);
-		this.toggles = [].slice.apply(this.oExpanderElement.querySelectorAll(this.options.selectors.toggle));
+		//Add data-o-component=o-expander in case it doesn't have it 
+		this.oExpanderElement.setAttribute('data-o-component','o-expander')
+		this.toggles = [].slice.apply(this.oExpanderElement.querySelectorAll(this.options.selectors.toggle)).filter((item) => 
+			//Do not get nested elements
+			item.closest('[data-o-component="o-expander"]') === this.oExpanderElement
```
notlee, line 119: Good find on the issue with the failing tests but I think this could lead to unexpected behaviour. Instead of setting `data-o-component` could you check for nested expanders by checking for the presence of the [initalized class](https://github.com/Financial-Times/origami/blob/e971320c09decb8e2d020f373320bca2f942394b/components/o-expander/src/js/expander.js#L29)?

The reason I think `data-o-component` could be a problem is because it's usually used to automatically initialise components by calling the `init` method, with no arguments, or by including the `o-autoinit` component. There's a small risk here that a project or its dependency tries to initialise all expanders after this constructor is called to manually initalise an expander without `data-o-component`, and therefore attempts to construct the same expander twice.

juanSanchezAlcala: Use o-expander--intialized class is a problem because we get the toggle buttons on the initialisation and therefore that class is added at the end of the process , so we couldn't use it before.
In case some component has been initialised before it shouldn't be initialised twice checking if it has the initialized className. This is not implemented yet. If you think its ok that solution I'll add the needed code.

juanSanchezAlcala:
Not related with this PR but just in case its not right,
 Is this [objectAssign](https://github.com/Financial-Times/origami/blob/e971320c09decb8e2d020f373320bca2f942394b/components/o-expander/src/js/expander.js#L22) right ?  shouldn't be the other way around ?  
I mean the options passed to the object should override the default options, right ? 
At the moment the default is being assign to the options passed so it will override passed options

components/o-expander/src/js/expander-utility.js
notlee: Good point. I think checking for the "intialized" class class to avoid initialising twice works 👍 Our new components are standard Custom Elements, so this shouldn't be an issue going forward 🙌

notlee:
@juanSanchezAlcala yes that does look like it should be the other way around, weird it hasn't come up before, I guess no one is customising them? Wonder if we have tests against that?

juanSanchezAlcala:
> @juanSanchezAlcala yes that does look like it should be the other way around, weird it hasn't come up before, I guess no one is customising them? Wonder if we have tests against that?

If i switch the objectAssing arguments and run the test they passed correctly, though i wouldn't change this without a breaking change, since if there is some code passing options to oExpander it will change the logic and potentially could change the behaviour and cause an issue.

components/o-expander/src/js/expander-utility.js
juanSanchezAlcala: I have changed a bit the way to check if its initialised. 
And also i set init method to return always the same instance even it was already initialised, take a look and let me know if you think its ok

notlee (approved):

PR #1534, opened 2024-04-04
chore: update svg tokens

akomiqaia:


notlee:
I think there might be something wrong with the cross icon svg? 

https://github.com/Financial-Times/origami/assets/10405691/c65705db-3f9b-40d9-8030-b496df054550

FYI I've rebased on main and rebuilt tokens locally

notlee (approved):
👍

PR #1536, opened 2024-04-05
backstage: Build components when building storybook.

notlee:
o3-storybook is failing to deploy to chromatic
https://github.com/Financial-Times/origami/actions/runs/8557812028/job/23450955531

```
Error: Can't resolve '@financial-times/o3-foundation/css/internal.css' in '/home/runner/work/origami/origami/components/o3-tooltip/src/css/brands'
```

I think this is because we're using a mono-repo, and dependencies are resolved locally. However components have a build step, and the built files aren't commited. So we need to build them.


akomiqaia (approved):
LGTM

PR #1540, opened 2024-04-08
O3 Button | Big button support, custom font-weight

ob6160:
## What does this change?

As part of migrating https://github.com/Financial-Times/pro-central-banking to `o3-button` in https://github.com/Financial-Times/pro-central-banking/pull/137 I needed to port across the existing `*--big` modifier for buttons and allow the setting of custom font-weights and padding.

- Sets up stories for `small` buttons in Storybook
- Allow customisation of `font-weight`
- Set `font-weight` to `700`
- Change `font-size` to `1rem`


notlee (approved):
Thanks :D

PR #1543, opened 2024-04-09
Correct o3 font and button font size

notlee:
See commit messages for details. The below screenshot shows o3-button vs o-buttons

Screenshot 2024-04-09 at 14 30 49

Horizontal padding differences are a conscious design choice

### DIFF
--- a/components/o3-foundation/fonts.css
+++ b/components/o3-foundation/fonts.css
@@ -2,8 +2,8 @@
 @font-face {
 	src: url('https://www.ft.com/__origami/service/build/v3/font?version=1.13&font_name=Metric2-VF&system_code=origami&font_format=woff2')
 		format('woff2');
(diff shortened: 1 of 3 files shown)


components/o3-foundation/fonts.css
```diff
@@ -2,8 +2,8 @@
 @font-face {
 	src: url('https://www.ft.com/__origami/service/build/v3/font?version=1.13&font_name=Metric2-VF&system_code=origami&font_format=woff2')
 		format('woff2');
-	font-family: var(--o3-font-family-metric);
-	font-weight: 400;
+	font-family: 'metric 2 VF';
```
notlee, line 5: We can't used CSS Custom Properties in `@font-face`.
We want Metric2 to support more than 400 in weight.

ob6160 (approved):
nice one 👾

PR #1549, opened 2024-04-11
backstage: fix storybook dev run

notlee:


akomiqaia (approved):
LGTM but we should apply chromatic label to this to be sure that deployment works

PR #1564, opened 2024-04-16
Set aria-modal according to `opts.modal` (o-overlay)

egargan-ft:
## Describe your changes

[As MDN explains](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-modal), non-modal dialogs (which o-overlay creates when the 'modal' option is set to false) shouldn't have `aria-modal=true`.

- [ ] I have applied `percy` label for o-overlay or `chromatic` label for o3-overlay on my PR before merging and after review.


notlee (approved):
🙇

PR #1566, opened 2024-04-16
feat: Add support for fluid buttons [OR-629]

notlee:
Please excuse Prettier.

Apply the `data-o3-fluid` data attribute to make buttons "fluid", and expand to the full width of their container. By default this applies only to small, mobile viewports. On larger viewports a fluid button's width is based on its content by default.
I chose to use a data attribute over a class so we can expand this to be fluid from different breakpoints if we need in the future e.g. `data-o3-fluid="until-medium"` and so it can potentially apply to multiple components within a parent

```html
<button class="o3-button o3-button--primary" data-o3-fluid>Click me</button>
```

For more control on when a button should be full width set the `--o3-button-fluid` CSS Custom Property. For example:

Force buttons to always be full width, given they are a direct child of a "my-button-container" element.

```css
.my-button-container > .o3-button {
	--o3-button-fluid: initial;
}
```

Make buttons full width, given they are a direct child of a "my-button-container" element, and "my-button-container" is less than `500px` wide.

```css
.my-button-container {
	container-type: inline-size;
}

@supports (container-type: inline-size) {
	@container (width < 500px) {
		.my-button-container > .o3-button {
			--o3-button-fluid: initial;
		}
	}
}
```


components/o3-button/main.css
```diff
@@ -17,6 +17,9 @@
 	--_o3-button-font-weight: var(--o3-font-weight-semibold);
 
 	min-width: var(--_o3-button-min-width);
+	/* Width based on content by default, until set to `inital` */
+	--o3-button-fluid: ;
```
frshwtr, line 21: Interesting 👀

frshwtr (approved):

components/o3-button/main.css
ob6160: is this defining the variable but as an `undefined` value? 🤔

notlee: https://css-tricks.com/the-css-custom-property-toggle-trick/

components/o3-button/main.css
```diff
@@ -288,6 +291,13 @@
 	}
 }
 
+/* FLUID BUTTON */
+
+.o3-button[data-o3-fluid] {
+	@media screen and (max-width: 490px) {
+		--o3-button-fluid: initial;
+	}
```
ob6160, line 299: Screenshot 2024-04-17 at 10 46 26
Screenshot 2024-04-17 at 10 46 17

I was just checking when we go full width in our codebase, and we seem to have two patterns: 
- 100% until we hit `M`
- 100% below `S`

I think we should standardise on one so hopefully the `490px` BP here will do the trick for all cases we have

ob6160 (approved):
⏺️

PR #1582, opened 2024-04-18
chore: use latest chromatic in CI

akomiqaia:


.github/workflows/chromatic.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: npm ci
         # 👇 Adds Chromatic as a step in the workflow
       - name: Publish to Chromatic
-        uses: chromaui/action@v1
+        uses: chromaui/action@latest
```
notlee, line 34: Can we pin to a version to avoid unexpected breakages?

akomiqaia:
Just realised that merging `release-please` will trigger the deployment to Chromatic because `release-please` modifies `package.json`. But we need to trigger a build on `package.json` in case some packages got updated.

notlee (approved):

PR #1590, opened 2024-04-23
docs: Add responsive fluid button prompt

notlee:


akomiqaia (approved):

notlee:
Looks good. It makes our html more complex but I think the trade off is appropriate. It supports [inverse themes](https://github.com/Financial-Times/origami/blob/76e55baba9b8353c54387796698b54fe15df2fb5/components/o3-foundation/focus.css#L94) and avoids repeating quite a lot of CSS given we support `:focus-visible` and a `:focus` fallback 👍

PR #1596, opened 2024-04-24
fix: fix issue with for toggletip when used by screen readers

akomiqaia:
## Describe your changes
The toggle tip did not announce the content of the toggle tip. This issue was related to the fact that elements that have `role=status` only get announced if the content changes. In this PR we are simulating content change so screen readers have a chance to announce content of the tooltip.

[OR-657](https://financialtimes.atlassian.net/browse/OR-657)

[OR-657]: https://financialtimes.atlassian.net/browse/OR-657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/o3-tooltip/README.md
```diff
@@ -51,16 +51,19 @@ Below are examples of how to use `<o3-tooltip-onboarding>` and `<o3-tooltip-togg
...
+  class="o3-tooltip"
+  content="click the button to see the tooltip"
+  title="Title"
+  info-label="information button"
+ >
+ </o3-tooltip-toggle> 
 </div>
 ```
 
+`info-label` attribute is used to provide a label for info icon to have better screen reader experience. By default, it is set to "information button".
```
notlee, line 65: The code example splits `content-id` and `info-label` property descriptions. Would it be more helpful to have properties and their purpose listed together either before or after the example?

notlee, line 65: "information button" will be read to screen reader users as something like "information button, button". I would suggest "more information" as a default value. We could use the `info-label` description as an opportunity to recommend users set a more descriptive label E.g. "More about [topic]"

components/o3-tooltip/README.md
```diff
@@ -133,11 +139,12 @@ The `ToggleTooltip` JSX element accepts the following `props`:
...
-| :-------: | :---------: | :------: | :-----: | :----------------------: |
-|  content  |   string    |   true   |    -    |  Content of the tooltip  |
-|   title   |   string    |  false   |    -    |   Title of the tooltip   |
-| placement | `Placement` |  false   |  'top'  | Placement of the tooltip |
+|   Prop    |    type     | required |      default       |       description        |
+| :-------: | :---------: | :------: | :----------------: | :----------------------: |
+|  content  |   string    |   true   |         -          |  Content of the tooltip  |
+|   title   |   string    |  false   |         -          |   Title of the tooltip   |
+| placement | `Placement` |  false   |       'top'        | Placement of the tooltip |
+| infoLabel |   string    |   true   | information button | Label for screen readers |
```
notlee, line 147: Could we link to the more detailed description provided above?

components/o3-tooltip/src/ts/toggleTooltip.ts
```diff
@@ -3,13 +3,27 @@ import {ToggleToolTipProps} from '../types';
...
+		this.infoLabel = 'information button';
+	}
 
 	connectedCallback() {
 		super.connectedCallback();
-		this.innerHTML = this._generateMarkup(this.title, this.content);
+		if (this.getAttribute('info-label')) {
+			this.infoLabel = this.getAttribute('info-label') as string;
+		}
+		console.log('connectedCallback', this.infoLabel);
```
notlee, line 19:
```suggestion
```

components/o3-tooltip/src/ts/toggleTooltip.ts
```diff
@@ -46,16 +62,34 @@ export class ToggleToolTip extends ToolTip implements ToggleToolTipProps {
 		if (!isChild && !isTarget) {
 			this._contentWrapper.style.display = 'none';
 			this.removeAttribute('open');
+			this._removeContentInLiveRegion();
```
notlee, line 65: These 3 lines are repeated on escape, perhaps there's an opportunity to abstract in a function?

components/o3-tooltip/src/types/index.ts
```diff
@@ -21,7 +21,9 @@ export interface OnboardingToolTipProps extends Omit<TooltipProps, 'placement'>
 		| 'left'
 		| 'left-end';
 }
-export interface ToggleToolTipProps extends TooltipProps {}
+export interface ToggleToolTipProps extends TooltipProps {
+	infoLabel: string;
```
notlee, line 25: This is a breaking change as it's now required and could break someone's build. Do we make it optional, as we provide a default. Or should we not provide a default, force the user to set a more descriptive label, and make a major release?

notlee:
I just tested this in VoiceOver and it doesn't seem to work? 🤔

components/o3-tooltip/src/types/index.ts
akomiqaia: goog point. I think we should make it optional and having `more information` as default is already pretty good i think

components/o3-tooltip/README.md
akomiqaia: I guess I put them after their respective components. felt more like there was a better readme flow. :D but can group together if you think that is better

components/o3-tooltip/README.md
akomiqaia: what do you mean by `link to` as in to actually link to lines that describe what `infoLabel` does or more to reflect changes made

notlee (approved):
It works!

PR #1601, opened 2024-04-25
ADSDEV-1832 - fix: remove elements directly in o-video rather than el.removeChild(c…

sami-ft:
## Describe your changes
Replaces instances where we remove an element by using remove() directly, rather than el.removeChild(childEl). 

It relies on less assumptions about which node is a parent of which other ones and fixes an issue on the adErrorHandler where a removeEl statement would fail because the node we're trying to remove isn't a descendant of the node on which removeChild is called 

https://financialtimes.atlassian.net/browse/ADSDEV-1832

## Link to Figma designs
N/A

### DIFF
--- a/components/o-video/src/js/ads.js
+++ b/components/o-video/src/js/ads.js
@@ -191,7 +191,7 @@ class VideoAds {
 
 		// Remove the preloading spinner
 		if (this.loadingStateEl) {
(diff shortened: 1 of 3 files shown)


notlee:
We live in the future now. By IE

notlee (approved):

PR #1606, opened 2024-04-25
docs: origami for nobody

notlee:
We aim to move origami-for-everyone.ft.com (now origami-beta.ft.com) to origami.ft.com, moving away from Origami For Everyone. That project name means different things to different people and is no longer required.


akomiqaia (approved):
LGTM

Issue #1609, opened 2024-04-26
o-tooltip's imperative interface could accept an HTMLElement node


aendra-rininsland:
One difficulty I often have with o-tooltip is that it requires an ID string for the target, which are often in short supply around React-land. Looking at the [code](https://github.com/Financial-Times/origami/blob/main/components/o-tooltip/src/js/tooltip.js#L34), the imperative interface only uses the opt.target value to run a `getElementById` to then pass on the returned node to the constructor.

Allowing opts.target to be an `HTMLElement` allows for patterns like the following (in React pseudocode):

```
import React, { useEffect } from 'react';
import OTooltip from '@financial-times/o-tooltip';

const Tooltip = ({target, ...opts}) => {
    const tooltip = useRef(null);
    const el =  useRef(null);

    useEffect(() => {
        tooltip.current = new OTooltip(el.current, {target, ...opts});
    }, []);

    return <div ref={el} />
}
```

...Where the `target` prop is likely another `ref` passed down from a parent element:

```
const ButtonWithTooltip = (props) => {
  const ref = useRef(null);
  return <>
    <button ref={ref}>Some content to tooltip</button>
    <Tooltip target={ref} {...props} />
  </>
}
```

PR #1611, opened 2024-04-26
feat: o-tooltip opts.target can be HTMLElement

aendra-rininsland:
## Describe your changes

This PR allows o-tooltip to receive a HTMLElement instead of an ID string, allowing for the usage patterns identified in the linked issue.

#1609 

### DIFF
--- a/components/o-tooltip/src/js/tooltip.js
+++ b/components/o-tooltip/src/js/tooltip.js
@@ -123,7 +123,7 @@ class Tooltip {
 			Tooltip.throwError("tooltip.target is not set. An target for the tooltip to point at must be provided");
 		}
 
-		if (!typeof opts.target === 'string' || !opts.target instanceof HTMLElement) {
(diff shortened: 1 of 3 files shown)


components/o-tooltip/src/js/tooltip.js
```diff
@@ -123,6 +123,10 @@ class Tooltip {
 			Tooltip.throwError("tooltip.target is not set. An target for the tooltip to point at must be provided");
 		}
 
+		if (!typeof opts.target === 'string' || !opts.target instanceof HTMLElement) {
```
notlee, line 126: I think we want this to be `&&` over `||`? Not string and not element. 

```suggestion
		if (typeof opts.target !== 'string' && !(opts.target instanceof HTMLElement)) {
```

notlee: Ah no. Looks like the test you added was passing. This does look easier to read to my eye though, and avoids the linting error

PR #1622, opened 2024-04-30
docs: add migration to o3-typography to o-typography

frshwtr:


components/o-typography/MIGRATION.md
```diff
@@ -1,5 +1,119 @@
 ## Migration guide
 
+### Migrating from v7 to o3-typography
+
+#### Fonts
+
+Fonts are now part of [o3-foundation](https://github.com/Financial-Times/origami/blob/main/components/o3-foundation/README.md). Please refer to the o3-foundation documentation for more information.
```
notlee, line 7:
```suggestion
```

This is quite vague and leaves it to the user to figure out what's changed. Could we be more specific about what they need to do?

We need to document a migration guide for Sass. This commentary might be more relevant there. As an example, what should the user do if they see a reference to `$o-typography-load-fonts` in their project?

components/o-typography/MIGRATION.md
```diff
@@ -1,5 +1,119 @@
...
+
+### Themes
+
+Brands are now applied using a parent data attribute data-o3-brand. Likewise, themes are applied using data-o3-theme – this may be placed on the on a parent element or the o3-typography component directly.
```
notlee, line 11:
```suggestion
Brands are now applied using a parent data attribute `data-o3-brand`. Likewise, themes are applied using `data-o3-theme` – this may be placed on the on a parent element or the `o3-typography` component directly.
```

components/o-typography/MIGRATION.md
```diff
@@ -1,5 +1,119 @@
...
+E.g. An FT Core brand Link with inverse theming, brand and theme inherited from a parent element.
+```html
+<body data-o3-brand="professional" data-o3-theme="inverse">
+	<a href="#" class="o3-typography o3-typography-link">To somewhere.</a>
+</body>
+```
+
+### Mixins
+
+#### oTypography
```
notlee, line 29: There's [a whole bunch of Sass missing here](https://o2.origami.ft.com/?path=/docs/o2-core_components-o-typography-sassdoc--docs&globals=backgrounds:!undefined#otypography).

components/o-typography/MIGRATION.md
```diff
@@ -1,5 +1,119 @@
...
+
+Update your markup to use the `o3-typography` JSX template ([o3-button JSX documentation](https://main--64faf6b1815b6c0106f82e74.chromatic.com/?path=/docs/core-o3-typography)). Alternatively, replace `o-typography` class prefixes with `o3-button` as described below.
```
notlee, line 31: This implies I can find and replace `o-typography` for `o3-typography` which isn't the case. Could this be rewritten to indicate the markup changes are documented below?

components/o-typography/MIGRATION.md
```diff
@@ -1,5 +1,119 @@
...
+
+##### Body elements
```
notlee, line 33: This section lives under "Mixins" but also applies to Build Service users who don't use Sass. Could we move this to a markup section?

notlee: If JSX is our preferred approach, we should also consider some JSX examples.

notlee (commented):
Nice start, though there are a number of APIs we haven't covered yet

components/o-typography/MIGRATION.md
```diff
@@ -1,5 +1,119 @@
...
+<h3 class="o-typography-heading-level-3">Heading 3</h3>
+<h4 class="o-typography-heading-level-4">Heading 4</h4>
+<h5 class="o-typography-heading-level-5">Heading 5</h5>
+<h6 class="o-typography-heading-level-6">Heading 6</h6>
+```
+
+Replace with:
+
+```html
+<h1 class="o3-typography o3-typography-heading o3-typography-heading--level-1">Heading 1</h1>
```
notlee, line 82: This shows o3 markup is quite a lot more verbose. Now we can [optionally set component tokens at the brand root](https://github.com/Financial-Times/origami/pull/1605), are we able to delete all uses of the `o3-typography` class? I would also recommend removing use of modifiers on all utility classes so. So 

`o3-typography o3-typography-heading o3-typography-heading--level-1`
becomes
`o3-typography-heading-level-1`
or even
`o3-typography-h1`!


This applies to other elements such as strong/lists/etc.

frshwtr: `o-typography-heading` is present as there are shared styles between each heading. As the class name varies between each in this solution, it increases duplication in our code as this following rule will need to be duplicated in each `o-typography-h1`, `o-typography-h2` etc..

I can see `o3-typography` being removed. But by culling all classes down to one makes it harder for us to maintain the headings.

components/o-typography/MIGRATION.md
frshwtr: This should also be applied to [o-buttons migration guide](https://github.com/Financial-Times/origami/blob/main/components/o-buttons/MIGRATION.md).

components/o-typography/MIGRATION.md
```diff
@@ -1,5 +1,348 @@
...
+```
+
+Can be replaced with:
+
+**CSS**
+```css
+@import "@financial-times/o3-foundation/css/core.css";
+
+.my-class {
+	font-size: var(--o3-font-size-1);
```
notlee, line 309: We have 2 font scales now, one for metric2, one for the rest. We should mention that here and link to the beta site.

components/o-typography/MIGRATION.md
```diff
@@ -1,5 +1,348 @@
...
+
+**SASS**
+
+```scss
+.my-class {
+	max-width: oTypographyMaxLineWidth(65);
+}
+```
+
+Our CSS Custom Property can be used instead where a value of `65ch` is required:
```
notlee, line 325: `65ch` might not always be true:

```suggestion
Use `--o3-typography-max-line-width` or set a [custom length using a "ch" value](https://developer.mozilla.org/en-US/docs/Web/CSS/length).
```

components/o-typography/MIGRATION.md
```diff
@@ -1,5 +1,348 @@
...
+
+#### Variables
+
+##### o-typography-is-silent
+##### o-typography-relative-units
+##### o-typography-load-fonts
+##### o-typography-error-for-missing-fonts
+##### o-typography-progressive-font-loading
+##### o-typography-baseline-unit
+
```
notlee, line 345: What about these?

components/o-typography/MIGRATION.md
notlee: https://origami-beta.ft.com/guides/typography/

components/o-typography/MIGRATION.md
frshwtr: I forgot to push the change 🙈 Fixed

notlee:
The font family is still incorrect, is that something for this PR or another?
Screenshot 2024-05-14 at 15 05 09

notlee (approved):

PR #1623, opened 2024-04-30
backstage: bump o-tracking internal version

danail-s:
## Describe your changes
Update the internal version of o-tracking

### DIFF
--- a/libraries/o-tracking/src/javascript/tracking.js
+++ b/libraries/o-tracking/src/javascript/tracking.js
@@ -17,7 +17,7 @@ const initPage = page.init;
  *
  * @type {string}
  */
-const version = '4.5.0';
+const version = '4.5.3';
 
 /**
  * The source of this event.
--- a/libraries/o-tracking/src/javascript/tracking.js
+++ b/libraries/o-tracking/src/javascript/tracking.js
@@ -17,7 +17,7 @@ const initPage = page.init;
  *
  * @type {string}
  */
-const version = '4.5.3';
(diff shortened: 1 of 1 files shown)


FranciscoRuiz-FT (approved):

libraries/o-tracking/src/javascript/tracking.js
```diff
@@ -17,7 +17,7 @@ const initPage = page.init;
  *
  * @type {string}
  */
-const version = '4.5.0';
+const version = '4.5.3';
```
notlee, line 20: o-tracking is currently 4.5.3 but, when this is released, it'll be 4.5.4 😅 

```suggestion
const version = '4.5.4';
```

notlee: Oh, right! That's correct. But if we don't release it won't be published to npm, and won't be available to users.

notlee: Oh, right! That's correct. However we'll need to release to get the version updated in production. Looks like it's send with tracking events and I guess that's what you want to update?

danail-s: Yes, that's correct. Thanks!

notlee (approved):

libraries/o-tracking/src/javascript/tracking.js
notlee: Approved! 👍 I'll go ahead and release

PR #1624, opened 2024-04-30
Or 556/introduce max line width variable

frshwtr:


components/o-typography/MIGRATION.md
```diff
@@ -1,5 +1,119 @@
 ## Migration guide
 
+### Migrating from v7 to o3-typography
```
frshwtr, line 3: Nono why is this part of this

components/o3-typography/main.css
```diff
@@ -5,6 +5,11 @@
 @import './footer.css';
 @import './caption.css';
 
+
+.o3-typography {
```
frshwtr, line 9: Would this be better under `:root`? Feels like we're adding more cases where `o3-typography` is required. Equally we may expose a var that someone doesn't need to use into their css scope.

components/o3-typography/README.md
```diff
@@ -193,6 +194,20 @@ import { Wrapper, Heading, Body } from '@financial-times/o3-typography';
...
+O3 Typography uses a custom property to set the maximum line width for text. This can be overridden by setting the `--o3-typography-optimal-chars-per-line` custom property in the selector where you wish to apply this rule.
+
+##### CSS
+```css
+.my-selector {
+	--o3-typography-optimal-chars-per-line: 60;
+	max-width: var(--o3-typography-max-line-width);
+}
+```
+
```
notlee, line 210: This example is broken as there's no unit provided and there's no positive example of Origami use, which could cause confusion regarding what we actually recommend. How about?

```suggestion
`o3-typography` includes a CSS Custom Property `--o3-typography-max-line-width` to limit typography line length for readability. 

E.g. 

```css
.my-selector {
	max-width: var(--o3-typography-max-line-width);
}
```

```

You may want to slightly decrease or increase this limit inline with our [design guidelines](https://origami-for-everyone.ft.com/guides/typography/). To do so we recommend using the [ch relative unit](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Values_and_Units#distance_units).

E.g. 

```css
.my-selector {
	max-width: 70ch;
}
```

```

File: components/o3-typography/main.css [hunk 4]
```diff
@@ -5,6 +5,11 @@
 @import './footer.css';
 @import './caption.css';
 
+
+:root {
+	--o3-typography-max-line-width: 65ch;
```
notlee, line 10: Using Metric2 and lorem ipsum I've seen a line as long as 90 characters with this, well above our target range. I'd suggest reducing this to 60ch. Happy to be correct though if my little test is inaccurate.

notlee (approved):

PR #1625, opened 2024-04-30
fix: allow targeting to be passed as an option in o-video instances

Shamarni96:
## Describe your changes
This PR allows the targeting property of the Video class to be passed in as an option when creating new o-video instances. currently when attempting to modify the targeting object in an instance class it requires two steps, creating a new instance variable and then accessing the targeting object on this variable. This change will allow for single step changes to be made when attempting to modify the properties of the targeting object. see (https://github.com/Financial-Times/next-article/pull/5507)

https://financialtimes.atlassian.net/browse/ADSDEV-1830


notlee:
@sami-ft @Shamarni96 Hey both, just dropping in to recommend against renaming `targeting`. This is part of the public API and so would be a breaking change, require a major version release with migration guide. An alternative approach could be to alias it, if you think the tech debt is worth it.

Shamarni96:
> @sami-ft @Shamarni96 Hey both, just dropping in to recommend against renaming `targeting`. This is part of the public API and so would be a breaking change, requiring a major version release with migration guide. An alternative approach could be to alias it, if you think the tech debt is worth it.

@notlee Oooh thanks for this, I have reverted that change 👍

umbobabo (approved):

notlee:
Hey @sami-ft, that's right. A new option could have been `adsTargeting` provided the public `targeting` property stayed the same. When I left my comment the `targeting` property had been renamed and I thought that's what we were talking about.

PR #1631, opened 2024-05-02
Add depth to o-multi-select

nadiasuleyman:
## Describe your changes
Include depth to each option in o-multi-select

[Jira](https://financialtimes.atlassian.net/browse/SP-3623) ticket 

### DIFF
--- a/components/o-multi-select/src/js/multi-select-options.js
+++ b/components/o-multi-select/src/js/multi-select-options.js
@@ -134,7 +134,7 @@ export function createOption(idBase, option, index, selected) {
 	optionEl.id = `${idBase}-${index}`;
 	optionEl.className = 'o-multi-select-option';
 	optionEl.setAttribute('aria-selected', selected);
-	optionEl.textContent = option.label;
+	optionEl.textContent = "- ".repeat(option.depth || 0) + option.label;
 	const tickSpan = document.createElement('span');
 	tickSpan.className = 'o-multi-select-option-tick';
 	optionEl.appendChild(tickSpan);
--- a/components/o-multi-select/src/js/multi-select.js
+++ b/components/o-multi-select/src/js/multi-select.js
@@ -3,11 +3,15 @@ import {
 	toggleDropdown,
 	updateCurrentElement,
 	_removeCurrentClass,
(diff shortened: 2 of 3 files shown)


notlee (changes requested):
Hey, thanks for the PR! There are a couple of issues with this approach which blocks our adoption in Origami.

1. The hierarchy is not communicated well to screen reader users. It hinders the current experience as nested options will be read annoyingly as "dash dash dash [option]". 
2. There's no interaction between the group and nested options. I would expect selecting a group would select all child options. E.g. [vue-multiselect](https://vue-multiselect.js.org/#sub-option-groups).
3. We would want to involve the design team to document design guidelines for this pattern and add this pattern to Figma.

As a next step I think we need design collaboration to work with us on creating an accessible UI. We need to validate this is the right design pattern for your input, spec out interactions on adding/removing groups, and write design guidelines. 

I'm happy to meet up to discuss next steps, we have our open meeting at 2pm today if that works for you / Arjun :) 

**Appendix:**

Some other thoughts / brain dump...

[Grouped listboxs](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/) may help us achieve this kind of pattern whilst communicating groups to assistive technologies.

Is this the right UI? The WordPress UI being replaced used nested checkboxes and the [indeterminate state](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate).

notlee:
I'm actually cancelling the open meeting because I'm the only one in today, but happy to chat separately :)

nadiasuleyman:
> [Grouped listboxs](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/) may help us achieve this kind of pattern whilst communicating groups to assistive technologies.

Thanks for this! I'd need the parent option to be an option too, not just the children so I don't think this would fit the requirements.

> Is this the right UI? The WordPress UI being replaced used nested checkboxes and the [indeterminate state](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate).

It would be nice to have a chat about this, as well as the other points you raised, sometime today/next week, if you have time? :)

PR #1633, opened 2024-05-07
Or 665 editorial components

akomiqaia:
## Describe your changes
Adds: 
- body elements
- details

and bigger components such as:

-  typography headline with underline
- quote
- big number
- byline


components/o3-editorial-typography/main.css
```diff
@@ -1 +1,4 @@
+@import './layer.css';
```
frshwtr, line 1: What's the purpose behind the layer css?

components/o3-editorial-typography/src/tsx/body.tsx
```diff
@@ -0,0 +1,14 @@
+import {BodyProps, Attributes} from '../types/index';
+
+export const Body = ({type = 'small', theme, children}: BodyProps) => {
+	const attributes: Attributes = {};
+	if (theme == 'inverse') {
```
frshwtr, line 5: This check is repeated across all three components that are currently being put into o3. We should look to reduce the duplication of this.

There's an implementation in [o3-typography](https://github.com/Financial-Times/origami/blob/main/components/o3-typography/src/tsx/getStyleAttributes.tsx). Maybe we can move the function to a more generic space for all o3 components to use.

components/o3-editorial-typography/stories/story-templates.tsx
```diff
@@ -1,25 +1,232 @@
...
+	BylineProps,
+} from '../src/types/index';
 
 type HeadlineStory = Omit<StoryObj, 'args'> & {
-	args: Omit<HeadlineProps, 'children'>;
-}
+	args: Omit<HeadlineProps, 'children'> & {content: string};
+};
 
+type BodyStory = Omit<StoryObj, 'args'> & {
```
frshwtr, line 24: `Omit<StoryObj, 'args'>` can be assigned to a type alias to reduce the repetition if needed.

```tsx
type StoryObjNoArgs = Omit<StoryObj, 'args'>
```

components/o3-editorial-typography/main.css
akomiqaia: nothing much except setting text colours for typography classes and inversing it. and it is imported before others so later timestamp colours can override default text/inverse-text colours. Also it reduces code and instead of repeating the same pattern in all files we can just put it in one file. We could just write that code in any other file and it will still work but it might not be as clean and readable for other contributors. Open to other suggestions as well

components/o3-editorial-typography/details.css
```diff
@@ -0,0 +1,175 @@
...
+	font-size: var(--_o3-editorial-typography-use-case-quote-caption-font-size);
+	font-weight: var(
+		--_o3-editorial-typography-use-case-quote-caption-font-weight
+	);
+	line-height: var(
+		--_o3-editorial-typography-use-case-quote-caption-line-height
+	);
+}
+
+blockquote.o3-editorial-typography-quote {
```
notlee, line 110: Is the distinction between `.o3-editorial-typography-quote` and `blockquote.o3-editorial-typography-quote` intentional?

components/o3-editorial-typography/details.css
```diff
@@ -0,0 +1,175 @@
...
+	text-transform: uppercase;
+	color: var(--_o3-editorial-typography-byline-timestamp);
+}
+
+[data-o3-theme='inverse'] .o3-editorial-typography-byline-timestamp,
+.o3-editorial-typography-byline-timestamp[data-o3-theme='inverse'] {
+	color: var(--_o3-editorial-typography-byline-timestamp-inverse);
+}
+
+.o3-editorial-typography-quote {
```
notlee, line 76: This is _very_ different from our current editorial quote. Are these intentional design changes or a mistake?

Left: o2; Right: o3
Screenshot 2024-05-09 at 11 38 01


https://o2.origami.ft.com/?path=/story/o2-core_components-o-quote--editorial-quote

components/o3-editorial-typography/details.css
```diff
@@ -0,0 +1,175 @@
...
+	background-color: currentColor;
+}
+
+blockquote.o3-editorial-typography-quote > p {
+	margin-top: 0;
+	margin: var(--o3-spacing-2xs) 0;
+}
+
+.o3-editorial-typography-quote-author,
+.o3-editorial-typography-quote-caption {
```
notlee, line 133: Following BEM these should use underscores to indicate they are an element of quote, which shouldn't be used independently

`o3-editorial-typography-quote__author`
`o3-editorial-typography-quote__caption`

components/o3-editorial-typography/details.css
```diff
@@ -0,0 +1,175 @@
...
+	font-style: normal;
+}
+
+.o3-editorial-typography-big-number,
+.o3-editorial-typography-big-number > div {
+	display: block;
+	unicode-bidi: isolate;
+}
+
+.o3-editorial-typography-big-number-title {
```
notlee, line 144: Same here `.o3-editorial-typography-big-number__title`

components/o3-editorial-typography/details.css
```diff
@@ -0,0 +1,175 @@
...
+	font-weight: var(
+		--_o3-editorial-typography-use-case-big-number-title-font-weight
+	);
+	line-height: var(
+		--_o3-editorial-typography-use-case-big-number-title-line-height
+	);
+	margin: 0 0 var(--o3-spacing-5xs);
+}
+
+.o3-editorial-typography-big-number-content {
```
notlee, line 161: `o3-editorial-typography-big-number__content` 

etc...

components/o3-editorial-typography/layer.css
```diff
@@ -0,0 +1,9 @@
+[class^='o3-editorial-typography-'] {
+	color: var(--_o3-editorial-typography-text-color);
+	margin: 0;
+}
+
+[data-o3-theme='inverse'] [class^='o3-editorial-typography-'],
```
notlee, line 6: Funky! Love it!

components/o3-editorial-typography/details.css
```diff
@@ -0,0 +1,175 @@
...
+
+.o3-editorial-typography-caption {
+	font-family: var(--_o3-editorial-typography-use-case-caption-font-family),
+		sans-serif;
+	font-size: var(--_o3-editorial-typography-use-case-caption-font-size);
+	font-weight: var(--_o3-editorial-typography-use-case-caption-font-weight);
+	line-height: var(--_o3-editorial-typography-use-case-caption-line-height);
+}
+
+.o3-editorial-typography-byline-author {
```
notlee, line 25: Our current [byline author](https://o2-core.origami.ft.com/?path=/story/components-o-editorial-typography-byline-with-timestamp--byline-with-timestamp&globals=backgrounds:!undefined) often links to the author's page. I believe it may not always be a link though.

When it is a link, how should it be styled? Is it changing from o2?

components/o3-editorial-typography/stories/core/detail.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import type {Meta} from '@storybook/react';
```
notlee, line 1: Detail feels like a quite random grouping, it doesn't make it very easy to find the right story. 

1. It seems to allow quote author / caption use outside of the quote component, which I don't think we want.
2. Perhaps speak to Mary about how we want to present the other "details". Maybe one story each. Maybe we can group some, like if it's the most common use we may want to present a byline Story.

Screenshot 2024-05-09 at 11 47 07
these.

components/o3-editorial-typography/stories/core/body.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import type {Meta} from '@storybook/react';
+import {Body as BodyTsx} from '../../src/tsx/index';
```
notlee, line 2: Are you adding support for SV? One for Mary and Deyan.

notlee (changes requested):

components/o3-editorial-typography/details.css
akomiqaia: not really. I think `blockquote` can be removed here

components/o3-editorial-typography/stories/core/body.stories.tsx
akomiqaia: I will add support but we don't have Figma files in SV or Core. I am developing just Core and then copy over story files.

components/o3-editorial-typography/details.css
notlee: 👍 on the icon issue, I always forget that

components/o3-editorial-typography/details.css
akomiqaia: well, byline is only a wrapper here and users can pass custom markup and styling of that should be up to users. We could include more in the readme maybe?

components/o3-editorial-typography/stories/core/detail.stories.tsx
akomiqaia: yeah this is a good suggestion. Will talk to Marry and see what she thinks

components/o3-editorial-typography/details.css
notlee: @akomiqaia There are 2 quote styles here

**Pull Quote**

This _is not_ currently in Origami. It's defined [custom in cp-content-pipeline](https://github.com/Financial-Times/cp-content-pipeline/blob/97ffc6689b6d8d78635728f77e3a35c0598689b4/packages/styles/scss/_pullquotes.scss#L1). Do we want to add to Origami now? Probably not a bad idea! 
Screenshot 2024-05-09 at 12 23 28

**Block Quote**

This _is_ in Origami already, and is [imported by cp-content-pipeline](https://github.com/Financial-Times/cp-content-pipeline/blob/97ffc6689b6d8d78635728f77e3a35c0598689b4/packages/styles/scss/_blockquotes.scss#L1). Here's an [example on a dummy ft article](https://www.ft.com/kitchenus-sinkus-articlus).

Screenshot 2024-05-09 at 12 25 30

We should port this over.

____

Perhaps Mary isn't aware. Are you happy to raise with her to decide whether to add both pull and block quotes?

components/o3-editorial-typography/details.css
notlee: Yes please 👍 We should document what customisations we allow and include an API to do that without overriding CSS. E.g. we give people a public CSS Custom Property to set rather than them setting random styles

components/o3-editorial-typography/stories/core/body.stories.tsx
notlee: 👍 Do you know if Mary has worked with Deyan or Marin to design SV alternatives?

components/o3-editorial-typography/details.css
akomiqaia: Mary will update the Figma file and will support both types of quote components.

components/o3-editorial-typography/details.css
akomiqaia: Okay, I will think about this. This sounds like a good idea.

components/o3-editorial-typography/main.css
frshwtr: I think the separation is fine :) But there might be a name for the file that helps communicate it's purpose. Something like `theme.css` or `theming.css`

components/o3-editorial-typography/README.md
```diff
@@ -89,10 +90,10 @@ import { Headline } from '@financial-times/o3-editorial-typography/cjs';
...
 O3 editorial paragraphs are styled with the `o3-editorial-typography-body__-[type]` class. Where `type` can be either `large` or `small`.
 
-| Type  | Selector                            |
-| ----- | ----------------------------------- |
-| Large | .o3-editorial-typography-body__large |
-| Small | .o3-editorial-typography-body__small |
+| Type  | Selector                               |
+| ----- | -------------------------------------- |
+| Large | .o3-editorial-typography-body\_\_large |
+| Small | .o3-editorial-typography-body\_\_small |
```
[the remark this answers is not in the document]
notlee: Actually... article body is supposed to be responsive, similar to the headline, there shouldn't be a large and small variant.

components/o3-editorial-typography/details.css
```diff
@@ -36,6 +48,17 @@
...
 
+a.o3-editorial-typography-byline-author {
+	--o3-editorial-typography-byline-author-color: var(--_o3-editorial-typography-byline-author);
+	--o3-editorial-typography-byline-author-hover-color: var(--_o3-editorial-typography-byline-author-hover);
+	color: var(--o3-editorial-typography-byline-author-color);
+	text-decoration: none;
+}
+
+a.o3-editorial-typography-byline-author:hover {
+	color: var(--o3-editorial-typography-byline-author-hover-color);
```
notlee, line 59: We need an inverse variant on hover here!

Screenshot 2024-05-14 at 14 03 54

components/o3-editorial-typography/stories/story-templates.tsx
```diff
@@ -1,25 +1,347 @@
...
+			},
+		},
+	},
+	render: args => {
+		if (args.openInNewTab) {
+			args.anchorTarget = '_blank'
+		}
+		return (
+			<BodyTsx type="small">
+				<div>
```
notlee, line 240: A `div` inside a `p` tag isn't valid html – I'm seeing errors in the console from the html tab plugin

components/o3-editorial-typography/details.css
```diff
@@ -0,0 +1,202 @@
...
+
+.o3-editorial-typography-quote__pull {
+	font-family: var(--_o3-editorial-typography-use-case-quote-font-family),
+		sans-serif;
+	font-size: var(--_o3-editorial-typography-use-case-quote-font-size);
+	font-weight: var(--_o3-editorial-typography-use-case-quote-font-weight);
+	line-height: var(--_o3-editorial-typography-use-case-quote-line-height);
+}
+
+.o3-editorial-typography-quote__block {
```
notlee, line 107: The blockquote is inline with content and should use Georgia.
Are the design changes to the author/caption intentional? 

Screenshot 2024-05-14 at 14 48 02

components/o3-editorial-typography/README.md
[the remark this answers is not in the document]
akomiqaia: so it depends on a view-port? I think I asked Mary about this but that was not the case. Might be wrong but can double check

components/o3-editorial-typography/list.css
```diff
@@ -0,0 +1,26 @@
...
+[class^="o3-editorial-typography-list--"] li::before {
+	margin-right: var(--o3-spacing-xs);
+}
+
+.o3-editorial-typography-list--unordered li::before {
+	content: '\2022'; /* dot character */
+}
+
+.o3-editorial-typography-list--ordered {
+	counter-reset: item;
```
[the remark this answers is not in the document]
akomiqaia: what are they? unordered and ordered sounds like list modifiers :D

components/o3-editorial-typography/details.css
akomiqaia: I am waiting for Figma designs and tokens for the quote component and updates can happen afterwards. Do you have an actual example of a block quote on ft pages? The quotes I checked were all in metric font on ft.com

notlee (approved):
✨

PR #1654, opened 2024-05-20
fix: small button icon size and icon-only ratio (square)

notlee:
Before / After:
Screenshot 2024-05-20 at 11 54 06

Small, icon only button now aligns with our designs:
- Square
- 20px icon

Screenshot 2024-05-20 at 11 54 24


notlee:
Second commit correctly sets the inline padding to 16px for the default button
Screenshot 2024-05-20 at 12 03 24

akomiqaia (approved):
LGTM

PR #1658, opened 2024-05-20
feat: Add o-autocomplete configurable highlighting

andygout:
## Describe your changes

This PR applies changes so that custom `suggestionTemplate` callbacks provided as an option when instantiating an instance of `oAutocomplete` will now receive two additional arguments:
- `highlightSuggestion()` — as declared [here](https://github.com/Financial-Times/origami/blob/55a9552fedef246e152623690efa2d441fef9dea/components/o-autocomplete/src/js/autocomplete.js#L21-L34)
- `isMatchHighlighted` — a boolean to determine whether matching or non-matching characters should be highlighted, which in turn is passed as an argument to the `highlightSuggestion()` function within a custom `suggestionTemplate()` function (that has been passed as an option when instantiating an instance of `oAutocomplete`) or within the [native `suggestionTemplate()` function](https://github.com/Financial-Times/origami/blob/55a9552fedef246e152623690efa2d441fef9dea/components/o-autocomplete/src/js/autocomplete.js#L392-L412)

In the scenario of a custom `suggestionTemplate`, I'm not sure if it's a code smell or anti-pattern that the `isMatchHighlighted` value is passed at instantiation and then reappears via the `suggestionTemplate()` callback where there is technically no compulsion to use that value as intended, i.e. given `suggestionTemplate()` is custom, the `isMatchHighlighted` value could be declared within that function, making its presence as a parameter redundant. However, it seems far neater to declare up-front at time of instantiation all the configuration that dictates how you would like the component to behave.

I would also a prefer a better variable name than `isMatchHighlighted` — set to `true` it makes sense, though when set to `false` it implies that it will simply not highlight matched characters (and by inference, that it will highlight nothing at all) rather than expressing that it _will_ highlight the characters that are not a match. Naming is hard…

### Scenario 1

Custom `suggestionTemplate` with highlights applied to options where it matches query

- `suggestionTemplate` option is provided
- `isMatchHighlighted` option is provided with a value of `true`

```js
function suggestionTemplate (option, query, highlightSuggestion, isMatchHighlighted) {
	…
}

new oAutocomplete(oAutocompleteElement, {
	suggestionTemplate,
	isMatchHighlighted: true,
	…
});
```

---

### Scenario 2

Custom `suggestionTemplate` with highlights applied to options where it does NOT match query

- `suggestionTemplate` option is provided
- `isMatchHighlighted` option is NOT provided (or is provided with a value of `false`)

```js
function suggestionTemplate (option, query, highlightSuggestion, isMatchHighlighted) {
	…
}

new oAutocomplete(oAutocompleteElement, {
	suggestionTemplate,
	…
});
```

OR

```js
function suggestionTemplate (option, query, highlightSuggestion, isMatchHighlighted) {
	…
}

new oAutocomplete(oAutocompleteElement, {
	suggestionTemplate,
	isMatchHighlighted: false,
	…
});
```

---

### Scenario 3

o-autocomplete native `suggestionTemplate` with highlights applied to options where it matches query

- `suggestionTemplate` option is NOT provided
- `isMatchHighlighted` option is provided with a value of `true`

```js
new oAutocomplete(oAutocompleteElement, {
	isMatchHighlighted: true,
	…
});
```

---

### Scenario 4

o-autocomplete native `suggestionTemplate` with highlights applied to options where it does NOT match query

- `suggestionTemplate` option is NOT provided
- `isMatchHighlighted` option is NOT provided (or is provided with a value of `false`)

```js
new oAutocomplete(oAutocompleteElement, {
	…
});
```

OR

```js
new oAutocomplete(oAutocompleteElement, {
	isMatchHighlighted: false,
	…
});
```

---

### DIFF
--- a/components/o-autocomplete/src/js/autocomplete.js
+++ b/components/o-autocomplete/src/js/autocomplete.js
@@ -16,18 +16,19 @@ import accessibleAutocomplete from '@financial-times/accessible-autocomplete';
 /**
  * @param {string} suggestion - Text which is going to be suggested to the user
  * @param {string} query - Text which was typed into the autocomplete text input field by the user
+ * @param {boolean} isMatchHighlighted - Boolean to determine whether matching or non-matching characters should be highlighted.
  * @returns {CharacterHighlight[]} An array of arrays which contain two items, the first is the character in the suggestion, the second is a boolean which indicates whether the character should be highlighted.
  */
-function highlightSuggestion(suggestion, query) {
+function highlightSuggestion(suggestion, query, isMatchHighlighted) {
 	const result = suggestion.split('');
 
 	const matchIndex = suggestion.toLocaleLowerCase().indexOf(query.toLocaleLowerCase());
 	return result.map(function(character, index) {
-		let shouldHighlight = true;
+		let shouldHighlight = !isMatchHighlighted;
 		const hasMatched = matchIndex > -1;
 		const characterIsWithinMatch = index >= matchIndex && index <= matchIndex + query.length - 1;
 		if (hasMatched && characterIsWithinMatch) {
-			shouldHighlight = false;
+			shouldHighlight = isMatchHighlighted;
 		}
 		return [character, shouldHighlight];
 	});
@@ -182,6 +183,7 @@ function initClearButton(instance) {
  * @property {MapOptionToSuggestedValue} [mapOptionToSuggestedValue] - Function which transforms a suggestion before rendering.
  * @property {onConfirm} [onConfirm] - Function which is called when the user selects an option
  * @property {SuggestionTemplate} [suggestionTemplate] - Function to override how a suggestion item is rendered.
+ * @property {boolean} [isMatchHighlighted] - Boolean to determine whether matching or non-matching characters should be highlighted.
  * @property {boolean} [autoselect] - Boolean to specify whether first option in suggestions list is highlighted.
  */
 
@@ -209,6 +211,9 @@ class Autocomplete {
 		}
 		if (opts.suggestionTemplate) {
 			this.options.suggestionTemplate = opts.suggestionTemplate;
+		}
+		if (opts.isMatchHighlighted) {
+			this.options.isMatchHighlighted = Boolean(opts.isMatchHighlighted);
 		}
 		if (opts.autoselect) {
 			this.options.autoselect = opts.autoselect;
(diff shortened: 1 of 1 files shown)


frshwtr:
Thanks for PR, sorry for such a late review :(

> I would also a prefer a better variable name than isMatchHighlighted — set to true it makes sense, though when set to false it implies that it will simply not highlight matched characters (and by inference, that it will highlight nothing at all) rather than expressing that it will highlight the characters that are not a match. Naming is hard…

I think this in itself is leading me to think the decision on what to highlight should be delegated to inside the `suggestionTemplate` function. This may help reduce confusion in its name and ensure; if there is custom behaviour, then the author of the function must decide if they want to highlight within that function.

Alternatively, renaming to `invertHighlighting` may also be an option. Though I think this may be just as unclear 😧

andygout:
No worries at all 🙂 

> I think this in itself is leading me to think the decision on what to highlight should be delegated to inside the `suggestionTemplate` function.

Would this prevent Scenario 3 described above? I.e. the consumer only wants to invert the highlighting but does not want to create a custom `suggestionTemplate` function to do so.

If I've understood that correctly then it feels like renaming the variable would be the better option to retain the flexibility of that scenario.

I don't think `invertHighlighting` is too bad, though it does depend on the consumer first identifying what the default is to understand what the inverse would be.

The alternatives I can suggest (and which are essentially equivalent to one another) are:
- `isHighlightCorrelativeToMatch`
- `isHighlightCorrespondentToMatch`
- `isHighlightCorrespondingToMatch`
- or a variation on this theme

The adjustment of wording means that it now implies that the highlight is always present, with the remaining words determining how that highlighting is applied.

How do you feel about these?

frshwtr:
> Would this prevent Scenario 3 described above? I.e. the consumer only wants to invert the highlighting but does not want to create a custom suggestionTemplate function to do so.

We could create this as a standard function that gets exported as part of the library also.

How do you feel about these?

`isHighlightCorrespondingToMatch` is the clearest to me out of these three.

I think we can go with the option to keep the boolean in place :)

frshwtr:
> If the latter then I will take this PR out of draft mode and mark as "Ready for review".
Sorry for late reply, this can be done later.

components/o-autocomplete/src/js/autocomplete.js
```diff
@@ -210,6 +212,9 @@ class Autocomplete {
 		if (opts.suggestionTemplate) {
 			this.options.suggestionTemplate = opts.suggestionTemplate;
 		}
+		if (opts.isHighlightCorrespondingToMatch) {
+			this.options.isHighlightCorrespondingToMatch = Boolean(opts.isHighlightCorrespondingToMatch);
+		}
```
frshwtr: Wondering if it's simpler to take of Boolean coercion in `Boolean` to set this to false instead of only applying when value is present? It would mean we can remove the guarding on L302. What do you think?

```suggestion
	this.options.isHighlightCorrespondingToMatch = Boolean(opts.isHighlightCorrespondingToMatch);

```

PR #1661, opened 2024-05-21
feat: update o3-foundation and remove o3-editorial use case tokens

akomiqaia:
## Describe your changes
removes unnecessary tokens and implements speech-left icon


notlee:
Closing as we reworked history on the existing branch

PR #1666, opened 2024-05-23
fix: Fix link focus rings [OR-702]

notlee:
Add some point the token `o3-focus-use-case-outline-color` was updated to represent a box shadow and not a colour. It would be a breaking change to rename without mapping. This PR uses the value as a shadow rather than outline colour.

### DIFF
--- a/components/o3-foundation/focus.css
+++ b/components/o3-foundation/focus.css
@@ -1,11 +1,12 @@
 /* focus styles */
 :focus {
-	outline: 2px solid var(--o3-focus-use-case-outline-color);
+	box-shadow: var(--o3-focus-use-case-outline-color);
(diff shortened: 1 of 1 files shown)


components/o3-foundation/focus.css
```diff
@@ -1,11 +1,12 @@
 /* focus styles */
 :focus {
-	outline: 2px solid var(--o3-focus-use-case-outline-color);
+	box-shadow: var(--o3-focus-use-case-outline-color);
+	outline: 0;
```
notlee, line 4: I thought best to set to `0` to avoid default browser styles / o2 conflicts

akomiqaia (approved):

PR #1672, opened 2024-05-29
feat: Allow font output to be turned off for o-layout

notlee:
Helpful if font faces have already been output, e.g. using `o-fonts` directly or custom fonts.

### DIFF
--- a/components/o-layout/main.scss
+++ b/components/o-layout/main.scss
@@ -14,20 +14,35 @@
 @import 'src/scss/grid';
 @import 'src/scss/grid-areas';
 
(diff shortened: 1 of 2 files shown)


components/o-layout/main.scss
```diff
@@ -37,7 +52,10 @@
 
 	// 	Typography.
 	@if ($typography) {
-		@include oFonts();
+		// By default fonts are output when typography is enabled.
+		@if ($fonts != false) {
+			@include oFonts();
+		}
```
notlee, line 58: Most of these changes are Prettier. Here's the important bit

akomiqaia (approved):

PR #1674, opened 2024-05-29
Typography component tokens

MaryGodservant:
- Use use-case tokens directly.
- Correct link focus styles across ui & editorial typography
- Delete unused/duplicate tokens
- Add whitelabel tokens for editorial type


akomiqaia (approved):
LGTM

PR #1676, opened 2024-05-30
feat: add preview element

akomiqaia:
## Describe your changes
Add preview component for astro web page

[OR-717](https://financialtimes.atlassian.net/browse/OR-717)

[OR-717]: https://financialtimes.atlassian.net/browse/OR-717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/for-everyone-website/src/components/utils/Preview.astro
```diff
@@ -0,0 +1,217 @@
...
+const {metaData} = Astro.props;
+
+const pathToMetadata = pathReoslve(metaData);
+const Component = await import(pathToMetadata);
+
+const fileContent = await readFile(pathToMetadata, {
+	encoding: 'utf8',
+});
+
+const regex = /\/\/\s*<preview>([\s\S]*?)\/\/\s*<\/preview>/;
```
notlee, line 17: Between `preview` comments this matches on white-space and non-white-space – effectively anything ([visualisation](https://regexper.com/#%2F%5C%2F%5C%2F%5Cs*%3Cpreview%3E%28%5B%5Cs%5CS%5D*%3F%29%5C%2F%5C%2F%5Cs*%3C%5C%2Fpreview%3E%2F))

Could we make it a little simpler with?

`/\/\/\s*<preview>(.*?)\/\/\s*<\/preview>/`

apps/for-everyone-website/package.json
```diff
@@ -34,5 +34,8 @@
 	},
 	"volta": {
 		"node": "22.2.0"
+	},
+	"devDependencies": {
+		"prettier": "3.2.5"
```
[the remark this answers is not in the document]
notlee: Should it be a `dependency` rather than in `devDependencies`? We do build a static site and don't ship prettier directly to users, so I think you could argue either way, but it is relied on by our application's source code and therefore isn't like a typical dev dependency such as linter

apps/for-everyone-website/src/components/utils/Preview.astro
```diff
@@ -0,0 +1,217 @@
...
+const match = fileContent.match(regex)!; // ! tels ts that match is not null
```
notlee, line 18: Can we guarantee it's not `null`? What if I make a mistake whilst writing a preview?

apps/for-everyone-website/src/components/utils/Preview.astro
```diff
@@ -0,0 +1,217 @@
...
+const JSXString = match[1];
+
+const componentJSX = await format(JSXString, {
+	parser: 'babel',
+});
+
+const componentHTMLString = renderToString(Component.preview());
+const componentFormatedHTML = await format(componentHTMLString, {
```
notlee, line 26: Let's rename this `componentHTML` instead of reassigning?

apps/for-everyone-website/src/components/utils/Preview.astro
```diff
@@ -0,0 +1,217 @@
...
+			} else {
+				codePreview?.classList.remove('visible');
+				componentPreveiw?.classList.add('visible');
+			}
+		});
+	});
+</script>
+
+<style>
+	/* The switch - the box around the slider */
```
notlee, line 93: Are these styles copied from somewhere and if so, why?

notlee (commented):
Love it, love it

apps/for-everyone-website/src/components/utils/Preview.astro
akomiqaia: I think it's a very tiny detail that people might miss and let's why I included it so if someone doesn't use space between comments than it will still work

apps/for-everyone-website/src/components/utils/Preview.astro
akomiqaia: well, it will throw an error. I thought of a fallback message but not sure if we should have it.

apps/for-everyone-website/src/components/utils/Preview.astro
akomiqaia: yeah, they are copied. why? I am not. sure seemed easier :D why?

apps/for-everyone-website/src/components/utils/Preview.astro
notlee: I think we're talking about different things. My point was we could write `([\s\S]*?)` as `(.*?)` and I believe get the same result

apps/for-everyone-website/src/components/utils/Preview.astro
notlee: Yeah that's a better approach, to handle the error rather then lie to poor TypeScript 😄

apps/for-everyone-website/src/components/utils/Preview.astro
notlee: The Starlight component html could change in a minor/patch and break without warning. Is it possible to include the CSS from node_modules?

notlee: Oh my mistake, I thought this was a Starlight component and we were importing its HTML

apps/for-everyone-website/src/components/utils/Preview.astro
akomiqaia: oh i see :D sorry i thought you meant different thing

notlee: Actualllllllyyy.... it looks like `.` doesn't capture line breaks where `\s\S` I think will, that's something to watch out for

notlee (approved):
🚀 🚀 🚀 🚀 🚀 🚀

PR #1686, opened 2024-06-04
Add SV support to editorial typography [OR-578]

itsdeyan:


apps/dictionary/tokens/sustainable-views/components/o3-editorial-typography.json
```diff
@@ -1,40 +1,10 @@
 {
   "o3": {
     "editorial-typography": {
-      "headline-large-l": {
```
frshwtr, line 4: Why were these removed?

itsdeyan:
they were obsolete for SV that's why I deleted them

akomiqaia (approved):
This looks good to me but before merging it we need to generate tokens and also create demos for professional.

akomiqaia (changes requested):
gonna block this since I accidentally approved it

components/o3-foundation/src/css/tokens/sustainable-views/_variables.css
```diff
@@ -18,7 +18,9 @@
   --o3-spacing-4xl: 6rem;
   --o3-font-family-metric: 'metric 2 VF';
   --o3-font-family-financier-display: 'financier display VF';
+  --o3-font-family-financier-text: 'financier text VF';
   --o3-font-family-georgia: georgia;
+  --o3-font-weight-thinnest: 1;
```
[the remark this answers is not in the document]
itsdeyan: The only way to make Financier Text VF appear properly in Figma was to set the variable font weight to the value of 1. That's why I created this, please let me know if it's wrong and what's the right way to do it, thanks.

components/o3-editorial-typography/src/tsx/detail.tsx
```diff
@@ -1,11 +1,20 @@
...
 
 export const TopicTag = ({href, theme, children}: TopicTagProps) => {
 	const attributes = getDataAttributes(theme);
 	const HtmlElement = href ? 'a' : 'span';
 	return (
-		<HtmlElement href={href} className="o3-editorial-typography-topic-tag" {...attributes}>
+		<HtmlElement
+			href={href}
+			className="o3-editorial-typography-topic-tag"
+			{...attributes}>
```
frshwtr, line 17: The `href` attribute will still be present when `HtmlElement` is a `span`. A cleaner approach would be to use the ternary to change what element is returned.
```suggestion
return href ? <a href className="o3-editorial-typography-topic-tag">{children}</a> : <span className="o3-editorial-typography-topic-tag">{children}</span>
```

notlee (changes requested):
Missing tokens for pull quotes and block quotes. Does SV support these? If not, they should be removed from Storybook.
Screenshot 2024-06-07 at 16 09 10

No paper background on SV stories! 🙏 
Screenshot 2024-06-07 at 16 09 34

The `inverse` theme is broken in places. Deyan said we don't need to support the inverse theme for SV. Can we remove it entirely for SV, including tokens and the SB control?
Screenshot 2024-06-07 at 16 10 00

notlee (approved):

PR #1689, opened 2024-06-04
fix!: ci-2234 coral illegal comments url

debugwand:
## Describe your changes

DSA compliance.

The Coral link to report illegal comments works by reopening the same page in a new window, with some added query string items. However, if someone is a first-click free user this will result in the barrier page loading instead. This means that anonymous users can't use the tool to report illegal comments, which is a condition of the DSA.

This change intercepts the clicks on this link, alters the path to a route that is not behind the paywall, then opens that url in a new window.

https://financialtimes.atlassian.net/jira/software/c/projects/CI/boards/1653?assignee=5be57e310e625003a46e9693&selectedIssue=CI-2234

## Link to Figma designs
N/A

### DIFF
--- a/components/o-comments/src/js/stream.js
+++ b/components/o-comments/src/js/stream.js
@@ -2,6 +2,7 @@ import events from './utils/events.js';
 import displayName from './utils/display-name.js';
 import auth from './utils/auth.js';
 import purgeJwtCache from './utils/purge-jwt-cache.js';
+import Delegate from 'ftdomdelegate';
 
 class Stream {
 	/**
@@ -21,6 +22,8 @@ class Stream {
 	}
 
 	init () {
+		this.redirectIllegalCommentReport();
+
 		const renderAndAuthenticate = (displayName) => {
 			return Promise.all([this.renderComments(), this.authenticateUser(displayName)])
 				.then(() => {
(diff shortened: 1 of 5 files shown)


components/o-comments/src/js/stream.js
```diff
@@ -40,6 +43,46 @@ class Stream {
...
+		let redirectReportPath = this.#tidyPath(this.options?.redirectReportPath) || '/article/comment-report/';
+		const sendToCommentReport = function (event, elem) {
+			event.preventDefault();
+			const href = elem.getAttribute('href');
+			// the below will actually work for comments on vanity urls, as the underlying article url is used in coral
+			const newUrl = href.replace(paywalledReportPath, redirectReportPath);
+			window.open(newUrl);
+		}
+
+		document.addEventListener('oComments.ready', () => {
```
notlee, line 75: It looks like the dispatched event is `oCommentsReady` (it aught to be what you're listening for)

components/o-comments/src/js/stream.js
```diff
@@ -40,6 +43,46 @@ class Stream {
...
+			const shadowContainer = this.streamEl.querySelector('#coral-shadow-container');
+			const shadowRoot = shadowContainer?.shadowRoot;
+			const shadowFirstContainer = shadowRoot?.children?.length && shadowRoot.children[0];
+			if (shadowFirstContainer) {
+				const commentReportDelegate = new Delegate(shadowFirstContainer);
```
notlee, line 80: Since we're setting a new root node and there's only one event listener within it, does `Delegate` add any benefit here vs. applying a click event listener directly?

components/o-comments/src/js/stream.js
debugwand: I've found that both fire, but `oCommentsReady` has next to no DOM available, while `oComments.ready` has the shadow root available. I think I got `oComments.ready` from the README

This is what's in the DOM at `oCommentsReady`
Screenshot 2024-06-10 at 10 36 27

This is what's in the DOM at `oComments.ready`
Screenshot 2024-06-10 at 10 37 38


using 
```
document.addEventListener('oComments.ready', () => {
			debugger
alert(".ready") //I put this here to check this event actually exists, which it does
		})
		document.addEventListener('oCommentsReady', () => {
			
			const shadowContainer = this.streamEl.querySelector('#coral-shadow-container');//this is null when debugger runs
			debugger
			})
```
and checking the DOM in the paused states for both

components/o-comments/src/js/stream.js
debugwand: I went for dom delegation because 1) when there are lots of comments, more comments and therefore more links can be loaded in after the initial load 2) different bits of comments load in asynchronously even at initial load. When we get this comments ready event, the outer shell has loaded but not the comments inside (checked by using `debugger` immediately after the ready event and seeing what was in the DOM)

notlee:
Thanks @debugwand, looks good to me! One last thing, do you intent to make this a major release? If so could you please update the [migration table in the readme](https://github.com/Financial-Times/origami/tree/main/components/o-comments#migration)?

debugwand:
> Thanks @debugwand, looks good to me! One last thing, do you intent to make this a major release? If so could you please update the [migration table in the readme](https://github.com/Financial-Times/origami/tree/main/components/o-comments#migration)?

ah sorry, missed that bit. yes I think it needs a major release. have updated both readme and migration files

notlee (approved):

PR #1695, opened 2024-06-06
Prerequisites to reduce duplicate content

notlee:
See commit messages.

### DIFF
--- a/apps/for-everyone-website/remove-branded-headings.mjs
+++ b/apps/for-everyone-website/remove-branded-headings.mjs
@@ -1,0 +1,89 @@
+import fs from 'node:fs/promises';
+import {globby} from 'globby';
+import {minify} from 'html-minifier';
+import {JSDOM} from 'jsdom';
+
+// Get all HTML files from the output directory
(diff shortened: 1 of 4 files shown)


apps/for-everyone-website/remove-branded-headings.mjs
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @param {Element[]} tocComponent - The element containing the table of contents.
+ * @param {string[]} headingIds - An array of heading IDs which exist on the current page.
+ * @returns {number} - A count of removed links to branded headings which do not exist.
+ */
+function removeLinksToMissingHeadings(tocComponent, headingIds) {
+	const items = tocComponent.querySelectorAll('li');
+	const extraItems = Array.from(items).filter(item => {
+		const anchor = item.querySelector('a');
+		const anchorId = anchor.href.replace(/^(.+)#/, '');
```
frshwtr, line 22: Think we might have runtime probllem here if no `a`'s come back. Possibly want to ignore filtering this specific element if there's no a?

PR #1696, opened 2024-06-06
Use new content branding tools to reduce buttons docs

notlee:
See commit messages


frshwtr (approved):

akomiqaia:
can we remove the yellow notification that says `Do these guidelines differ for your brand? Let us know in #design-system-guild.` and mono still seems to appear in the sidebar for different brands

notlee:
@akomiqaia 

1. Yes. That message is unhelpful
2. It will for development. That's a downside of this post-processing approach vs. maintaining a fork. Will create a follow up PR with documentation 👍

PR #1699, opened 2024-06-06
chore: Move patch-package to root

notlee:
This seems to apply now with a fresh install from root


akomiqaia (approved):
Better to move postinstall script

PR #1703, opened 2024-06-10
feat: Add custom params on video

AniaMakes:
## Describe your changes
The Ads team needs to ensure the video ad requests include all the relevant targeting values in `cust_params`.

To achieve this, this PR adds a new o-video `config` object `advertCustomParams`. It is added in such a way that if that object is not present, there is no `cust_params` in the advert targeting url, thus this change can be released as a minor version. 

This screenshot is an example of this wired up into next-video-page - note the presence of `cust_params` in the third line of Request URL and details such as `guid`, `gender`, `slv` which are not there at the moment.
Screenshot 2024-06-10 at 12 04 35

https://financialtimes.atlassian.net/browse/ADSDEV-1881

## Link to Figma designs
n/a


components/o-video/src/js/ads.js
```diff
@@ -109,7 +109,22 @@ class VideoAds {
...
+		const paramsObject = {
+			env: "vp",
+			gdfp_req: 1,
+			impl: "s",
+			output: "xml_vast2",
+			iu: this.video.targeting.site,
+			sz: this.video.targeting.sizes,
+			unviewed_position_start: 1,
+			scp: encodeURIComponent(targeting),
+			...(this.video.advertCustomParams && {cust_params: new URLSearchParams(this.video.advertCustomParams).toString()})
```
AniaMakes, line 121: [Point for discussion] I took the option of relying on `undefined`  to check if the data is present or not. If this doesn't feel defensive enough, I could use `let` and default to `null` - as an empty object isn't falsy I wouldn't be able to easily run the `&&` check.

components/o-video/README.md
```diff
@@ -101,6 +101,7 @@ Where `opts` is an optional object with properties
 	- `position:` sets the `pos` param on the ad request which is used to target video ads similar to how `pos: native` is used to target partner content.
 	- `sizes:` also added to ad request to specify video ad sizes to be returned.
 	- `videoId:` added to targeting as a param for targeting ads at a specific video.
+- `advertCustomParams` `[Object]` object containing custom parameters that are used to request video ads. Values include, but are not limited to: user ID, page type, subscription level, resolution.
```
olusoladavid, line 104: _question/suggestion:_ Could we re-use the `targeting` field? It was created for the same purpose of passing video ad targeting information

components/o-video/src/js/ads.js
```diff
@@ -109,7 +109,22 @@ class VideoAds {
 			targeting += `&brand=${brand}`;
 		}
 
-		const advertisingUrl = `http://pubads.g.doubleclick.net/gampad/ads?env=vp&gdfp_req=1&impl=s&output=xml_vast2&iu=${this.video.targeting.site}&sz=${this.video.targeting.sizes}&unviewed_position_start=1&scp=${encodeURIComponent(targeting)}`;
+		const paramsObject = {
```
olusoladavid, line 112: **praise:** thanks for bringing a fresh air of readability to this code

components/o-video/README.md
AniaMakes: I guess we could? I think in my brain `targeting` in itself doesn't make it clear that it's for adverts, that's why I set up a key that is explicitly so.

notlee: I think you're right the property could be named more clearly. However advert configuration is now split between `targeting` and `advertCustomParams`, which seems more confusing. 

At a glance it looks like `targeting` may only be used for advert config, is that the case?

AniaMakes: @notlee I understand that's the case, yes. Looking at the docs `targeting` is `the targeting data used to configure, serve and track video ads`.

olusoladavid: Totally agree that the generic `targeting` name could use more specificity. Given the growing consensus, would it make sense to deprecate the old name for a minor version update (while still merging in its object when present), rename the option to something like `advertTargeting` and then completely drop support for `targeting` in a future major release?

frshwtr: I agree these can be considlated. We can possibly release under a minor release following an expand/contract pattern providing this PR is updated to handle both prior and new interface. And a major PR can be introduced later on to remove the old interface.

olusoladavid (approved):
Great work! Thanks for the code improvements too

andrewgeorgiou1981 (approved):
Looks great to me 👍

components/o-video/README.md
oliverturner: Just a small thing, but my preference would for `ads` as a more consistent prefix, so `adsTargeting`

components/o-video/src/js/ads.js
oliverturner: Personally I'd opt for robustness here...
```js
const paramsObject = {
  sz: this.video.targeting.sizes,
  unviewed_position_start: 1,
  scp: encodeURIComponent(targeting),
}

const { customParams } = this.video.advertTargeting
if(customParams && Object.keys(customParams).length) {
  paramsObject.cust_params = new URLSearchParams(customParams).toString()
}
```

AniaMakes: Nice, that'll do the trick. Thank you!

components/o-video/README.md
```diff
@@ -96,11 +96,12 @@ Where `opts` is an optional object with properties
 - `showCaptions` `[Boolean]` whether or not to add captions to the video. Defaults to `true`.
 - `data` `[Object]` JSON object representing a [response from next-media-api](https://next-media-api.ft.com/v1/eebe9cb5-8d4c-3bd7-8dd9-50e869e2f526). If used, the component will not make a call to the API and use this data instead.
 - `systemcode` `[String]` a valid [Bizops system code](https://biz-ops.in.ft.com/list/Systems) for the project using `o-video`.
-- `targeting` `[Object]` object containing the targeting data used to configure, serve and track video ads. If targeting is passed as an option but is missing properties, it will use the properties set in the `defaultTargeting` object instead. here is what each property on the targeting object affects:
+- `advertTargeting` (previously `targeting`) `[Object]` object containing the targeting data used to configure, serve and track video ads. If targeting is passed as an option but is missing properties, it will use the properties set in the `defaultTargeting` object instead. here is what each property on the targeting object affects:
```
notlee, line 99: Looks great :) Just need to update the readme with your latest changes 

```suggestion
- `adsTargeting` (previously `targeting`) `[Object]` object containing the targeting data used to configure, serve and track video ads. If targeting is passed as an option but is missing properties, it will use the properties set in the `defaultTargeting` object instead. here is what each property on the targeting object affects:
```

AniaMakes: of course there was one that I missed 🤦🏻

notlee: Oh! To avoid this being breaking could we also fall back to the deprecated `.targeting` value?

AniaMakes: It's not a breaking change; it uses `targeting` if `adsTargeting` isn't available https://github.com/Financial-Times/origami/pull/1703/files#diff-e150b16d129a24eb92461b88ce5b483c2c7b1177da804d757f79395d132e9c8cR195 (unless you meant something else - ie in the readme context?)

notlee (approved):

components/o-video/README.md
notlee: Ah sorry I missed that line 😅 Approved! 🙌

PR #1705, opened 2024-06-11
docs: introduce grid guidelines

frshwtr:


apps/for-everyone-website/src/components/grid/GridAnatomy.astro
```diff
@@ -0,0 +1,97 @@
+---
+---
+
+<svg
```
notlee, line 4: This needn't be an Astro component since it's not branded, and doesn't include a live preview. It could just be a `.svg` file which might be a but easier to manage in terms of running optimisations etc. over `.svg`

apps/for-everyone-website/src/content/docs/components/grid.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+## Grid Behaviours
+
+Our grid uses 100% of the screen’s width until the max width of 1220px is reached. This allows for maximum use of screen real estate, in complex screens and readability of page content in large and high-definition screens.
+
+### Contained Width
+
+Contained feature keeps the max width of the grid at 1220px. After this, the margins grow to fill the screen. Exceptions can be made for contents, e.g full-bleed banners and graphics, to break out of the contained grid.
+
+![](/assets/images/components/grid/contained-width.png)
```
notlee, line 96: Could these be svg? It would render text much more clearly

apps/for-everyone-website/src/content/docs/components/grid.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+## Dimensions
+
+<table>
+	<thead>
+	<tr>
+		<th>Breakpoint</th>
+		<th>Range (px)</th>
+		<th>columns</th>
+		<th>Margin</th>
```
notlee, line 26: Do we mean gutters or margin here? Further down we mention that margins keep increasing when the grid reaches the contained width, so a 64px margin would only make sense given a viewport of 1348px (1220 + 64 + 64)

apps/for-everyone-website/src/content/docs/components/grid.mdx
```diff
@@ -0,0 +1,131 @@
...
+		<td>1220+</td>
+		<td>12</td>
+		<td>64px</td>
+	</tr>
+	</tbody>
+</table>
+
+### Breakpoints
+
+Breakpoints are defined as min-width attributes. This means that screen sizes in between two defined breakpoint dimensions will inherit all dimensions of the breakpoint of smaller size.
```
notlee, line 65:
```suggestion
```

Do we need this line? This reads like it's for an engineer, when we don't have any CSS in place yet. I don't think we should enforce `min-width`. Media queries support range comparisons now, for example. 

The below break-down is clear from a design guideline perspective.

apps/for-everyone-website/src/content/docs/components/grid.mdx
```diff
@@ -0,0 +1,131 @@
...
+Our grid adapts to the following breakpoints to accommodate various screen sizes:
+
+* **Default (XS)**: 320px-489px. Default/smallest screen.
+* **Small (S)**: 490-739px. For larger mobile & small tablets.
+* **Medium (M)**: 740-979px. For tablets.
+* **Large (L)**: 980-1219px. For large tablets and small desktops.
+* **Extra Large (XL)**: 1220px and above. For desktop screens.
+
+
+However, you could define custom settings to set specific page size/or view type to use a design outside it’s breakpoint. For instance, the landscape width of the iPad pro is 1366px, which falls within the XL breakpoint. You can set this to maintain the X breakpoint design instead by indicating to your engineer.
```
notlee, line 76: 👍 I'd like to revisit and expand on this when we create supporting tools to direct people away from relying on breakpoints at all. In favour of a content first, "intrinsic layout" kinda approach

apps/for-everyone-website/src/content/docs/components/grid.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+### Columns
+
+![](/assets/images/components/grid/columns.png)
+
+Column widths are fluid and change with the size of the grid.
+
+### Baseline Grid
+
+The legacy grid uses a 4px baseline. We have not moved this yet because we are deciding between 4px and 8px baseline. This decision affects typography and spacing, so it will be considered during the typography and spacing optimisation.
```
notlee, line 86: Is the detail of our deliberations appropriate here? We have a 4px baseline grid and until that's changed, that's what people should use.

```suggestion
We use a 4px baseline which the grid, spacing, and typography is based on.
```

I wonder if there's a better place we can capture decision making or ongoing conversations? Feels like a Confluence thing

apps/for-everyone-website/src/content/docs/components/grid.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+## Grid Behaviours
+
+Our grid uses 100% of the screen’s width until the max width of 1220px is reached. This allows for maximum use of screen real estate, in complex screens and readability of page content in large and high-definition screens.
+
+### Contained Width
+
+Contained feature keeps the max width of the grid at 1220px. After this, the margins grow to fill the screen. Exceptions can be made for contents, e.g full-bleed banners and graphics, to break out of the contained grid.
```
notlee, line 94: These 2 sections repeat themselves. The first version omits an important note "exceptions can be made". Are we able to combine these two sections to reduce repetition and increase certainty around the exception?

```suggestion
## Contained Width Behaviour

Our grid uses 100% of the screen’s width until the max width of 1220px is reached. After this, the margins grow to fill the screen. This allows for maximum use of screen real estate, in complex screens and readability of page content in large and high-definition screens.

Exceptions can be made for contents, e.g full-bleed banners and graphics, to break out of the contained grid.
```

apps/for-everyone-website/src/content/docs/components/grid.mdx
```diff
@@ -0,0 +1,131 @@
...
+	<p slot="description">
+		![](/assets/images/components/grid/usage-best-practise.png)
+		Use the gutter width to help determine the space between layout regions. This space can be larger or smaller to help
+		group regions hierarchically on the page.
+	</p>
+</Guideline>
+
+<Guideline summary="Aligning content to grid" do={true}>
+	<p slot="description">
+		Create a container around the sub-items in a page e.g ads banners to align with the grid columns, avoid having a
```
notlee, line 128: I'm struggling to understand what this means, could it be reworded for clarity or perhaps enhanced with a visual aide available?

notlee (approved):

PR #1725, opened 2024-06-18
feat: remove current edition and change cross style

danieleruiz:
## Describe your changes
Due to the redesign initiative of the search bar navigation, the current edition will become part of the "edition-switcher" component of [dotcom-page-kit](https://github.com/Financial-Times/dotcom-page-kit). 

Changes:

- Move the close icon to the left side
- Remove the logo
- Update the style of the edition switcher
- Modify the style of the search bar and search button
- Change the placeholder text

https://financialtimes.atlassian.net/browse/CON-3426
## Link to Figma designs
https://www.figma.com/design/L6qyUFPPLRmf59OxTkz2VJ/Semantic-Search?node-id=279-91827&m=dev
# Screenshots:
**Before:**
image

**After:**
image


components/o-header/test/search.test.js
```diff
@@ -45,7 +45,7 @@ describe('Header with search', () => {
 					<div class="o-header__container">
 						<form class="o-header__search-form" action="/search" role="search" aria-label="Site search">
 							<label class="o-header__visually-hidden" for="o-header-search-term-js">Search the <abbr title="Financial Times">FT</abbr></label>
```
frshwtr, line 47: The text here should be updated to match the placeholder text

danieleruiz: updated

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -144,11 +162,12 @@
...
-		height: 32px;
+		height: 40px;
 		padding: 8px (div($_o-header-drawer-padding-x, 2));
 		border: 1px solid;
 		// prevent zoom on focus
 		font-size: 100%;
+		line-height: 40px;
 	}
 
 	.o-header__drawer-search-term {
```
notlee, line 173: This is an improvement to the submit button but it lacks hover and active states. Could we replace this with [o-buttons](https://o2-core.origami.ft.com/?path=/story/components-o-buttons--big-button&args=icon:search;type:primary;iconOnly:!true&globals=backgrounds:!undefined)? The [oButtonsContent](http://localhost:6006/?path=/docs/components-o-buttons-sassdoc--docs&globals=backgrounds:!undefined#obuttonscontent) mixin allows us to apply button styles to a custom class. 

Something like

```scss
.o-header__drawer-search-submit {
    @include oButtonsContent($opts: (
        'type': 'primary',
         'icon-only': true,
        'icon': 'search'
    ));
}
```

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -164,6 +183,7 @@
 		// margin copied from o-forms since introducing a new dependency would
 		// be a breaking change, this colour could go in o-colors as a usecase
```
notlee, line 184: Should we reconsider the form dependency? The design team are working on some fundamental design language changes for later this year which will impact forms – think shape and colour. By adding forms as a dependency we:
1. Can be more confident of consistency today.
1. Know where forms are used, and can check changes work in this context.
1. Design changes will flow apply via the dependency, without us needing to remember to update multiple components

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -233,6 +253,13 @@
 	.o-header__drawer-menu-item {
 		margin-top: 1px;
 	}
+	.o-header__drawer-divider {
+		margin-top: 16px;
+		width: 1px;
+		height: 16px;
```
notlee, line 259: Is this height intended to create margin? Using `margin` would be more explicit and therefore easier to read.

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -233,6 +253,13 @@
...
+		background-color: #333333;
```
notlee, line 260: This grey is not part of our colour palette, could we use a [black tint](https://o2-core.origami.ft.com/?path=/story/components-o-colors-color-mixer--color-mixer&globals=backgrounds:!undefined) instead?

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -164,6 +183,7 @@
 		// margin copied from o-forms since introducing a new dependency would
 		// be a breaking change, this colour could go in o-colors as a usecase
 		border: 1px solid oColorsByName('black-50');
+		margin-right: 10px;
```
notlee, line 186: Please stick to our [standard spacing scale](https://o2-core.origami.ft.com/?path=/docs/components-o-spacing-readme--docs&globals=backgrounds:!undefined). This could be `oSpacingByName('s1')` (12px). 

(o-header is an old component and is full of negative examples)

components/o-header/src/scss/features/_drawer.scss
danieleruiz: done

components/o-header/src/scss/features/_drawer.scss
danieleruiz: done

components/o-header/src/scss/features/_drawer.scss
danieleruiz: done

components/o-header/src/scss/features/_drawer.scss
danieleruiz: this line is to make the divider line visible:
Captura de pantalla 2024-07-02 a las 13 17 37

components/o-header/src/scss/features/_drawer.scss
notlee: What styles would you need to override? Perhaps the design is off

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -164,21 +182,20 @@
...
-			@include oIconsContent('search', oColorsByName('white'), $size: $_o-header-icon-size);
-			content: '';
-			margin-top: -(div($_o-header-drawer-padding-y, 2));
-		}
+		@include oButtonsContent($opts: (
+			'type': 'primary',
+			'icon-only': true,
+			'icon': 'search'
+		));
+		background-size: contain;
```
notlee, line 198: We should be able to remove `background-size: contain;` here by setting the button size `'size': 'big'` and removing the styles for `.o-header__drawer-search-submit` above, then we'll be relying on standardised button styles without custom css

danieleruiz: done! thank you!

components/o-header/src/scss/features/_drawer.scss
danieleruiz: o-forms had some styles that only applies if are inside a specific form element so if i include only the input clases somethimes is not 100% the same, anyways im using a common mixin for the button and changing the height of the  input. so i dont think that worth it to include a dependency and new clases with need parents etc

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -71,30 +71,58 @@
...
+
+			.o-header__drawer-menu-item {
+				padding-top: 12px;
+				padding-right: 12px;
+				font-size: 18px;
+			}
+		}
 	}
 
 	.o-header__drawer-tools-logo {
```
notlee, line 104: I believe this class is redundant now, can we safely delete it?

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -71,30 +71,58 @@
...
+		padding: ($_o-header-drawer-padding-y * 2.5) 0 0 $_o-header-drawer-padding-x;
 		background-color: _oHeaderGet('drawer-tools-background');
 		color: _oHeaderGet('drawer-tools-text');
+		display: flex;
+		flex-direction: row;
+		justify-content: space-between;
+		align-items: center;
+
+		.o-header__drawer-edition-switcher {
+			order: 2;
```
notlee, line 83: Replying on `order` here means the [tab order](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html) is a little surprising. Could we reorder the markup instead of relying on CSS `order` here?

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -71,30 +71,58 @@
...
 		// dividing by 1px removes the unit
 		@include oIconsContent('cross', _oHeaderGet('drawer-tools-close'), $size: $_o-header-icon-size-large);
 		box-sizing: border-box;
-		float: right;
 		border: 0;
 		opacity: 0.75;
-		cursor: pointer;
-		margin-top: -10px;
+		order: 1;
+		margin-left: -10px;
```
notlee, line 124: This negative margin feels a bit hacky and difficult to maintain. Could we align the padding of the parent elements instead? `o-header__drawer-tools` `o-header__drawer-search` `o-header__drawer-actions` should all have the same horizontal padding.

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -71,30 +71,58 @@
...
+			display: flex;
+
+			.o-header__drawer-menu-link {
+				padding-left: 12px;
+				text-decoration: underline;
+			}
+
+			.o-header__drawer-current-edition {
+				font-weight: 600;
+				color: var(--Black-Tints-Black-60, #66605C);
```
notlee, line 93: This custom property doesn't exist.

```suggestion
				color: oColorsByName('black-60');
```

notlee (changes requested):
The vertical spacing on desktop search looks a little off, there is more space above than below:

Screenshot 2024-07-03 at 16 35 58

components/o-header/src/scss/features/_drawer.scss
danieleruiz: removed

components/o-header/src/scss/features/_drawer.scss
danieleruiz: the paddings are the same, the problem here is that the background of "x" close button have like 10px of margin centered on a svg file so to make it be aligned i had to add this negative margin

components/o-header/src/scss/features/_drawer.scss
```diff
@@ -71,30 +71,43 @@
 	//
 	.o-header__drawer-tools {
 		overflow: hidden;
-		padding: ($_o-header-drawer-padding-y * 1.5) 0 $_o-header-drawer-padding-y $_o-header-drawer-padding-x;
+		padding: ($_o-header-drawer-padding-y * 2.5) 0 0 $_o-header-drawer-padding-x;
```
notlee, line 74: Is the top padding too high here? The close button should align with where the burger menu was
Screenshot 2024-07-04 at 13 03 23

danieleruiz: changed !

danieleruiz:
> The vertical spacing on desktop search looks a little off, there is more space above than below:
> 
> Screenshot 2024-07-03 at 16 35 58

fixed it

notlee (approved):

PR #1730, opened 2024-06-21
fix: fix assign of options in conjunction with data-atributtes

juanSanchezAlcala:
Fix options initialisation along with data-attributes
We were choosing between options or data-attributes instead of overriding and assigning on top of each other.
So the preference now is options passed programmatically have preference over data-attributes and overrides values got from it but don't remove the values of other options assign by data-attributes


AliceArmstrong (approved):
Looks good, are there any tests to update?

frshwtr (approved):

PR #1732, opened 2024-06-21
feat: add tracking and optional links for subscribe link and login link

juanSanchezAlcala:
Add tracking and allow to pass links to subscribe and login. In case we don't pass that property the text will be shown without links

[ticket](https://financialtimes.atlassian.net/browse/CI-1493)

### DIFF
--- a/components/o-comments/src/js/stream.js
+++ b/components/o-comments/src/js/stream.js
@@ -260,19 +260,13 @@ class Stream {
 				}
 
 				if (mappedEvent.oTracking && !this.options.disableOTracking) {
-					const oTrackingEventOptions = {
-						bubbles: true,
-						detail: {
-							...defaultDetailWithContentAdded
-						}
-					};
+					
 
 					if (error) {
-						oTrackingEventOptions.detail.error = error;
+						defaultDetailWithContentAdded.error = error;
 					}
 
-					const oTrackingEvent = new CustomEvent('oTracking.event', oTrackingEventOptions);
-					document.body.dispatchEvent(oTrackingEvent);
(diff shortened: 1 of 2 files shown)


components/o-comments/src/js/stream.js
```diff
@@ -317,28 +311,84 @@ class Stream {
...
 				<p>
-					View our full <a href='https://subs.ft.com/products'>subscription packages</a> to join the conversation.
+					View our full ${this.options.linkSubscribe ? `<a class="linkSubscribe" href='${this.options.linkSubscribe}'>subscription packages</a>` : `subscription packages` } to join the conversation.
 				</p>
 			`;
 			customMessageContainer.innerHTML = this.isTrialist ? messageForTrial : this.isRegistered ? messageRegistered : messageForAnonymous;
+			// this content is attached after oTracking.init is called therefore set data-trackable to the links doesn't work, therefore we raise an event when click on it
+			customMessageContainer.querySelector('.linkSubscribe')?.addEventListener('click', (event) => {
+				event.preventDefault();
+				const trackData = {
```
debugwand, line 333: this object and the one below could probably have 1 base object and then use object assign to add in the linkType for their own objects, which is the only bit that differs

umbobabo: Yeah, we could add the base object [here](https://github.com/Financial-Times/origami/pull/1732/files#diff-f9b3e1aac4c3bf3a99d472de9df47b5c6736a8ff5de7f20fd4ebde9d13c35a17R389)

umbobabo (approved):

PR #1762, opened 2024-07-26
fix: Make the toggle tip component inline.

notlee:
This removes excess space below the toggle tip and allows it to be positioned correctly in flow.

It will then be possible to remove these overrides with a couple extra steps for the product. Remove the `div` with the brand attribute, and apply it to an existing parent element. Use the `gap` property and o3 CSS Custom Properties for spacing on the partner content title element:
https://github.com/Financial-Times/next-article/blob/5c2eb8785dfcf7e11341dec1d6d9b5c4395658ae/client/partner-content.scss#L164 

Before: 

Screenshot 2024-07-26 at 10 54 06

After:

Screenshot 2024-07-26 at 10 53 50

### DIFF
--- a/components/o3-tooltip/main.css
+++ b/components/o3-tooltip/main.css
@@ -10,7 +10,7 @@
 
 o3-tooltip-toggle.o3-tooltip {
 	position: relative;
-	display: block;
+	display: inline-flex;
 }
 
 .o3-tooltip > .o3-tooltip-wrapper {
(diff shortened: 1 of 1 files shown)


components/o3-tooltip/main.css
```diff
@@ -10,7 +10,7 @@
 
 o3-tooltip-toggle.o3-tooltip {
 	position: relative;
-	display: block;
+	display: inline-flex;
```
notlee, line 13: This is the change. The rest is Prettier

akomiqaia (approved):

PR #1763, opened 2024-07-29
feat: new color to attend ad slots

fabioleiteparserft:
## Describe your changes
Including a new `wheat-subtle` color in the palette file to be used primarily as a background color for ad slots.

https://financialtimes.atlassian.net/browse/ADSDEV-1961

### DIFF
--- a/components/o-colors/src/scss/_palette.scss
+++ b/components/o-colors/src/scss/_palette.scss
@@ -28,6 +28,7 @@ $_o-colors-default-palette-colors: join((
 		('crimson', #cc0000),
 		('graphics-dark-blue', #006f9b),
 		('wheat', #f2dfce, ('allow-tones': true)),
+		('light-wheat', #fcede0),
 		('candy', #ff7faa),
 		('wasabi', #96cc28),
 		('jade', #00994d),
--- a/components/o-colors/src/scss/_palette.scss
+++ b/components/o-colors/src/scss/_palette.scss
@@ -28,6 +28,7 @@ $_o-colors-default-palette-colors: join((
 		('crimson', #cc0000),
 		('graphics-dark-blue', #006f9b),
 		('wheat', #f2dfce, ('allow-tones': true)),
+		('wheat-subtle', #fcede0),
 		('candy', #ff7faa),
 		('wasabi', #96cc28),
 		('jade', #00994d),
--- a/components/o-colors/src/scss/_palette.scss
+++ b/components/o-colors/src/scss/_palette.scss
@@ -28,7 +28,6 @@ $_o-colors-default-palette-colors: join((
 		('crimson', #cc0000),
 		('graphics-dark-blue', #006f9b),
 		('wheat', #f2dfce, ('allow-tones': true)),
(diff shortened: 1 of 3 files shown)


notlee:
Hey @fabioleiteparserft 👋 Thanks for the PR. Adding new colours to our palette is rare, and usually involves the brand team. Could you share:

1. Where this colour will be used in more detail (designs / screenshots?)
2. Is this colour derived from others in the colour palette?

fabioleiteparserft:
> Hey @fabioleiteparserft 👋 Thanks for the PR. Adding new colours to our palette is rare, and usually involves the brand team. Could you share:
> 
> 1. Where this colour will be used in more detail (designs / screenshots?)
> 2. Is this colour derived from others in the colour palette?

Hello @notlee 

1. I've added a screenshot as an example where we'll use the new color.
2. No, it isn't.

fabioleiteparserft:
@notlee Now using the use-case approach. Feel free to check 👍

components/o-colors/src/scss/_palette.scss
```diff
@@ -28,6 +28,7 @@ $_o-colors-default-palette-colors: join((
 		('crimson', #cc0000),
 		('graphics-dark-blue', #006f9b),
 		('wheat', #f2dfce, ('allow-tones': true)),
+		('light-wheat', #fcede0),
```
notlee, line 31: Let's avoid adding to the palette. This suggests it's a colour that can be used more widely. The palette is rarely added to and co-ordinated with the brand team....
```suggestion
```

components/o-colors/src/scss/_use-cases.scss
```diff
@@ -21,6 +21,7 @@
 		('hero', ('background': 'wheat')),
 		('hero-opinion', ('background': 'oxford')),
 		('hero-highlight', ('background': 'claret')),
+		('ads', ('background': 'light-wheat')),
```
notlee, line 24: Instead, let's add a specific usecase for ads. We'll include a comment to explain how it's derived from the current palette. This makes it clear that, in affect, we're creating a colour that acts as wheat with opacity on an paper background. ([original use](https://github.com/Financial-Times/origami/tree/00b35280286192d08dcf564dbb21738bf8da02d8/components/o-colors#usecases))
```suggestion 
		('ads', ('background': #fcede0)), // oColorsMix(wheat, paper, 20);
```

components/o-colors/src/scss/_palette.scss
fabioleiteparserft: @notlee  I see your point, but I couldn't add the use case without this, because we can't add the HEX color directly or even use the oColorsMix function. I was wondering what would be the best solution in this case.

notlee: A hex value here should work, can I check you didn't use quotes? We want Sass to interoperate the value as a [color](https://sass-lang.com/documentation/values/colors/) and not a [string](https://sass-lang.com/documentation/values/strings/).

notlee (approved):

PR #1764, opened 2024-07-30
feat: introduce o3 grid CSS custom properties

frshwtr:


components/o3-foundation/grid.css
```diff
@@ -0,0 +1,30 @@
...
+	@media screen and (min-width: 980px) {
+		--o3-foundation-grid-columns: 12;
+		--o3-foundation-grid-gap: var(--o3-spacing-2xl);
+		--o3-foundation-grid-margin: var(--o3-spacing-2xl);
+	}
+
+	@media screen and (min-width: 1220px) {
+		--o3-foundation-grid-columns: 12;
+		--o3-foundation-grid-gap: var(--o3-spacing-2xl);
+		--o3-foundation-grid-margin: var(--o3-spacing-2xl);
```
notlee, line 28: Looks like a good start, nice use of custom properties here.

This doesn't yet capture [contained width behaviour](https://origami-beta.ft.com/guides/grid/#contained-width-behaviour). In the past we've provided a CSS class to set a max width, and put the grid within it. However we know from our design guidelines that content can break out of this container. Our old approach forced users to layout with negative margins, which are somewhat hacky and difficult to work with. 

I wonder if instead we're able to use the grid template to contain those 12 columns, and have 2 extra columns (1 either side) to represent the infinite margins on desktop.

components/o3-editorial-typography/stories/core/grid.stories.tsx
```diff
@@ -0,0 +1,38 @@
+import type {Meta} from '@storybook/react';
```
frshwtr, line 1: This isn't needed for the grid. It can be removed or potentially tidied up to demo grids if need be.

components/o3-editorial-typography/stories/core/grid.css
```diff
@@ -0,0 +1,27 @@
+
```
frshwtr, line 1: Likewise, this isn't necessary.

components/o3-foundation/stories/grid.tsx
```diff
@@ -0,0 +1,78 @@
...
+		text: 'Bleed only left',
+	},
+];
+
+export function O3Grid() {
+	// on window resize update the grid
+	const [gridItems, setGridItems] = useState(itemDetails);
+
+	useEffect(() => {
+		const handleResize = () => {
```
frshwtr, line 41: When it comes to users of o3 grid, are they going to need to do something similar to this to handle content flow on resizes?

components/o3-foundation/grid.css
```diff
@@ -0,0 +1,35 @@
...
+	--o3-grid-template: 0px
+		repeat(var(--o3-grid-columns), 1fr) 0px;
+	--o3-grid-area: 'spread-left content-start . . content-end spread-right';
+	--o3-grid-columns: 4;
+	--o3-grid-gap: 16px;
+
+	@media screen and (min-width: 740px) {
+		--o3-grid-template: 8px
+			repeat(var(--o3-grid-columns), 1fr) 8px;
+		--o3-grid-area: 'spread-left content-start . . . . . . content-end spread-right';
```
notlee, line 11: There are some issues to work out here still.

On wider viewports the content area is larger than it should be by a few px.

Screenshot 2024-08-08 at 16 49 36

And on a tablet-ish viewport we end up with 8 columns between content grid lines rather than 12
Screenshot 2024-08-08 at 16 49 59

notlee (changes requested):

notlee:
There's no JSX so let's delete this page? It doesn't add anything the story doesn't provide.

Screenshot 2024-08-08 at 16 53 30

We should improve documentation with detailed technical guides in the README and a summary with link on the website.

components/o3-foundation/grid.css
akomiqaia: we have 8 columns on that screen size. https://origami-for-everyone.ft.com/guides/grid/ we get 12 on 980px.

akomiqaia: And there is some rounding issue because at 1220px (code media query is 1268px because of margins) screen we set the width of the column using calc and it's not quite divisible perfectly.

akomiqaia:
> ae01cbe

[ae01cbe](https://github.com/Financial-Times/origami/pull/1764/commits/ae01cbe327dd658dd494226e0291083aa796310b)

components/o3-foundation/grid.css
notlee: Ah sorry, I didn't realise we have 8 as an intermediate number of columns 👍

notlee: I've asked Mary whether there was a strong rationale for the 1220px container. I'd prefer to have an exact container width and equal columns when the contained width is reached (but it doesn't matter too much, I don't think). We might want to choose a different container size. Definitely for a separate PR if so though

_____

Current o-grid:
1180px container with 12 columns, 11 gutters, 20px gutter: (1180 - (11 * 20)) / 12 = 76px columns

New o3 guidelines don't divide well at contained width :grimacing:
1220px container with 12 columns, 11 gutters: (1220 - (11 * 24)) / 12 = 79.66666666666666px columns

1200px or 1260px work better with the same gutter:
1200px container with 12 columns, 11 gutters: (1200 - (11 * 24)) / 12 = 76px columns
1260px container with 12 columns, 11 gutters: (1260 - (11 * 24)) / 12 = 83px columns

notlee:
Thanks @akomiqaia! Could you also add to the foundations readme to document the grid including the named grid lines? 

Suggestion: Might bleed start / end be easier to remember? Since it aligns with the wording we use in design guidelines and content start/end?

```diff
-bleed-start content-start . . . . . . content-end bleed-end
+spread-left content-start . . . . . . content-end spread-right
```

components/o3-foundation/grid.css
```diff
@@ -0,0 +1,35 @@
...
+	@media screen and (min-width: 980px) {
+		--o3-grid-template: 0px
+			repeat(var(--o3-grid-columns), 1fr) 0px;
+		--o3-grid-area: 'spread-left content-start . . . . . . . . . . content-end spread-right';
+		--o3-grid-columns: 12;
+		--o3-grid-gap: 24px;
+	}
+
+	@media screen and (min-width: 1268px) {
+		--column-width: calc((1220px - 11 * var(--o3-grid-gap)) / 12);
```
notlee, line 24: 💡It might be handy to have a public `--o3-grid-column-width` available at all viewports? It would allow us to align elements to grid columns when they're nested in other elements and can't be positioned with the actual CSS Grid. For example Deyan suggested form elements should be x columns wide by default

```
input {
    min-width: calc(4 * var(--o3-grid-column-width));
}
```

Thoughts?

components/o3-foundation/grid.css
```diff
@@ -0,0 +1,35 @@
...
+		--o3-grid-template: 1fr
+			repeat(var(--o3-grid-columns), var(--column-width)) 1fr;
+	}
+}
+
+.o3-grid {
+	display: grid;
+	grid-template-columns: var(--o3-grid-template);
+	grid-template-areas: var(--o3-grid-area);
+	gap: var(--o3-grid-gap);
```
notlee, line 34: We don't want to force this gap between rows 
```suggestion
	column-gap: var(--o3-grid-gap);
```

components/o3-foundation/grid.css
akomiqaia: This is a good thought. We should consider making a ticket for it.

components/o3-foundation/README.md
```diff
@@ -121,12 +124,73 @@ Example:
...
-	margin: var(--o3-spacing-2xs);
+ margin: var(--o3-spacing-2xs);
 }
 ```
 
 See our documentation website for a [full list of spacing tokens](https://origami-for-everyone.ft.com/guides/spacing/).
 
+### Grid
+
+o3-foundation provides CSS Custom Properties for grid that you can set on your class:
```
notlee, line 135: Consider the order of this by the message you want to convey, since it's likely the reader will be glancing through rather than reading from top to bottom.

It's impossible to use the grid effectively without understand the design guidelines. A brief introduction and link to the design guidelines would be a better way to open: https://origami-beta.ft.com/guides/grid/

If `o3-grid` is the convenient option, perhaps share that first and the detail of custom properties later if needed.

components/o3-foundation/README.md
```diff
@@ -121,12 +124,73 @@ Example:
...
+</div>
+```
+
+The `o3-grid` template follows a specific structure:
+
+```txt
+`bleed-start content-start [REST_OF_THE_COLUMNS] content-end bleed-end`,
+```
+
+- The 12-Column Grid Structure: The `o3-grid` system is built on a 12-column grid structure. These 12 columns are utilized within the `content-start` and `content-end` boundaries.
```
notlee, line 163: This isn't true on mobile and tablet like viewports. We could perhaps say something like "the number of columns is variable according to viewport size, see design guidelines for details (link)". This highlights important information to know without repeating details which could get outdated more easily

components/o3-foundation/README.md
```diff
@@ -121,12 +124,73 @@ Example:
...
+- Bleed Columns for Extended Layouts: `bleed-start` & `bleed-end`: These special columns extend beyond the 12-column grid, reaching the edges of the viewport. This allows you to create "full-bleed" layouts where content extends off the screen. These columns are used as margin areas, providing visual breathing room around your central content.
```
notlee, line 164: 👍 Sounds good. Could even link to the relevant design guidelines here

notlee (approved):

PR #1769, opened 2024-08-08
Or 647/develop checkbox

akomiqaia:


components/o3-form/src/tsx/FormAtoms.tsx
```diff
@@ -0,0 +1,26 @@
...
+    </div>
+  );
+};
+
+export const FeedBack = ({ message, type }: FeedBackProps) => {
+  return <div className={`feedback ${type}`}>{message}</div>;
+};
+
+
+export const CheckBox = ({label, checked, error}: CheckBoxProps) => {
```
frshwtr, line 19: Can components like `CheckBox` and `TextInput` be kept in separate files from the atoms that make up a form field? It will help create a semantic separation/difference between a form field component with a child and the components that make up form fields.

components/o3-form/src/tsx/FormAtoms.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+	return (
+		<FormField label={label}>
+			<input type="checkbox" checked={checked} className={error ? 'o3-form-error' : undefined} />
```
frshwtr, line 23: IIRC we wanted error classes to be set on the parent of this child, and styling gets passed down?

PR #1771, opened 2024-08-13
docs: Add storybook commands for o3 and apps/for-everyone-website 

AnnaDraganova:
## Describe your changes

- Updated the root README to include commands for running Storybook for Origami's o3 components and the Origami Design System website


notlee (approved):
Thanks :D

PR #1776, opened 2024-08-13
feat: introduce `o3-grid-column-width` CSS Custom Property

frshwtr:
## Describe your changes

Some components outside of grids may use grid dimensions. For example, form elements may use `max-width` constrained by a number of columns in grid, without being part of one.

This PR adds CSS custom properties at each breakpoint to give components outside of a grid some values to work with.

### DIFF
--- a/components/o3-foundation/grid.css
+++ b/components/o3-foundation/grid.css
@@ -1,4 +1,5 @@
 :root {
+	--_o3-grid-gutters: calc(var(--o3-grid-columns) - 1);
 	--o3-grid-template: 0px
 		repeat(var(--o3-grid-columns), 1fr) 0px;
 	--o3-grid-area: 'bleed-left content-start . . content-end bleed-right';
@@ -10,6 +11,7 @@
 			repeat(var(--o3-grid-columns), 1fr) 8px;
 		--o3-grid-area: 'bleed-left content-start . . . . . . content-end bleed-right';
 		--o3-grid-columns: 8;
+		--o3-grid-column-width: calc((740px - var(--_o3-grid-gutters) * var(--o3-grid-gap)) / var(--o3-grid-columns));
 	}
 
 	@media screen and (min-width: 980px) {
@@ -18,12 +20,14 @@
 		--o3-grid-area: 'bleed-left content-start . . . . . . . . . . content-end bleed-right';
 		--o3-grid-columns: 12;
 		--o3-grid-gap: 24px;
+		--o3-grid-column-width: calc((980px - var(--_o3-grid-gutters) * var(--o3-grid-gap)) / var(--o3-grid-columns));
 	}
 
 	@media screen and (min-width: 1268px) {
-		--column-width: calc((1220px - 11 * var(--o3-grid-gap)) / 12);
+		--o3-grid-columns: 12;
 		--o3-grid-template: 1fr
-			repeat(var(--o3-grid-columns), var(--column-width)) 1fr;
+			repeat(var(--o3-grid-columns), var(--o3-grid-column-width)) 1fr;
+		--o3-grid-column-width: calc((1220px - var(--_o3-grid-gutters) * var(--o3-grid-gap)) / var(--o3-grid-columns));
 	}
 }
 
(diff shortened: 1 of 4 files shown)


components/o3-foundation/grid.css
```diff
@@ -10,6 +11,7 @@
 			repeat(var(--o3-grid-columns), 1fr) 8px;
 		--o3-grid-area: 'bleed-left content-start . . . . . . content-end bleed-right';
 		--o3-grid-columns: 8;
+		--o3-grid-column-width: calc((740px - var(--_o3-grid-gutters) * var(--o3-grid-gap)) / var(--o3-grid-columns));
```
frshwtr, line 14: Unsure if a hard value of `740px` is going to work here. It's going to need the px value for calculation but I don't see how I can get a fluid value outside of `vw` 🤔

notlee: It's not going to work because there's no 740px container, the grid is flexible. 

We should only need to set the two Custom Property twice. Once for the fluid grid, then once when it's fixed at 1220px.

I haven't thought about it deeply but: `vw` is closer to what we actually mean but I we might want to be careful of scrollbars or [dynamic viewports](https://12daysofweb.dev/2022/new-viewport-units/).

Alternatively, can we assume the grid is a [containing block ](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#effects_of_the_containing_block)and use 100%?
```
--o3-grid-column-width: calc((100% - var(--_o3-grid-gutters) * var(--o3-grid-gap)) / var(--o3-grid-columns));
```

notlee: Finally, maybe `--o3-grid-column-width` isn't actually what we want public! If our aim is to allow people to set widths of nested items, to align with grid columns, we need to account for gaps.

This is not a very fun API:
```
input {
    width: calc((var(--o3-grid-column-width) + var(--o3-grid-gap)) * 4)
}
```

But if we embed the grid gap the naming could be confusing? Maybe fine?
```
input {
    width: calc(var(--o3-grid-column-width) * 4)
}
```
Screenshot 2024-08-13 at 17 25 26

frshwtr: > Alternatively, can we assume the grid is a [containing block ](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#effects_of_the_containing_block)and use 100%?

Does this imply that any component using this var must be inside a grid? Is this going to be problematic with form elements which may not necessarily be in a grid?

notlee: >Does this imply that any component using this var must be inside a grid? Is this going to be problematic with form elements which may not necessarily be in a grid?

From a design perspective, the grid can makes sense when applied on a full width element.

notlee: >Is this example correct?

That depends on user expectations. See the screenshot. If you were working on a design that had 2 evenly spaced elements side by side, but you could use CSS grid because they were heavily nested for some reason, how would you want to lay them out?

frshwtr: I've introduced a new variable to allow this, it will need to be paired with a user-controlled var to work. Let me know what you think, I'll write docs once we've agreed an approach.

notlee: That's a great idea 👍 Naming these are hard but I like your initial suggestion.

These seem quite clearly connected: 
- `--o3-grid-span-columns`
- `--o3-grid-span-columns-count`


These seem more loosely associated by name: 
- `--o3-grid-span-columns`
- `--o3-grid-columns-to-span`

notlee: Maybe?

- `--o3-grid-span-columns-width`
- `--o3-grid-span-columns-count`

frshwtr: Adjusted, thanks :)

notlee:
Looks good 😄 Now just to document in the README and potentially add a demo to the Story?

components/o3-foundation/stories/grid.tsx
```diff
@@ -74,6 +74,10 @@ export function O3Grid() {
 					{item.text}
 				</div>
 			))}
+			<div style={{gridColumn: `content-start / span 6`, backgroundColor: 'tomato'}}>
+				Span 6
+				<div className='nested-element'>Nested Element</div>
```
notlee, line 79: Styles should be inline here so they are visible in Storybook's HTML tab

notlee (approved):
🙌

PR #1784, opened 2024-08-15
chore: add checkbox back after separating it from form-fields

akomiqaia:
[OR-647](https://financialtimes.atlassian.net/browse/OR-647)

[OR-647]: https://financialtimes.atlassian.net/browse/OR-647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/o3-form/src/tsx/CheckBox.tsx
```diff
@@ -1,21 +1,13 @@
-import {uidBuilder} from '@financial-times/o-utils';
-
 import {FormField, FormFieldset} from './fieldComponents/FormField';
-
 import type {CheckBoxProps, FormFieldsetProps} from '../types';
-
-const uniqueId = uidBuilder('o3-form');
+import React from 'react';
```
akomiqaia, line 3: We want to actively avoid importing react in our components since these components will need to be used in other JSX-supported environments and maybe we can find a way of not using React imports in TSX files.

components/o3-form/src/tsx/CheckBox.tsx
```diff
@@ -51,5 +42,14 @@ export const CheckBox = (props: CheckBoxProps) => {
 };
 
 export const CheckBoxGroup = (props: FormFieldsetProps) => {
-	return <FormFieldset {...props}>{props.children}</FormFieldset>;
+	const {children, ...restProps} = props;
+
+	return (
+		<FormFieldset {...restProps}>
+			{React.Children.map(children, child =>
```
akomiqaia, line 49: Would `props.children` work here instead? and maybe that will resolve the React import issue as well?

components/o3-form/stories/story-template.tsx
```diff
@@ -47,41 +44,50 @@ export const CheckBoxGroupStory: CheckBoxGroupStory = {
 		label: 'Check all that apply',
 		description: 'This is a description of the checkbox group',
 		feedback: {
+			checkboxIds: ['checkbox_5', 'checkbox_6'],
```
akomiqaia, line 47: feedback might be a generic component and specifying `checkboxIds` seems too specific. maybe we can update it to something more generic like `errorElementIds`? Just a suggestion 😆 maybe not a great name. 

Also our `BaseInputProps` has `error` attribute and maybe that can be passed down as an attribute. Maybe error state should be controlled by the input element itself?

components/o3-form/src/css/components/checkbox.css
```diff
@@ -40,6 +40,16 @@
...
+.o3-form-input__checkbox-label:focus-visible {
+	outline: none;
+	box-shadow: 0 0 0 var(--o3-spacing-5xs) #4D4845; /* TOKEN NEEDED */
+	border-radius: var(--o3-spacing-5xs); 
+	padding: var(--o3-spacing-5xs);
+	margin-top: var(--o3-spacing-4xs);
+	margin-left: calc(-1 * var(--o3-spacing-5xs));
+	margin-bottom: calc(-1 * var(--o3-spacing-5xs));
+}
+
```
akomiqaia: focus styles  should be applied automatically but if not we have a helper class defined here https://github.com/Financial-Times/origami/tree/main/components/o3-foundation#focus-rings

AnnaDraganova: Yes, the style was the same as in the design. Since the outline is not going to be rounded. I'll update this to adjust the space between the outline and the checkbox only.

components/o3-form/src/tsx/CheckBox.tsx
AnnaDraganova: The highlighted logic gives access to the feedback prop from the parent element to the children, in this case, it's the `CheckboxItem`. That's done so we can properly highlight the checkbox that's with an error. Maybe we can improve this logic. 🤔

AnnaDraganova: Logic is refactored in the following commit - https://github.com/Financial-Times/origami/pull/1784/commits/9a2af6ced867ac6a01ed01772250bbd2fd601bab

components/o3-form/src/tsx/fieldComponents/Feedback.tsx
```diff
@@ -0,0 +1,25 @@
...
+					fill="none"
+					xmlns="http://www.w3.org/2000/svg">
+					<path
+						id="Error"
+						fill-rule="evenodd"
+						clip-rule="evenodd"
+						d="M12.5 7.5C12.5 10.5376 10.0376 13 7 13C3.96243 13 1.5 10.5376 1.5 7.5C1.5 4.46243 3.96243 2 7 2C10.0376 2 12.5 4.46243 12.5 7.5ZM14 7.5C14 11.366 10.866 14.5 7 14.5C3.13401 14.5 0 11.366 0 7.5C0 3.63401 3.13401 0.5 7 0.5C10.866 0.5 14 3.63401 14 7.5ZM10 8.5V6.5L4 6.5V8.5H10Z"
+						fill="#CC0000"
+					/>
+				</svg>
```
AnnaDraganova, line 20: **note:** This should be changed after the token for the error icon is added.

components/o3-form/src/css/components/checkbox.css
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	&:indeterminate + label::after {
+		position: absolute;
+		content: '';
+		width: var(--o3-spacing-s);
+		height: var(--o3-spacing-s);
+		background-color: var(--o3-color-palette-white);
+		/* TOKEN NEEDED FOR MINUS SIGN*/
+		mask-image: url("data:image/svg+xml,%3Csvg width='24' height='24' viewBox='0 0 24 24' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Crect x='5' y='10' width='14' height='4' rx='2' fill='%23CC0000'/%3E%3C/svg%3E%0A");
```
AnnaDraganova, line 82: **note:** This should be changed after the token for the minus icon is added.

components/o3-form/src/tsx/fieldComponents/Feedback.tsx
```diff
@@ -0,0 +1,25 @@
+import type {FeedbackProps} from '../../types';
+
+export const Feedback = ({message, type}: FeedbackProps) => {
+	return (
+		<div className={`o3-form-feedback o3-form-feedback__${type}`}>
+			{type === 'error' ? (
+				<svg
```
frshwtr, line 7: Should this be a token?

components/o3-form/src/types/index.ts
```diff
@@ -5,33 +5,31 @@ export type BaseInputProps = {
 	label?: string;
 	description?: string;
 	optional?: boolean;
-	error?: string;
+	error?: boolean;
 	attributes?: JSX.IntrinsicElements['input'];
 };
 
+// CheckBox specific props
```
frshwtr, line 12: These comments here and on L15 can be removed.

components/o3-form/src/types/utils.ts
```diff
@@ -0,0 +1,15 @@
+declare module '@financial-times/o-utils' {
```
frshwtr, line 1: [utils.d.ts](https://github.com/Financial-Times/origami/blob/main/components/o3-form/src/types/utils.d.ts) already exists on main, we don't need this here.

components/o3-form/src/types/index.ts
AnnaDraganova: Done in the following commit - https://github.com/Financial-Times/origami/pull/1784/commits/74020b40d9f33cb9293a3ed6b0ec0338c5922ef5

components/o3-form/src/types/utils.ts
AnnaDraganova: Done in the following commit - https://github.com/Financial-Times/origami/pull/1784/commits/74020b40d9f33cb9293a3ed6b0ec0338c5922ef5

components/o3-form/src/tsx/fieldComponents/Feedback.tsx
frshwtr: @AnnaDraganova will this token be added as part of this PR or is it a future piece?

AnnaDraganova: I don't have permission to add tokens. I thought that one of the designers would add it.

frshwtr (approved):
Approved, will check with designers to get the icon added, thanks :)

PR #1785, opened 2024-08-15
feat: create new formfield for checkboxes and radio buttons.

akomiqaia:


components/o3-form/src/tsx/fieldComponents/FormField.tsx
```diff
@@ -12,24 +12,42 @@ export const FormField = ({
...
-					{optional && <span className="o3-form-optional-label">optional</span>}
-				</label>
 			)}
+			{children}
+			{feedback && <Feedback {...feedback} />}
+		</div>
+	);
+};
+
+export const FormFieldCheckboxRadioButton = ({
```
frshwtr, line 36: This name can be more generic to make its difference clearer. It looks like this is used on componnets as part of a group, where Label is nested and appears more than once, Something along the lines of `FormFieldGroup`?

components/o3-form/src/tsx/fieldComponents/FormField.tsx
```diff
@@ -50,7 +68,9 @@ export const FormFieldset = ({
 	const labelId = uniqueId('checkbox_');
 	const descriptionId = uniqueId('checkbox_');
 	return (
-		<fieldset className='o3-form-field' aria-labelledby={`${labelId} ${descriptionId}`}>
+		<fieldset
+			className="o3-form-field"
+			aria-labelledby={`${labelId} ${descriptionId}`}>
```
notlee, line 73: Description belongs to [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby)

>Both aria-labelledby and aria-describedby reference other elements to calculate a text alternative, but a label should be concise, while a description is intended to provide more verbose information; a label describes the essence of an object, while a description provides more information that the user might need.

akomiqaia: @notlee [a62b50b](https://github.com/Financial-Times/origami/pull/1785/commits/a62b50b9782e8ea600514da47ac4e55a94424d0c) I think you are right

frshwtr (approved):

PR #1786, opened 2024-08-16
OR-637/text input component

frshwtr:


components/o3-form/main.css
```diff
@@ -1,4 +1,31 @@
...
+	--_o3-form-text-input-max-width: 3.5em;
+}
+
+.o3-form-text-input--short-5  {
+	--_o3-form-text-input-max-width: 4.5em;
+}
+.o3-form-text-input--error {
+	border: var(--_o3-form-text-input-border-error);
+	background-color: var(--_o3-form-text-input-background-color-error);
+}
```
akomiqaia, line 31: maybe we can move this file to a separate css file? There is a folder in src/css for the component styles.

components/o3-form/src/tsx/TextInput.tsx
```diff
@@ -0,0 +1,26 @@
+import type {FormFieldProps} from '../types';
+import '../../main.css';
+import {FormField} from './fieldComponents/FormField';
+
+export type TextInputProps = {
+	disabled?: boolean,
+	length?: 2 | 3 | 4 | 5
```
akomiqaia, line 7: maybe this is something for our implementor's catchup but the checkbox uses props called `attributes` that include disabled in it as well. I  guess we would want to enable users to pass event handlers as well and maybe a generic `attributes` prop could cover this. Or we can update the checkbox to follow this pattern

components/o3-form/stories/whitelabel/textInput.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+		label: 'Security code',
+		description:
+			'3 digit security code as printed on the back of your credit card.',
+		length:
+			3,
+	},
+	render: (args) => {
+		return (<TextInputTsx label={args.label} disabled={args.disabled} description={args.description} feedback={args.feedback} length={args.length} labelId="label"/>)
+	}
+};
```
akomiqaia, line 57: Do we need different stories for these states or maybe controls can be used to show different states for input element? It can reduce the number of snapshot tests, but we might also miss errors by not having these states as stories. I am quite happy with any approach, but @AnnaDraganova has developed a group checkbox that displays all the checkbox states. Maybe we should take this approach and we should show all form states in one story to reduce the snapshot tests but also cover all errors. but this might give us reduced control over individual input text controls.

frshwtr:
Note from Ako: this needs a readme

frshwtr:
This PR is awaiting some styling changes from #1784

akomiqaia (approved):
LGTM

akomiqaia:
Before merging We need to add figma links to storybook

PR #1802, opened 2024-08-27
fix: align text input max width to 4 columns

frshwtr:

### DIFF
--- a/components/o3-form/src/css/components/text-input.css
+++ b/components/o3-form/src/css/components/text-input.css
@@ -1,10 +1,13 @@
+@import "@financial-times/o3-foundation/grid.css";
 
 .o3-form-text-input {
 	border: var(--_o3-form-text-input-border);
 	padding: var(--o3-spacing-3xs) var(--o3-spacing-2xs);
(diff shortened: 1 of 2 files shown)


components/o3-form/src/css/components/text-input.css
```diff
@@ -1,10 +1,13 @@
+@import "@financial-times/o3-foundation/grid.css";
 
 .o3-form-text-input {
 	border: var(--_o3-form-text-input-border);
 	padding: var(--o3-spacing-3xs) var(--o3-spacing-2xs);
 	background-color: var(--_o3-form-text-input-background-color);
 	border-radius: var(--_o3-form-text-input-border-radius);
-	max-width: var(--_o3-form-text-input-max-width);
+
+	--o3-grid-columns-to-span-count: 4;
```
notlee, line 9: Design question: What about mobile, where their are only 4 columns? Assuming forms are placed in some kind of containing element with its own padding, they will overflow. Should they instead be 100%?

notlee: Ignore that. I think I'm being a novice here because we're setting max-width and not width! 😄

PR #1804, opened 2024-08-28
Styles token update

MaryGodservant:


frshwtr:
Note: this will need to be a breaking change as the token names have changed.

frshwtr:
Looks good, let me know once you've committed the changes to typography based on the decisions we made earlier and i'll give the final review.

components/o3-foundation/src/css/components/Typography/body.css
```diff
@@ -0,0 +1,63 @@
...
+.o3-typography-wrapper > p {
+	font-family: var(--o3-typography-use-case-body-standard-font-family);
+	font-weight: var(--o3-typography-use-case-body-standard-font-weight);
+	font-size: var(--o3-typography-use-case-body-standard-font-size);
+	line-height: var(--o3-typography-use-case-body-standard-line-height);
+	margin: 0 0 var(--o3-spacing-s);
+	color: var(--o3-color-use-case-body-text);
+}
+
+.o3-typography-body--small {
```
notlee, line 11: This isn't a modifier `o3-typography-body o3-typography-body--small`, instead it should be `o3-typography-body-small`

notlee, line 11: I think there's a large/big style too?

notlee, line 11: FYI I noticed the prop on the story only makes some of the body copy small

notlee:
Added a commit to fix some inverse styles (incorrect custom prop names) and ensured the inverse theme applies to new body classes – not confident I caught each case, any way we can increase confidence with visual regression tests in the future?

notlee (approved):

PR #1809, opened 2024-09-03
fix: align padding in drawer

danieleruiz:
## Describe your changes
We need to make some design adjustments that we saw after release new version of search drawer
Changes:

- Padding between the search bar and the “Top sections” box should be 16px. If there’s an “ask FT” button, then it should be 16px between the bar and the button and 16px between the button and the “Top sections”
- Text inside the bar seems very faint so we use black tint 60 (happens in safari)
- The underline on the option that’s not selected should be black tints 80, not teal (happens in safari)

https://financialtimes.atlassian.net/browse/CON-3544
## Link to Figma designs
https://docs.google.com/document/d/1qsmZwl4AREo__B46fivT5MHLc90lmClil_XBPH0DYw4/edit

- [ x] I have applied `percy` label for o-[COMPONENT] or `chromatic` label for o3-[COMPONENT] on my PR before merging and after review. Find more details in [CONTRIBUTING.md](https://github.com/Financial-Times/origami/blob/main/CONTRIBUTING.md#pull-requests-and-visual-regression-tests)

### DIFF
--- a/components/o-header/src/tsx/drawer.tsx
+++ b/components/o-header/src/tsx/drawer.tsx
@@ -11,7 +11,6 @@ export function Drawer({
 	const subscribeAction = data.subscribeAction;
 	const navItems = data.drawer.items;
 	const userData = userIsLoggedIn ? data.user : data.anon;
-	const showAskFtButton = showAskButton;
 	const showSubscribeButton = !userIsSubscribed && subscribeAction;
 	const showDrawerActions =
 		showAskFtButton || showSubscribeButton;
--- a/components/o-header/src/tsx/drawer.tsx
+++ b/components/o-header/src/tsx/drawer.tsx
@@ -21,15 +21,8 @@ export function Drawer({data, showAskButton, userIsLoggedIn, userIsSubscribed }:
 					otherEditions={editions.others}
 				/>
 				<DrawerSearch />
-				{showAskButton && 
-					<AskFtButton
-						variant="drawer"
-						dataTrackable="ask-ft-button-drawer"
-						id="ask-ft-button-drawer"
-					/>
-				}
 				{!userIsSubscribed && subscribeAction && (
-					<DrawerAction action={subscribeAction} />
+					<DrawerAction action={subscribeAction} showAskButton={showAskButton} />
 				)}
 				<DrawerMenu navItems={navItems} />
 				<DrawerUser {...userData} />
@@ -61,9 +54,16 @@ function DrawerTools({
 	);
 }
 
-function DrawerAction({action}: {action: TNavAction}) {
+function DrawerAction({action, showAskButton}: {action: TNavAction, showAskButton: boolean}) {
 	return (
 		<div className="o-header__drawer-actions">
+			{showAskButton &&
+				<AskFtButton
+					variant="drawer"
+					dataTrackable="ask-ft-button-drawer"
+					id="ask-ft-button-drawer"
+				/>
+			}
 			<a className="o-header__drawer-button" role="button" href={action.url}>
 				{action.name}
 			</a>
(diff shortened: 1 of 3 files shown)


frshwtr:
Space between the "ask FT" button and subscribe seems to be very small now. Is that an intended change?

image

danieleruiz:
> Space between the "ask FT" button and subscribe seems to be very small now. Is that an intended change?
> 
> image

That already happened before, those buttons never appear together

frshwtr:
> > Space between the "ask FT" button and subscribe seems to be very small now. Is that an intended change?
> > image
> 
> That already happened before, those buttons never appear together

Something did change, here's the live version for comparison.

image


Given they won't be displayed together it shouldn't be an issue.

frshwtr:
The spacing looks rather large between the search button and the "ask FT" button. I see padding from the search box of `16px` and then margin of `12px` on the top of the "ask FT" button. Resulting in a gap of `28px` Is this the correct spacing?

image

danieleruiz:
> The spacing looks rather large between the search button and the "ask FT" button. I see padding from the search box of `16px` and then margin of `12px` on the top of the "ask FT" button. Resulting in a gap of `28px` Is this the correct spacing?
> 
> image
> The space between top sections and ask FT is also `12px` instead of `16px`

the placement of the button in the storybook demo is incorrect it should be inside "drawer-actions div" in production you can see that is there.

components/o-header/src/tsx/drawer.tsx
```diff
@@ -21,15 +21,8 @@ export function Drawer({data, showAskButton, userIsLoggedIn, userIsSubscribed }:
...
-				{showAskButton && 
-					<AskFtButton
-						variant="drawer"
-						dataTrackable="ask-ft-button-drawer"
-						id="ask-ft-button-drawer"
-					/>
-				}
 				{!userIsSubscribed && subscribeAction && (
-					<DrawerAction action={subscribeAction} />
+					<DrawerAction action={subscribeAction} showAskButton={showAskButton} />
```
frshwtr, line 25: L24 above means that the ask FT button will only show when a user is not subscribed, but this ask button should display when they are already subscribed. 

Can this code be modified so that the subscribe button is shown if they are not already, otherwise the code shows the ask button?

danieleruiz: i think the logic is completely so in all cases, because i think they are cases when you are subscribed but the ask button should not be shown, anyway, the concrete logic is not important here because it is only the storybook template.

frshwtr: > i think the logic is completely so in all cases, because i think they are cases when you are subscribed but the ask button should not be shown

Can you confirm if these criteria describe your behaviour correctly:

```
Given I am a subscribed FT User
When I view the header menu
Then I should see the "Ask FT"
And I should not see the "Subscribe for full access" button
```

and inversely:

```
Given I am not a subscribed FT user
Or I am not logged in
When I view the header menu
Then I should see the "Subscribe for full access" button
And I should not see the "Ask FT" button
```
It will help me understand what behaviours you intend to have for the drawer.

> the concrete logic is not important here because it is only the storybook template.

It is still a behaviour of the Drawer component which we release to anyone consuming o-header, meaning any user that uses this component will have this logic. It's important that we show the expected behaviours for all users in Storybook. So this demo should work correctly.

danieleruiz: the first case should be like this:
    ```
Given I am a subscribed FT User
    When I view the header menu
    Then I should see the "Ask FT" button if the "showAskButton" property is present
    And I should not see the "Subscribe for full access" butto
```n

frshwtr (approved):

PR #1830, opened 2024-09-30
fix: remove unnecessary token building for o3-social-sign-in

frshwtr:
## Describe your changes
o3-social-sign in has a tokens build step in it's `build` script, there are no tokens in this package so it can be removed.

This will help clean up local dev by reducing the generation of unnecessary files.

### DIFF
--- a/components/o3-social-sign-in/package.json
+++ b/components/o3-social-sign-in/package.json
@@ -23,7 +23,7 @@
     "social-sign-in"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


notlee (approved):

components/o3-social-sign-in/package.json
```diff
@@ -23,7 +23,7 @@
     "social-sign-in"
   ],
   "scripts": {
```
notlee, line 25: I see `main` is wrong and `module` is missing. Could we through that into this PR?

```
	"main": "./cjs/index.js",
	"module": "./esm/index.js",
```

notlee: actually, I got it in another PR

PR #1834, opened 2024-10-02
feat: add error summary component to o3-forms

akomiqaia:
## Describe your changes
Create an error summary component for o3-form.

[OR-90](https://financialtimes.atlassian.net/browse/OR-910)

## Link to Figma designs
[Error Summary](https://www.figma.com/design/5ATknbGociZMlnNXX4sy4f/branch/R6fcpcKjdAtJZ7qxuNgN0G/Whitelabel---Origami-(o3)?node-id=8236-447&node-type=instance&m=dev)


components/o3-form/src/css/components/error-summary.css
```diff
@@ -0,0 +1,34 @@
...
+			margin-bottom: var(--o3-spacing-4xs);
+		}
+	}
+
+	.o3-forms__error-summary__list {
+		list-style-type: none;
+		padding-left: 0;
+		margin-top: 0;
+		a {
+			line-height: var(--o3-font-lineheight-metric2-2);
```
notlee, line 23: Do we actually want to use `line-height` here or is Deyan's intention to increase `margin`? I.e:

1. If the error copy goes over multiple lines, do we want those lines to be further apart? (line-height) Or;
2. Do we  want space between different error messages? (margin)

If (1), I'd question why we're diverting from our own type guidelines here, and request that is documented in the design guidelines as a design decision.

If (2), that makes sense!

components/o3-form/src/css/components/error-summary.css
```diff
@@ -0,0 +1,34 @@
...
+			color: var(--o3-color-use-case-body-text);
+			text-decoration-color: var(--o3-color-use-case-body-text);
+			text-decoration-thickness: 2px;
+			text-underline-offset: var(--o3-spacing-5xs);
```
notlee, line 27: Does this align with our standard link styles? Can we align these by either:

1. Use the foundations link style, and add a feature with docs to customise link colours with custom properties; or
2. Add link usecase tokens for underline

Thoughts? Probably (1) would be more effective here as it allows us to consistently apply/change all link css properties

components/o3-form/src/css/components/error-summary.css
```diff
@@ -0,0 +1,34 @@
+.o3-form__error-summary {
```
notlee, line 1: Does this have a `.o3-form` parent element? If not:

```suggestion
.o3-form-error-summary {
```

components/o3-form/src/css/components/error-summary.css
akomiqaia: it might be 2.

components/o3-form/src/css/components/error-summary.css
akomiqaia: maybe? or I would think it might have `<form>` element that has a class of `o3-form`? or do you think this error summary won't end up inside the form element? o2 form element has classes [this way](https://o2.origami.ft.com/?path=/docs/o2-core_components-o-forms-readme--docs#error-summary)

components/o3-form/src/css/components/error-summary.css
akomiqaia: This might be caused because line-height is set to 28 in Figma. I will ask about it.

components/o3-form/src/css/components/error-summary.css
notlee: Enforcing that it's a child of `o3-form` would make its placement less flexible – which may or may not be what we want from a design perspective – but I think you're probably right that it is. Perhaps run in by Deyan to decide. Do we want to enforce the error summary is displayed within the form or might it be displayed at the top of the page, with some other elements, followed be the form?

components/o3-form/src/css/components/error-summary.css
notlee: Could you confirm and use margins instead if that is the case?

components/o3-form/src/css/components/error-summary.css
akomiqaia: we agreed that currently this will be part of the Form element

frshwtr: Have we aligned on the convention? We use a mix of declaring a field as a BEM element, and declaring it as a BEM block.

components/o3-form/src/tsx/Error-summary.tsx
```diff
@@ -0,0 +1,30 @@
...
+				className="o3-form__error-summary__heading"
+				role="alert"
+				aria-labelledby="error-summary-title">
+				<span id="error-summary-title" className="o3-typography-heading5">
+					{errorMessage}
+				</span>
+				<ul className="o3-forms__error-summary__list o3-typography-body-small">
+					{errors.map(({id, fieldName, message}) => (
+						<li key={id}>
+							<a className="o3-typography-link" href={`#${id}`}>
```
frshwtr, line 20: The text in the `a` and `span` have different sizes, in the [design](https://www.figma.com/design/R6fcpcKjdAtJZ7qxuNgN0G/%D0%BE3-input-summary?node-id=7167-918&m=dev) they have equal size.

image

frshwtr:
Other components on `staging-o3-form` have stories for other brands now displaying whitelabel demos as placeholders. Can you add the same here?

PR #1843, opened 2024-10-08
feat: introduce password input

frshwtr:
## Checklist before requesting a review

- [ ] I have applied `percy` label for o-[COMPONENT] or `chromatic` label for o3-[COMPONENT] on my PR before merging and after review. Find more details in [CONTRIBUTING.md](https://github.com/Financial-Times/origami/blob/main/CONTRIBUTING.md#pull-requests-and-visual-regression-tests)
- [ ] If it is a new feature, I have added thorough tests.
- [ ] I have updated relevant docs.
- [ ] I have updated relevant env variables in Doppler.

PR #1845, opened 2024-10-09
fix: use use case tokens for checkboxes

frshwtr:


frshwtr:
Some form tokens have been included as part of #1843, this PR contains changes for the rest.

akomiqaia (commented):
this looks good to me but shall we update `.o3-form-input__checkbox-label` font styles with use case tokens? they are body-small and maybe using those makes sense? Thinking about it maybe we should have a generic class to apply label styles on all form elements

components/o3-form/src/css/components/checkbox.css
```diff
@@ -107,9 +111,9 @@
 .o3-form-input__checkbox-label {
```
akomiqaia, line 111: I know I defined these classes but should we rename them to `o3-form-input-checkbox__label` ?

akomiqaia (approved):
I have one comment that I am happy you to resolve it as you think is more correct. Otherwise this is good to be merged

PR #1860, opened 2024-10-31
Refactor token build

notlee:
- Use one dictionary and target multiple platforms
- Remove noise from output: no warnings due to filtered tokens, error if token reference is missing
- All tokens are now rebuild by `npm run build -w apps/dictionary/`, and the output tokens committed. Tokens are not built as part of the component build. This ensures all token changes are expected, and components and tooling are updated to the latest tokens at once. Previously we've forgotten to build tokens which has happened on release, leading to broken components.
- Icons are currently built multiple times for each brand. Opportunity for optimisation missed but I think we want this, given it's not impossible some brands will introduce unique tokens.

```
.
├── README.md 
├── src 
│   ├── build.js
│   ├── config 
│      ├── platforms 👈 configuration for CSS target / JS tooling target
│   ├── formatters
│   ├── transforms
│   └── utils.js
└── tokens 👈 raw tokens from Token Studio
```

I deleted the tests here because they were tightly coupled to implementation. As build output is committed I propose relying on that for review, to ensure tokens are changed only as we expect.

For a future PR I'd like to simplify the output path:
```diff
-components/o3-button/src/css/tokens/core/professional/o3-button/_variables.css
+components/o3-button/src/css/tokens/_professional.css
```


libraries/o3-tooling-token/build/core/_variables.js
```diff
@@ -1593,7 +1593,7 @@ export default {
 		"shortName": "5xs",
 		"value": "4px",
 		"originalValue": "4px",
-		"type": "spacing",
+		"type": "dimension",
```
notlee, line 1596: I'm not quite sure why the type here has gone from `spacing` to `dimension` but I remember Ako mentioning something about this before... perhaps because tokens-studio wasn't set as a preprocessor? Either way there's more shared config now so I updated the website to work with this

frshwtr: We changed some types to dimension before due to figma and style dictionary interpreting `spacing` differently.

frshwtr (approved):

PR #1862, opened 2024-11-04
fix: Correctly detect truly circular paths in JS objects, rather than objects with multiple references

georgecrawford:
[This conversation](https://financialtimes.slack.com/archives/C07U72QTQCA/p1730548886171719?thread_ts=1730484139.054259&cid=C07U72QTQCA) uncovered an issue whereby multiple offline calls to `track()` could potentially share object references (for example, in the additional `context` object).

The previous `containsCircularPaths()` implementation wasn't really looking for **circular** paths, but rather multiple references to the same object. See the [original PR](https://github.com/Financial-Times/o-tracking/pull/263) and [issue](https://github.com/Financial-Times/o-tracking/issues/115).

A use-case which is common in an offline-first application such as FT App - queueing multiple events to be stored and sent when next online - is somewhat likely to include a shared reference across the events at some point in the event tracking. This causes an `o-tracking does not support circular references in the analytics data` error, which discards the whole queue of events.

## Describe your changes
This implementation is based on https://github.com/sindresorhus/decircular and https://github.com/sindresorhus/decircular, and draws on the utility of `JSON.serialize()`'s [**replacer** parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter). 

Circular references are removed before sending over the network, or storing in LocalStorage. They are replaced with a string reference (e.g. ``Circular reference between `$.context.customTrackingData` AND `$.context.customTrackingData.circular` ``), and an error is thrown in the next event loop to warn that the data has been modified.

### DIFF
--- a/libraries/o-tracking/src/javascript/utils.js
+++ b/libraries/o-tracking/src/javascript/utils.js
@@ -241,100 +241,131 @@ function assignIfUndefined (subject, target) {
 	}
 }
 
-/**
- * Used to find out all the paths which contain a circular reference.
- *
- * @param {*} rootObject The object we want to search within for circular references
- * @returns {string[]} Returns an array of strings, the strings are the full paths to the circular references within the rootObject
- */
-function findCircularPathsIn(rootObject) {
-	const traversedValues = new WeakSet();
-	const circularPaths = [];
-
-	function _findCircularPathsIn(currentObject, path) {
-		// If we already saw this object
-		// the rootObject contains a circular reference
-		// and we can stop looking any further into this currentObj
-		if (traversedValues.has(currentObject)) {
-			circularPaths.push(path);
-			return;
-		}
-
-		// Only Objects and things which inherit from Objects can contain circular references
-		// I.E. string/number/boolean/template literals can not contain circular references
-		if (currentObject instanceof Object) {
-			traversedValues.add(currentObject);
-
-			// Loop through all the values of the current object and search those for circular references
-			for (const [key, value] of Object.entries(currentObject)) {
-				// No need to recurse on every value because only things which inherit
-				// from Objects can contain circular references
-				if (value instanceof Object) {
-					const parentObjectIsAnArray = Array.isArray(currentObject);
-					if (parentObjectIsAnArray) {
-					// Store path in bracket notation when value is an array
-						_findCircularPathsIn(value, `${path}[${key}]`);
-					} else {
-					// Store path in dot-notation when value is an object
-						_findCircularPathsIn(value, `${path}.${key}`);
-					}
-				}
-			}
-		}
-	}
-
-	_findCircularPathsIn(rootObject, "");
-	return circularPaths;
-}
 
 /**
- * Used to find out whether an object contains a circular reference.
+ * Identify circular references in 'object', and replace them with a string representation
+ * of the reference. Returns a succesfully serialised JSON string, and a list of circular
+ * references which were removed.
+ * 
+ * Inspired by https://github.com/sindresorhus/safe-stringify and 
+ * https://github.com/sindresorhus/decircular
  *
- * @param {object} rootObject The object we want to search within for circular references
- * @returns {boolean} Returns true if a circular reference was found, otherwise returns false
+ * @param {*} object The object we want to stringify, and search within for circular references
+ * @returns {Object: {jsonString: string, warnings: array}} The stringified object, and a warnings for each circular reference which was removed
  */
-function containsCircularPaths(rootObject) {
-	// Used to keep track of all the values the rootObject contains
-	const traversedValues = new WeakSet();
+function removeCircularReferences(object) {
 
-	/**
-	 *
-	 * @param {*} currentObject The current object we want to search within for circular references
-	 * @returns {boolean|undefined} Returns true if a circular reference was found, otherwise returns undefined
-	 */
-	function _containsCircularPaths(currentObject) {
-		// If we already saw this object
-		// the rootObject contains a circular reference
-		// and we can stop looking any further
-		if (traversedValues.has(currentObject)) {
-			return true;
+	// WeakMaps release memory when all references are garbage-collected
+	const circularReferences = new WeakMap();
+	const paths = new WeakMap();
+
+	const warnings = [];
+
+	function getPathFragment(parent, key) {
+		if (!key) {
+			return '$';
 		}
 
-		// Only Objects and things which inherit from Objects can contain circular references
-		// I.E. string/number/boolean/template literals can not contain circular references
-		if (currentObject instanceof Object) {
-			traversedValues.add(currentObject);
-			// Loop through all the values of the current object and search those for circular references
-			for (const value of Object.values(currentObject)) {
-				// No need to recurse on every value because only things which inherit
-				// from Objects can contain circular references
-				if (value instanceof Object) {
-					if (_containsCircularPaths(value)) {
-						return true;
-					}
-				}
-			}
+		if (Array.isArray(parent)) {
+			return `[${key}]`;
 		}
+
+		return `.${key}`;
 	}
 
-	// _containsCircularPaths returns true or undefined.
-	// By using Boolean we convert the undefined into false.
-	return Boolean(
-		_containsCircularPaths(
-			rootObject
-		)
-	);
+	function formatCircularReferencesWarning(references) {
+		const paths = references.map(path => '`' + path.join('') + '`');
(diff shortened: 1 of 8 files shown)


libraries/o-tracking/test/utils.test.js
```diff
@@ -122,28 +121,28 @@ describe('Utils', function () {
...
 
-		it('should return an empty array if given a boolean literal', function() {
-			proclaim.deepStrictEqual(findCircularPathsIn(true), []);
+		it('should return null if given a boolean literal', function() {
+			proclaim.strictEqual(safelyStringifyJson(true), 'true');
 		});
 
-		it('should return an empty array if given null', function() {
-			proclaim.deepStrictEqual(findCircularPathsIn(null), []);
+		it('should return null if given null', function() {
```
notlee, line 141:
```suggestion
		it('should return a string if given null', function() {
```

notlee (changes requested):
This looks like a fun one, thanks for the PR! Just some test naming to flag and a little clarification to ensure my understanding. o-tracking is owned by the Data Platform team now but I'm happy to approve and release this on their behalf if they are

libraries/o-tracking/src/javascript/utils.js
```diff
@@ -224,97 +224,79 @@ function assignIfUndefined (subject, target) {
...
+			&& !(value instanceof Date)
+			&& !(value instanceof Number)
+			&& !(value instanceof RegExp)
+			&& !(value instanceof String)
+		) {
+			oldPath = traversedValues.get(value);
+
+			if (oldPath !== undefined) {
+				circularPaths.push(path);
+				return { $ref: oldPath };
```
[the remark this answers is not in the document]
georgecrawford: Thanks - I've added a comment to explain this, as it's completely confusing otherwise!

georgecrawford:
@notlee Thanks for your review. I've addressed your issues, but I don't think I've helped matters by force-pushing. GitHub is confused, or I've missed a step. Could you check over my changes please? Thank you!

libraries/o-tracking/src/javascript/utils.js
```diff
@@ -222,99 +222,94 @@ function assignIfUndefined (subject, target) {
 }
 
 /**
- * Used to find out all the paths which contain a circular reference.
+ * Used to find out all the paths which contain a circular reference.
```
rowanbeentje, line 225: **Nitpick:** I think this particular line now only applies to the contained function ;)

libraries/o-tracking/src/javascript/utils.js
```diff
@@ -222,99 +222,94 @@ function assignIfUndefined (subject, target) {
...
  *
- * @param {*} rootObject The object we want to search within for circular references
- * @returns {string[]} Returns an array of strings, the strings are the full paths to the circular references within the rootObject
+ * @param {*} object The object we want to stringify, and search within for circular references
+ * @returns {string} The safely stringified JSON string
+ * @throws {Error} Error with message highlighting the paths which are circular references
  */
-function findCircularPathsIn(rootObject) {
-	const traversedValues = new WeakSet();
+function safelyStringifyJson(object) {
```
rowanbeentje, line 232: **Question:** This function doesn't quite do this - should we change the name?

Right now it does three things:
1) Try to JSON.stringify the object. If this succeeds, the object contains no circular references, and is returned.
2) If the object failed to stringify, a safe-encoding routine is run, recording circular references.
3) If circular references were found, an error is thrown.

So really the only reason for (2) is so that (3) can report the circular reference paths. And it also means this function is not safely stringifying the json - instead it's designed to error.

I think I'd be in favour of actually safely stringifying the data! I don't think we need to go as far as using the JSON schema's support, as it seems unlikely that we'd ever have a server try to reconstruct the original data. Instead I think we could replace it with a human-readable path reference, so the tracking data isn't lost; potentially also paired with a console.error - or perhaps instead an actual thrown error in a timeout, so that global error handlers can also catch/report?

libraries/o-tracking/src/javascript/utils.js
```diff
@@ -222,99 +222,94 @@ function assignIfUndefined (subject, target) {
...
+			&& value !== null
+			&& !(value instanceof Boolean)
+			&& !(value instanceof Date)
+			&& !(value instanceof Number)
+			&& !(value instanceof RegExp)
+			&& !(value instanceof String)
+		) {
+			oldPath = traversedValues.get(value);
+
+			if (oldPath !== undefined) {
```
rowanbeentje, line 260: **Thought:** much like the old function, I think this isn't actually detecting only circular references, is it? It's detecting object values seen before, by putting every object as a key in a weakmap and then using references if those objects are encountered again.

Your use of json.stringify() in a try/catch is hiding this from the tests I think 😁 

(That said, I think the new function is a significant improvement over the old one because it has much better skipping of built-in types including null. But I think if we're relying on a list of circular references in the error, it won't be...)

rowanbeentje (commented):
I think this would work, but mostly because of the try/catch 😁 I've got a couple of nitpicks about the approach, and as we don't have proper circular detection *anyway* I'm wondering how good the resulting errors would be.

I'm still in favour of sending the events with some values replaced with error strings - I think that's nicer behaviour for a tracking library than dropping events?

libraries/o-tracking/src/javascript/utils.js
georgecrawford: Thanks for this considered feedback. 

I hope I've improved the API now. I'm also throwing in a timeout - I like that suggestion.

And, fundamentally, this is now genuinely checking for circular references rather than pretending to.

libraries/o-tracking/src/javascript/utils.js
georgecrawford: You're correct. I hope I've improved the logic for circular reference detection now.

georgecrawford:
@rowanbeentje I'm happy with this PR now. Would you be able to take another look before we ask Data Platform for a review? Thank you!

notlee (approved):
👍 from me, I'll wait for a review from data platform before merging

danail-s (approved):

danail-s:
Hi @georgecrawford , if you could just bump the version in `/src/javascript/tracking.js` to `4.6.1`, everything else should be good and we can release on Monday just because we avoid pushing code on Friday

libraries/o-tracking/src/javascript/utils.js
```diff
@@ -241,101 +241,113 @@ function assignIfUndefined (subject, target) {
...
 
+
+/**
+ * Stringify an object to JSON, removing any circular references. When circular references
+ * are found, an error is thrown in a new event loop so that global error handlers can report it.
+ *
+ * @param {*} object The object we want to stringify, and search within for circular references
+ * @returns {string} The safely stringified JSON string
+ */
+function safelyStringifyJson(object) {
```
rowanbeentje, line 334: **Praise**: ooh this, and the removeCircularReferences, are very neat. Took me a little while to work out *how* it was doing it, lots of recursive walking of data structures to detect genuine circles 👍🏻 Really cleanly commented and well tested too! 🎉 

**Suggestion**: The one issue I think might exist with this is because of the amount of recursive tracking it has to do to identify those genuine circular references. I've tested it with a standard Apps spoor event which doesn't have much data in it, and there this code is almost 20x slower than a basic `JSON.stringify` 😁 In your previous approach you wrapped the new code in a try { JSON.stringify } catch () { do safe encode }, and applying that same approach here for me makes it only a few percent slower than the base stringify (assuming no circular references!!).

Obviously on a fast machine even this code is ridiculously fast - 31K encodes per second on a M2 Max - *but* the reason I think this may be worth adding is that when offline, store.js will rapidly collect tens of events so this will rapidly take meaningful time on a slower Android CPU...

georgecrawford: Hah. OK, while 'back to v1' in a sense, this is a very good reason to do so - performance, rather than detection. I'm going to add this with a comment, and _not_ change the tests as there's no testable difference. Thanks @rowanbeentje.

rowanbeentje (approved):
Yay, thanks, looks good to me! 🎉 

I have *not* verified this with a client integration but looks good within the codebase 😁 
**Edit:** gave it a quick test in my local app, working online and offline with a few standard events 👍🏻

georgecrawford:
@danail-s Thanks for this. The code is rebased, approved and I've bumped the version number. I'm happy for you to release on Monday.

PR #1868, opened 2024-11-13
Replace o-buttons with o-private-foundation in o-table [OR-1051]

notlee:
Takes o3-button and refactors to work in an o2 world. Slightly
odd use of Sass/CSS Custom Properties, with Sass conditionals
used to output duplicate CSS Custom Properties which override
each other in the same declaration. This was the safest way
to reuse o3 styles by effectively replacing class names for
Sass if statements. It's also more readable and simple than o2
Sass. Despite duplicating some properties, CSS output and
performance will be improved by moving away from the Origami
Image Service for coloured icons.

Currently supports:
- Types
- Themes
- Sizes
- Icons
- Icons only

Does not yet support:
- Custom theme maps

Will not support:
- Outputting CSS granularly. There are not many buttons left in
o2 components, and bundle size is reduced overall as noted above.

Other things of note:
- `o-private-foundation-get` allows other o2 components to grab tokens directly.
- Added Metric2 variable font

### DIFF
--- a/components/o-private-foundation/main.scss
+++ b/components/o-private-foundation/main.scss
@@ -1,4 +1,18 @@
 @use './src/scss/brand' with (
 	$o-brand: $o-brand
 );
+
+@forward './src/scss/brand' as o-private-foundation-*;
+
 @import './src/scss/variables';
+
+@import './src/scss/o-buttons/main';
+
+@font-face {
+	src: url('https://www.ft.com/__origami/service/build/v3/font?version=1.13&font_name=Metric2-VF&system_code=origami&font_format=woff2')
+		format('woff2-variations');
+	font-family: 'metric 2 VF';
+	font-weight: 300 800;
+	font-style: normal;
+	font-display: swap;
+}
--- a/components/o-private-foundation/src/scss/o-buttons/main.scss
+++ b/components/o-private-foundation/src/scss/o-buttons/main.scss
@@ -21,7 +21,7 @@
 	);
 	--_o-pf-button-font-weight: #{brand.get('o3-font-weight-semibold')};
 
-	/* SMALL BUTTON */
+	/* Button Sizes */
 	@if ($size == 'big') {
 		--_o-pf-button-min-width: 80px;
 		--_o-pf-button-min-height: 40px;
--- a/components/o-private-foundation/src/scss/o-buttons/main.scss
+++ b/components/o-private-foundation/src/scss/o-buttons/main.scss
@@ -57,7 +57,7 @@
 	font-size: var(--_o-pf-button-font-size);
 	line-height: var(--_o-pf-button-lineheight);
 	font-weight: var(--_o-pf-button-font-weight);
-	font-family: #{brand.get('o3-font-family-metric')} + ', sans-serif';
+	font-family: '#{brand.get('o3-font-family-metric')}', 'sans-serif';
 	border: var(--_o-pf-button-border-size) solid transparent;
 	text-align: center;
 	text-decoration: none;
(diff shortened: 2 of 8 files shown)


components/o-private-foundation/main.scss
```diff
@@ -1,4 +1,18 @@
 @use './src/scss/brand' with (
 	$o-brand: $o-brand
 );
+
+@forward './src/scss/brand' as o-private-foundation-*;
```
notlee, line 5: Allow dependencies to directly access tokens through `o-private-foundation-get('token');`

components/o-private-foundation/main.scss
```diff
@@ -1,4 +1,18 @@
...
+
 @import './src/scss/variables';
+
+@import './src/scss/o-buttons/main';
+
+@font-face {
```
notlee, line 11: Getting starting on o-typography, by adding Metric2 for button type! 🎉

components/o-table/demos/src/demo.scss
```diff
@@ -1,16 +1,15 @@
...
 @include oNormalise();
-@import "@financial-times/o-typography/main";
+@import '@financial-times/o-typography/main';
 @include oTypography();
-@import "@financial-times/o-forms/main";
+@import '@financial-times/o-forms/main';
 @include oForms();
-@import "@financial-times/o-buttons/main";
-@include oButtons();
+
```
notlee, line 12: Removed from demos because it's no longer needed for demo purposes (it's an odd quirk that it was included to begin with)

components/o-table/src/js/Tables/OverflowTable.js
```diff
@@ -203,56 +233,91 @@ class OverflowTable extends BaseTable {
...
+						: ''
+				}
 				<div class="o-table-overflow-control-overlay">
-					${this.wrapper && supportsArrows ? `
+					${
+						this.wrapper && supportsArrows
+							? `
 						<div class="o-table-control o-table-control--back o-table-control--hide">
-							<button aria-label="visually scroll table back" disabled="true" class="o-buttons o-buttons--primary o-buttons--big o-buttons-icon o-buttons-icon--icon-only o-buttons-icon--arrow-left"></button>
+							<button aria-label="visually scroll table back" disabled="true"></button>
```
notlee, line 248: Removed `o-buttons` classes from the JS. o-table overrode these directly with button mixins but also added the classes, and didn't output o-buttons CSS anyway outside demos!? 

This isn't a breaking change since o-table owns this markup, it's added client side dynamically

components/o-table/src/js/Tables/OverflowTable.js
```diff
@@ -118,26 +135,31 @@ class OverflowTable extends BaseTable {
...
-		this._expanderUpdateScheduled = window.requestAnimationFrame(function () {
-			this.rootEl.setAttribute('data-o-table-expanded', Boolean(expand));
-			this.container.classList.toggle('o-table-container--expanded', expand);
-			this.container.classList.toggle('o-table-container--contracted', contract);
-			expanderButton.style.display = canExpand ? '' : 'none';
-
-			if (!canExpand) {
-				this.rootEl.removeAttribute('aria-expanded');
-			}
+		this._expanderUpdateScheduled = window.requestAnimationFrame(
```
notlee, line 138: Most of this is Prettier...


...I only removed o-buttons markup

components/o-private-foundation/src/scss/o-buttons/main.scss
```diff
@@ -0,0 +1,158 @@
+@mixin oPrivateButtonsContent(
```
frshwtr, line 1: file structure differs between `o-normalise` and `o-buttons`

`o-normalise`: `o-private-foundation/o-normalise/main.scss`
`o-buttons`: `o-private-foundation/src/scss/o-buttons/main.scss`

Which approach should we align to?

notlee: The latter, as it follows the same structure as other o2 components: one `main.scss` in the root with the rest of sass under `src/scss`

frshwtr:
Thanks Lee, this looks good. Are you intending to swap out `o-buttons` for `o-private-foundation` in components in another PR?

components/o-private-foundation/src/scss/o-buttons/main.scss
frshwtr: 👍 o-normalise will need updating in this case to follow the same convention

notlee: Let's do that in a separate PR

frshwtr (approved):

PR #1881, opened 2024-11-21
Remove o-buttons from o-tabs, o-header, o-forms, o-comments

notlee:
I needed to make some changes to other packages to support this. See commit messages.

### DIFF
--- a/components/o-header/main.scss
+++ b/components/o-header/main.scss
@@ -4,7 +4,7 @@
 @import '@financial-times/o-colors/main';
 @import '@financial-times/o-icons/main';
 @import '@financial-times/o-grid/main';
-@import "@financial-times/o-private-foundation/o-normalise/main";
+@import '@financial-times/o-private-foundation/main';
 @import '@financial-times/o-typography/main';
 @import '@financial-times/o-visual-effects/main';
 @import '@financial-times/o-forms/main';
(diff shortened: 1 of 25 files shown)


components/o-private-foundation/main.scss
```diff
@@ -1,11 +1,8 @@
-@use './src/scss/brand' with (
-	$o-brand: $o-brand
-);
-
-@forward './src/scss/brand' as o-private-foundation-*;
+@import './src/scss/brand';
 
 @import './src/scss/variables';
 
+@import './o-normalise/main';
```
notlee, line 5: Other o2 components import all files in `main` and dependencies then only import one main file, we should follow  for o-private-foundation

components/o-header/main.scss
```diff
@@ -4,7 +4,7 @@
 @import '@financial-times/o-colors/main';
 @import '@financial-times/o-icons/main';
 @import '@financial-times/o-grid/main';
-@import "@financial-times/o-private-foundation/o-normalise/main";
```
frshwtr, line 7: o-buttons import on L2 can be removed.

notlee: Good catch! That's done

frshwtr (approved):

PR #1890, opened 2024-12-03
2025 release typography

notlee:
Remove o-typography from a selection of components covering:

- oTypographySans
- oTypographyLink
- oTypographyList

Including refactored support for custom themes.

### DIFF
--- a/components/o-banner/src/scss/themes/_theme.scss
+++ b/components/o-banner/src/scss/themes/_theme.scss
@@ -62,9 +62,11 @@
 		color: $text-color;
 		border-bottom-color: $brand-border;
 	}
+
 	.o-banner__content {
 		a {
-			@include oTypographyLink(
+			@include oPrivateTypographyLink();
+			@include oPrivateTypographyLinkThemeOverride(
 				$theme: (
 					'base': $link-text-color,
 					'hover': $link-text-color,
@@ -104,7 +106,12 @@
 				'context': $background-color,
 			)
 		);
-		@include oTypographyLink($theme: $link-theme);
+		@if (type-of($link-theme) != 'map') {
+			@include oPrivateTypographyLink($link-theme);
+		} @else {
+			@include oPrivateTypographyLink();
+			@include oPrivateTypographyLinkThemeOverride($link-theme);
+		}
 	}
 	.o-banner__close {
 		@include oIconsContent(
@@ -115,7 +122,7 @@
 
 		@include oPrivateNormaliseFocusApply() {
 			@include oPrivateNormaliseFocusContentForElementColour($text-color);
-		};
+		}
 	}
 
 	// When combined with the "small" and "compact" themes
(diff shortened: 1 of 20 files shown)


components/o-banner/src/scss/themes/_theme.scss
```diff
@@ -62,9 +62,11 @@
 		color: $text-color;
 		border-bottom-color: $brand-border;
 	}
+
 	.o-banner__content {
 		a {
-			@include oTypographyLink(
+			@include oPrivateTypographyLink();
+			@include oPrivateTypographyLinkThemeOverride(
```
notlee, line 69: Optimising for clarity, simplicity, and separating custom themes (which we aren't carrying over into o3). At the slight expense of CSS bundle size optimisation. Bundle size will be reduced overall in the 2025 release due to button optimisations.

components/o-private-foundation/src/scss/o-typography/main.scss
```diff
@@ -0,0 +1,266 @@
...
+@mixin oPrivateTypographySans(
+	$scale: null,
+	$line-height: null,
+	$weight: null,
+	$style: null
+) {
+	// Adjust scale down, to support migrating from
+	// Metric2 to Metric. Metric2 renders larger for the same font-size.
+	@if $scale {
+		$scale: $scale - 1;
```
notlee, line 14: When we apply the mobile type optimisation we may want to remove this by updating the `oPrivateTypographySans` calls. We'll want to ensure at this point `-2`, which becomes `-3`, is never called as we will delete that as an o3 token.

components/o-private-foundation/main.scss
```diff
@@ -17,4 +18,6 @@
 		font-style: normal;
 		font-display: swap;
 	}
+
+	// @todo include italic font
```
frshwtr, line 22: Is this to be done on this PR?

components/o-video/src/scss/_info.scss
```diff
@@ -31,14 +31,15 @@
 	}
 
 	.o-video__info-title {
-		@include oTypographySans(4, $line-height: 30px, $weight: semibold);
+		@include oPrivateTypographySans(4, $weight: semibold);
```
frshwtr, line 34: Does this `semibold` need to be in quotes like L8?

frshwtr (approved):

components/o-private-foundation/main.scss
notlee: Good spot, I'll add that here 👍

components/o-video/src/scss/_info.scss
notlee: No, an [unquoted string](https://sass-lang.com/documentation/values/strings/#unquoted) is fine here.

components/o-private-foundation/main.scss
notlee: I'll pop these in the next PR actually, I don't think it's needed at this point

PR #1891, opened 2024-12-03
2025 release – typography – everything else [OR-1050]

notlee:


components/o-editorial-typography/main.scss
```diff
@@ -1,6 +1,5 @@
-@import '@financial-times/o-typography/main';
+@import '@financial-times/o-private-foundation/main';
 @import '@financial-times/o-spacing/main';
```
frshwtr, line 2: Looks like this was missed when working on spacing.

notlee: 👍 Just removed that as part of this PR

notlee: Oh no I didn't, I different call to oSpacingByName 🙈

notlee: Done!

components/o-forms/src/scss/modifiers/_state.scss
```diff
@@ -6,11 +6,8 @@
 	.o-forms-input__state {
 		// the loading spinner and tick icons for state are mismatched in terms of size/padding,
 		// the custom line-height aligns them as closely as possible (with minimal style changes)
-		@include oTypographySans(
-			$scale: -1,
-			$line-height: 1.75,
-			$include-font-family: false
```
frshwtr, line 12: This mixin had `include-font-family` set to `false` which does not include `font-family` in this case.. `oPrivateTypographySans` will include `font-family` when the mixin is used.

Is this a correct update?

notlee: Yeah that's fine, I've optimised for API simplicity. Its intention was to reduce bundle size slightly, but we'll see much bigger gains through o-buttons refactoring and ultimately moving to o3

components/o-private-foundation/main.scss
```diff
@@ -19,5 +19,12 @@
 		font-display: swap;
 	}
 
-	// @todo include italic font
+	@font-face {
```
frshwtr, line 22: comment mentions italic font but this looks like a standard definiton for financier, does italic still need including?

notlee: It does not! Italics are not used in o2 https://github.com/Financial-Times/n-ui-foundations/blob/main/typography/main.scss

frshwtr (approved):

PR #1895, opened 2024-12-12
Fix some errors

notlee:
See commit messages


templates/test-workflow.yml
```diff
@@ -51,7 +51,6 @@ jobs:
       - name: Install node modules
         run: npm ci
       - run: npm run -w ${{ matrix.component }} lint --if-present
-      - run: npm run -w ${{ matrix.component }} build --if-present
```
frshwtr, line 54: should this be removed?

notlee: Yeah, it happens as part of the test script now: [scripts/component/test.bash](https://github.com/Financial-Times/origami/pull/1895/files#diff-a2bca38b55758646d1538532d688bfe586a2281879a32eb9a038bbec20d2b7e7)

We need to build as part of the test so `ally` and `validate-component-demo-html` run successfully. 

This allows us to run `npm run test` and actually get the same results as ci

frshwtr (approved):

PR #1919, opened 2025-01-09
Add `display` typography styles [OR-1205]

notlee:
**Add token attributes to the tooling export**
The token export for tooling includes extra attributes to make it easier to identify groups of type tokens after the token tree has been flattened.

**Add a utility class for display typography usecases**
- Ensure the default font family has low specificity with `:where`
- Add storybook demos which point to the website. Otherwise we will
just be repeating information and doubling work. We may want to revisit
this.

**Refactor the Preview component**
A new PreviewPane component is responsible for the UI of the preview
interface. This makes it more flexible. It can now be used for smaller
previews, such as for utility classes which aren't a component and
do not have JSX.

The Preview component retains logic to render a JSX component,
fetch its JSX from a given file path for a code snippet, and
generates a HTML code snippet from there.

**Add display type usecase docs to the website**

This differs to the Figma view because:

1. We need a code snippet for the class.
2. We need to document each of the CSS Custom Properties which make up the composite token. This is handy to apply these styles without needing to make markup changes.

Screenshot 2025-01-09 at 18 17 24
Screenshot 2025-01-09 at 18 17 27


apps/for-everyone-website/src/components/TypeUsecase.astro
```diff
@@ -0,0 +1,47 @@
...
+});
+const TypeUsecaseTokensByItem = Object.groupBy(TypeUsecaseTokens, (token) => token.attributes?.subitem);
+
+---
+
+{
+
+
+    Object.entries(TypeUsecaseTokensByItem).map(([key, tokens]) => {
+        console.log(tokens);
```
frshwtr, line 20: Is this needed?

apps/for-everyone-website/src/components/utils/PreviewFrame.astro
```diff
@@ -0,0 +1,191 @@
...
+	/* Rounded sliders */
+	.slider.round {
+		border-radius: 16px;
+	}
+
+	.slider.round:before {
+		border-radius: 50%;
+	}
+
+	/* LABEL ADDED BY BRIAN */
```
frshwtr, line 140: Who is brian?

apps/for-everyone-website/src/content/docs/guides/typography.mdx
```diff
@@ -319,3 +292,42 @@ Keeping typography consistent and sticking to logical hierarchies ensures that e
...
+			<TypeScale brand={props.brand} scaleFamily="georgia" />
+		</TabItem>
+	</BrandedContent>
+</Tabs>
+
+## Font Weight
+
+Font weight can be used to express hierarchy or urgency of information.
+
+We recommend using our standard [UI typography](#ui-typography) <BrandedContent currentBrand={props.brand} brands="core, professional, sustainable-views, whitelabel">or [Editorial typography](#editorial-typography)</BrandedContent> styles where possible. For custom typography, align to these standard styles where possible. For example when using Metric2 for UI use `semibold` over other bold weights.
```
frshwtr, line 326: `For example when using Metric2 for UI use `semibold` over other bold weights.` Why?

components/o3-foundation/stories/core/proffesional/usecases.mdx
```diff
@@ -0,0 +1,16 @@
+import { Meta } from '@storybook/blocks';
```
frshwtr, line 1: There is a typo in the directory path `proffesional`.

components/o3-foundation/stories/core/proffesional/usecases.mdx
```diff
@@ -0,0 +1,16 @@
...
+
+<Meta title="Core/Professional/o3-foundation/o3-typography/Usecases" />
+
+# Typography Usescases
```
frshwtr, line 5: These are nice additions to Storybook :) A note that in future we may run into increased maintenance where one update will need to be applied across multiple files.

It might be possible to extract the contend into a Markdown file and import in `mdx`



```mdx
import { Description } from '@storybook/addon-docs/blocks';
import Readme from '.README.md';

<Meta title="Documentation|Readme" />

<Description>{Readme}</Description>
```

apps/for-everyone-website/src/components/TypeUsecase.astro
notlee: No, thanks

apps/for-everyone-website/src/content/docs/guides/typography.mdx
notlee: 👍 Let's save this kind of feedback for a review of type guidelines when we're done. I'm adding new usecases here but not removing any old stuff yet. We have another ticket to apply these usecases, which I think should include removing those that have been replaced.

components/o3-foundation/stories/core/proffesional/usecases.mdx
notlee: Good spot. Let's rename in another PR to keep the changelog down

components/o3-foundation/stories/core/proffesional/usecases.mdx
notlee: Yeah good shout, it's a real pain. I'm also wondering whether with the new tags/filtering feature we can tag stories with brands and avoid a whole load of repetition. I'm going to leave this pre migration though – unless by some miracle we find ourselves with some time

frshwtr (approved):

PR #1923, opened 2025-01-14
Begin applying use-case tokens to o3 components.

notlee:
- Big number
- Editorial body 
- Byline
- Headline Large (renamed Display)

This renames the JSX component `Headline` to `Heading`. It accepts a prop "type" of "headline", "chapter", etc. As Headline is a type of heading,`<Headline type="headline" >`, this was confusing. I would prefer these to be separate components as the style may be applied outside of a heading context, but this rename is pragmatic given our hard deadline.

I chose to stop and open a PR here since it was getting _chunky_ and will continue applying use-case tokens to other `Heading` types in a follow on PR.


components/o3-editorial-typography/stories/whitelable/headline.stories.tsx
```diff
@@ -24,8 +24,8 @@ export default {
 	},
 } as Meta;
 
-export const Heading = StoryTemplates.Heading;
-Heading.argTypes = {
+export const Headings = StoryTemplates.Heading;
+Headings.argTypes = {
 	type: {
 		options: ['headline', 'subheading', 'chapter', 'label'],
```
frshwtr, line 30: Should these have changed like with the other stories?

components/o3-editorial-typography/stories/sustainable-views/headline.stories.tsx
```diff
@@ -1,12 +1,12 @@
 import type {Meta} from '@storybook/react';
```
frshwtr, line 1: May need to update file name to `heading.stories.tsx`

components/o3-editorial-typography/stories/whitelable/headline.stories.tsx
j-mes: Should it be `whitelabel` or is `whitelable` a thing? (Not in the code but the file directory)

apps/for-everyone-website/src/components/editorial-typography/HeadlineLarge.astro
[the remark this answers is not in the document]
j-mes: Oh I see a `Display` file elsewhere… hmmm.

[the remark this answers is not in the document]
j-mes: Me again. Uh. Different place 😂 so I assume this should be renamed to `Display` for Astro to compile later on.

components/o3-editorial-typography/stories/whitelable/headline.stories.tsx
notlee: No this shouldn't have changed as it doesn't include a `headline-large` option

notlee:
@frshwtr those tokens are missing because I haven't updated other Heading tokens yet, just Display

frshwtr (approved):

PR #1924, opened 2025-01-14
feat: introduce UI feedback icons

frshwtr:
## Describe your changes

Icons such as 'error' were missing from the recent iconography work, causing the error icon on o3-form's error message to not be present. These have been introduced.

[Designs](https://www.figma.com/design/6HPuyTvaqZVHZ7hP8LeZP9/Icons-Audit---Working-file?node-id=4397-1713&p=f&t=uYMXbiafaFnsjBbb-0)


notlee:
@frshwtr [conversations are ongoing design side](https://financialtimes.slack.com/archives/C078EBTL9DM/p1736777075101259) as to whether to add them. Could we add as a private o3-form token for now? It will be very hard to remove from our standard set if we go ahead with warming/error icons in a different form

notlee:
Does this also update the info icon? We need more buy-in from the wider design team before doing this right after the audit

notlee (changes requested):
See above

frshwtr:
> Does this also update the info icon? We need more buy-in from the wider design team before doing this right after the audit
What I can do is only introduce the error icon as a private icon as part of o3-form. Which can then be moved to private once we have sign-off

notlee (approved):

PR #1926, opened 2025-01-14
Apply more use-case tokens to o3 components. 

notlee:
- Topic tag 
- Standfirst
- Drop cap


frshwtr:
I can't find a story for DropCap, can we add one?

frshwtr (approved):

PR #1939, opened 2025-01-20
chore: deleted link body-link and body-content-link tokens for all li…

notlee:


notlee:
Oh no. I opened the PR and so can't review. Deyan, I approve

PR #1940, opened 2025-01-20
chore: Upgrade o-teaser-collection to o3 type usecases

notlee:


components/o-teaser-collection/src/scss/_mixins.scss
```diff
@@ -40,7 +40,17 @@
 	$half-width: index($sizes, 'half-width');
 	$small: index($sizes, 'small');
 
-	@include oPrivateTypographySans(3, $weight: 'regular');
+	font-family: oPrivateFoundationGet(
```
frshwtr, line 43: I think scale 3 is equivalent to `title-md`. Is `title-sm` correct here?

notlee: I think this is right, the original scale is scale 2 in Metric2. Don't forget `oPrivateTypographySans` takes the given scale and minuses by one, as it replaces `oTypographySans` which was based on Metric not Metric2.

frshwtr (approved):

PR #1942, opened 2025-01-22
o-forms corrections and o-message build error fix

notlee:
* Apply missing radius to o-forms checkbox.
* Correct o-forms date input width which is too small with larger text.
* Update o-forms demos / stories to show latest button design, since o-buttons will be deprecated.
* Correct token use in o-message since we deleted link tokens. I noticed the build failed.

Screenshots: https://financialtimes.slack.com/archives/G03ACR1MJ/p1737543643692009?thread_ts=1737455978.705169&cid=G03ACR1MJ

### DIFF
--- a/components/o-forms/demos/src/data/pa11y.json
+++ b/components/o-forms/demos/src/data/pa11y.json
@@ -111,9 +111,7 @@
 						"label": "error-date-group-title"
 					},
 					"error": "Not a valid date",
-					"modifiers": [
-						"invalid"
-					]
+					"modifiers": ["invalid"]
 				},
 				"title": {
 					"main": "Invalid date input"
@@ -136,7 +134,6 @@
 						"value": "24",
 						"id": "error-month",
 						"name": "error-month"
-
 					},
 					{
 						"type": "text",
@@ -145,7 +142,6 @@
 						"value": "2019",
 						"id": "error-year",
 						"name": "error-year"
-
 					}
 				]
 			},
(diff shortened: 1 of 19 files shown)


j-mes:
It seems like the tests are failing on this part.

```src/scss/shared/_single-inputs.scss
 57:4  ✖  Unexpected duplicate "line-height"   declaration-block-no-duplicate-properties```


If I go to [src/scss/shared/_single-inputs.scss](https://github.com/Financial-Times/origami/blob/2025-release-o-forms-updates/components/o-forms/src/scss/shared/_single-inputs.scss#L57). It seems to be the `line-height` that's causing it to fail.

Are we defining the `line-height` somewhere much earlier and elsewhere and this is a duplicate of the same value?

File: components/o-forms/demos/src/data/pa11y.json [hunk 1]
```diff
@@ -425,11 +416,9 @@
 				},
 				"input": {
 					"type": "text",
-					"modifiers": [
-						"suffix"
-					],
+					"modifiers": ["suffix"],
 					"value": "",
-					"button": "o-buttons o-buttons--secondary o-buttons--big",
+					"button": "demo-submit-button",
```
j-mes, line 421: Is this correct due to the change?

components/o-forms/demos/src/demo.scss
```diff
@@ -31,7 +54,7 @@ body {
 $custom-theme: (
 	controls-base: 'o3-color-palette-crimson',
 	controls-checked-base: 'o3-color-palette-white',
-	controls-border: 'o3-color-palette-crimson'
+	controls-border: 'o3-color-palette-crimson',
```
j-mes, line 57: Do love a trailing comma!

PR #1944, opened 2025-01-22
chore: Apply type usecases to o-stepped-progress

notlee:
Apply a custom weight to the heavy varient. A redesign is required to fit the current type tokens, and this varient looks pretty borked in other ways (icon size).

Text becomes a little larger:
Screenshot 2025-01-22 at 14 13 13

The `strong` element of this "heavy" variant is a custom type style. I decided to stick with this as this variant is internal only, not our focus right now, and its icons are so small I wonder if it's used. This sets a stronger weight on the `detail` type token – stepping up in size to `body-base` and `body-highlight` is quite large, and may visually break longer copy. 
Screenshot 2025-01-22 at 14 14 07

### DIFF
--- a/components/o-stepped-progress/src/scss/_mixins.scss
+++ b/components/o-stepped-progress/src/scss/_mixins.scss
@@ -38,7 +38,17 @@
 /// @access private
 @mixin _oSteppedProgressBase() {
 	.o-stepped-progress {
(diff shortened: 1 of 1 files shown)


j-mes (approved):
LGTM :+1:

PR #1945, opened 2025-01-22
2025 release forms design tweaks

notlee:
- Reduce checkbox / radio label type size to body base. We've already done this by accident in o-forms. Leave text input, etc, at `body-lg` by design.
- Correct tense in show password field.


j-mes (approved):
LGTM :+1:

PR #1948, opened 2025-01-23
o-tracking add IntersectionObserver.threshold as opt

gyss:
## Description

I'm configuring the App Homepage `scrolldepth` tracking event using `oTracking.view.init()` in [this PR](https://github.com/Financial-Times/ft-app/pull/2878) by setting a selector that will trigger this event every time a `<section />` comes into the viewport. In the beginning, it was working as expected, but after some days I retested it and the `scrolldepth` events were not being sent as expected.

After some investigation, I realized that the problem was located in the **1.0** IntersectionObserver threshold defined [here](https://github.com/Financial-Times/origami/blob/4d2111699890b9cce57da3150401fbd722001773/libraries/o-tracking/src/javascript/events/component-view.js#L78). Looking [at the docs](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#threshold), this means that "A value of 1.0 means that the threshold isn't considered passed until every pixel is visible". This is not a problem in medium / big screens, since they can fit a whole `<section />` in the viewport (that's why it worked in the beginning), but for small screens, is rather difficult to fit the whole `<section />` in it.

I've considered changing the selector provided to this function by selecting only the first teaser of each `<section />`, but the payload sent to Spoor changes drastically.

So, to fix this, we can modify the threshold to a lover value so the event triggers with smaller portions of the `<section />`.

Relates to [this ticket](https://financialtimes.atlassian.net/browse/HETD-56)

### DIFF
--- a/libraries/o-tracking/src/javascript/events/component-view.js
+++ b/libraries/o-tracking/src/javascript/events/component-view.js
@@ -42,7 +42,7 @@ const decorateEventData = (eventData, viewedEl, opts) => {
  * Listen for view events.
  *
  * @alias view#init
- * @param {object} opts - To set custom category[String], selector[String], getContextData[Function]
+ * @param {object} opts - To set custom category[String], selector[String], getContextData[Function], intersectionObserverThreshold[Any]
  * @returns {undefined}
  */
 const init = (opts = {}) => {
@@ -75,7 +75,7 @@ const init = (opts = {}) => {
 		});
 	}
 
-	const observer = new IntersectionObserver(onChange, { threshold: [ 1.0 ] });
+	const observer = new IntersectionObserver(onChange, { threshold: opts.intersectionObserverThreshold || [ 1.0 ] });
 
 	elementsToTrack.forEach(el => observer.observe(el));
 };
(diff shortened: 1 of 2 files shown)


libraries/o-tracking/src/javascript/events/component-view.js
```diff
@@ -75,7 +75,8 @@ const init = (opts = {}) => {
 		});
 	}
 
-	const observer = new IntersectionObserver(onChange, { threshold: [ 1.0 ] });
+	const threshold = Array.isArray(opts.threshold) ? opts.threshold : [ 1.0 ]
```
metart43, line 78: I am not opposed to this change. Having control over `threshold` is a good idea and might be needed for some apps. I'd just pick a more descriptive name for this property. 
`threshold` name is clear when we initialize IO, but not as much when we initialise oTracking.

metart43 (approved):

libraries/o-tracking/src/javascript/events/component-view.js
gyss: any suggestion? something like `observerThreshold`?

metart43: Now that I read MDN docs I can see that threshold can be either array of numbers or a single number. https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver#threshold

I think if we are extending the possibility to set this, we need to take this into account

notlee (commented):
Hey Jose, this looks sensible to me. Please document the new option. 

I'll leave to the Data Platform team to approve, as they own o-tracking. You can get hold of them in `#data-support`.

metart43 (approved):

PR #1951, opened 2025-01-28
chore: Correct o-header font size

notlee:
- Where a px value was used directly, font size had visually increased due to Metric2
- Restores button radius for standalone buttons

**production / 2025-release currently (the problem)**
Screenshot 2025-01-28 at 15 52 17

**now**
- Slightly larger copy inline with standard use-cases (16px, but with metric2).
- Note form placeholder text is massive. This is by design but this specific scenario could use a review. Out of scope for this PR.
Screenshot 2025-01-28 at 16 26 13

### DIFF
--- a/components/o-header/main.scss
+++ b/components/o-header/main.scss
@@ -70,6 +70,9 @@
 						'theme': 'mono',
 					)
 				);
+			}
+			.o-header__top-button,
+			.o-header__nav-button {
 				border-radius: 0;
 			}
 		}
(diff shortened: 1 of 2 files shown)


j-mes (approved):
Looks good to me, and I'm wondering re: failing test. I'll check that out for you.

notlee:
@j-mes yeah I assumed that was an existing error that your PR fixes 🤞 I'll wait to merge this to see

PR #1869, opened 2024-11-13
2025 release

notlee:
A long running feature branch for a major 2025 release of design foundation changes.

PR #1966, opened 2025-02-17
feat: add fallback fonts

frshwtr:
## Describe your changes
* Adds fallback fonts for Metric, Financier Display, and Georgia tokens.

Note that this will only affect usages of tokens on web, [Figma does not use fallbacks](https://docs.tokens.studio/token-types/token-type-typography-font-fallbacks)


frshwtr:
Leaving this as draft for now. Need to verify these fallbacks (Specifically SCSS) work as expected.

j-mes:
I've been trying to review this… and with your brilliant suggestion on Slack to use `o-teaser` as a demo to check out if the fallback fonts are indeed working, with a legacy demo. I've tried a random one and noticed the font-family stack looked like this in the CSS file for the legacy demo. Is that intended?

```css
.o-teaser__timestamp {
  font-family: "metric 2 VF, sans-serif", sans-serif;
}
```

Surely it would work with the sans-serif at the end regardless of having it under the double quotes as well?

Though, the rest of the CSS is absolutely fine. Some examples below looking absolutely fine to me.

```css
.o-teaser {
  font-family: metric 2 VF, sans-serif;
}
```

```css
.o-teaser__meta {
  font-family: metric 2 VF, sans-serif;
}
```

The list goes on.

```css
.o-teaser--small.o-teaser--post .o-teaser__heading,
.o-teaser--small.o-teaser--post .o-teaser__heading a {
  font-family: "financier display VF, serif", sans-serif;
}
```
Also I'm not sure why this is like this? Is the serif fallback enough that it wouldn't need `sans-serif` at the end? Also might just be me looking into this a bit too much. I'm not sure if it's actual changes or that has always has been there all this time.

Other than that, it does look good to me. One for Lee too?

notlee:
Good spot @j-mes. I think we'll want to remove instances (in Sass and CSS) where we've set the fallback font manually. [Here's an example in o2](https://github.com/Financial-Times/origami/pull/1869/files#diff-8b59a0ce85275bb8850df3934708cea91e70e262d15eca26ed08df184ad41732R61) which is causing the teaser issue, and o3:
Screenshot 2025-02-17 at 13 45 48

components/o-teaser/src/scss/elements/_default.scss
```diff
@@ -77,6 +77,7 @@
 }
 
 @mixin _oTeaserTagPrefix {
+	@debug oPrivateFoundationGet('o3-type-detail-font-family');
```
notlee, line 80: **OOOOOOOOOOPPPSSSSSSS** 🚓

notlee (approved):

PR #1971, opened 2025-02-20
chore: Reverse o3 migration changes in o-quote and o-big-number

notlee:
These components have an o3 alternative and will be deprecated.


j-mes (approved):
LGTM :+1:

PR #1988, opened 2025-02-26
fix: Remove unused o-icons import

notlee:


j-mes (approved):
Looks good to me, it's just a removal of o-icons from the o-header? 😅

PR #2004, opened 2025-02-27
docs: Restructure o2 storybook

notlee:
- Update index page of composite storybook to reflect changes in o3 and our adoption schedule. We shouldn't refer to "o2" components as "legacy" when many are actively maintained, and will be used for the foreseeable future.
- Remove "Brandless" grouping. This is repeated with "Brandless/o-tracking" at a top level and other "Brandless" packages nested within each brand's storybook, including o-tracking repeated again. Instead, keep each component under each brand it supports. This also means it is visible when viewing a single brand's storybook in isolation.
- Group stories by support status. This makes it clear which components are supported and which have been deprecated for o3 alternatives. When we upgrade to storybook 8 [tags](https://storybook.js.org/docs/writing-stories/tags) would be a better option. This would allow component story titles to stay the same when support status changes (maintaining links) and allow components to be filtered by support status (as with the old Origami registry)

Before:
Screenshot 2025-02-27 at 15 28 22

After:
Screenshot 2025-02-27 at 15 28 04

### DIFF
--- a/apps/o2-storybook-composition/.storybook/main.ts
+++ b/apps/o2-storybook-composition/.storybook/main.ts
@@ -1,10 +1,7 @@
 import type {StorybookConfig} from '@storybook/react-webpack5';
 
 const config: StorybookConfig = {
-	stories: [
-		'../stories/*.stories.@(js|jsx|mjs|ts|tsx|mdx)',
-		'../../../libraries/o-tracking/stories/*.stories.@(js|jsx|mjs|ts|tsx|mdx)',
-	],
+	stories: ['../stories/*.stories.@(js|jsx|mjs|ts|tsx|mdx)'],
 	addons: [
 		'@storybook/addon-links',
 		'@storybook/addon-essentials',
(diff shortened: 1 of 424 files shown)


frshwtr:
Return of an error :(

```(node:36022) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /Users/ben.freshwater/github/origami/node_modules/unicorn-magic/package.json
    at exportsNotFound (node:internal/modules/esm/resolve:299:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:589:13)
    at resolveExports (node:internal/modules/cjs/loader:595:36)
    at Module._findPath (node:internal/modules/cjs/loader:685:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1172:27)
    at Module._resolveFilename (/Users/ben.freshwater/github/origami/node_modules/esbuild-register/dist/node.js:4794:36)
    at Module._load (node:internal/modules/cjs/loader:1012:27)
    at Module.require (node:internal/modules/cjs/loader:1271:19)
    at require (node:internal/modules/helpers:123:16)
    at Object.<anonymous> (./node_modules/globby/index.js:7:22)
```

frshwtr (approved):

PR #2012, opened 2025-03-07
fix: Remove `o3-editorial-typography-byline-location`.

notlee:
This has the potential to be a visually breaking change where `o3-editorial-typography-byline-location` is used outside of `o3-editorial-typography-byline`. We're happy with that given what we know about `o3-editorial-typography` use, vs. keeping that class around deprecated.

From Kara:
if we look at the example in the docs, "Joe Doe in London", it's marked up with Joe Doe as author and in London as location. but not all the text in the byline that isn't an author is a location, you can have bylines that are like "Dan McCrum in London and Antoine Gara in New York", where the "and" isn't part of a location. in the content pipeline data, this is structured as `author: Dan McCrum, text: in London and, author: Antoine Gara, text: in New York`, and the text nodes don't get any wrapper elements.

### DIFF
--- a/components/o3-editorial-typography/details.css
+++ b/components/o3-editorial-typography/details.css
@@ -1,202 +1,202 @@
 .o3-editorial-typography-topic-tag {
-    font-family: var(--o3-type-body-highlight-font-family);
-    font-size: var(--o3-type-body-highlight-font-size);
-    font-weight: var(--o3-type-body-highlight-font-weight);
-    line-height: var(--o3-type-body-highlight-line-height);
+	font-family: var(--o3-type-body-highlight-font-family);
+	font-size: var(--o3-type-body-highlight-font-size);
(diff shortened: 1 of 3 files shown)


components/o3-editorial-typography/details.css
```diff
@@ -1,202 +1,202 @@
...
+		--_o3-editorial-typography-byline-author-hover-inverse-color
+	);
 }
 
-.o3-editorial-typography-byline-location {
-    font-family: var(--o3-type-body-base-font-family);
-    font-size: var(--o3-type-body-base-font-size);
-    font-weight: var(--o3-type-body-base-font-weight);
-    line-height: var(--o3-type-body-base-line-height);
+.o3-editorial-typography-byline {
```
notlee, line 85: Gah, sorry for the indentation changes. Here's where the magic happens

frshwtr (approved):

PR #2019, opened 2025-03-11
fix: exclude anchor tags from type color selectors

frshwtr:
## Describe your changes

Anchor tags in wrappers need to have an `o3-typography-body-base` class somewhere in the cascade to receive text styles. 

When an anchor tag is outside another typographic tag, such as `p`, it may be necessary to apply `o3-typography-body-base` directly to the anchor tag. This class provides colour which may have precedence over any anchor styles from `o3-typography-wrapper`.

This PR adds additional selectors to typography-base classes so that anchor tags do not receive its colour styles.

### DIFF
--- a/components/o3-foundation/src/css/components/typography/usecases.css
+++ b/components/o3-foundation/src/css/components/typography/usecases.css
@@ -132,7 +132,7 @@
 
 /* Default body colours */
 
-:where(
+:where(:not(a)):where(
 		[class^='o3-type-body'],
 		.o3-type-detail,
 		.o3-type-label,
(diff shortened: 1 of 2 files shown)


components/o3-foundation/src/css/components/typography/usecases.css
```diff
@@ -132,30 +132,31 @@
 
 /* Default body colours */
 
-:where(
+:where(:not(a)):where(
```
j-mes, line 135: This looks good to me. As an explainer to myself, this will look for all the things but not the links and apply the styling there.

PR #2021, opened 2025-03-13
Chore: introduced success colours and updated error colours to pass a…

MaryGodservant:
## Describe your changes

1. Evaluated with @frshwtr and @MaryGodservant regarding usage of `error` and we both questioned why there's no `success` equivalent that can be used alongside with `error` states as on a form it would possibly be used.
2. Discussed further regarding naming tokens in the dictionary. Naming things are hard!
3. Agreed on `foreground` for the text colour and/or the border colours for the states either success of error.
4. Agreed on `background` for the background colour of the `success` or `error` states.
5. Updated the dictionary so that it would be utilised in different places with the design tokens.
6. Would be a breaking change in Figma.

[OR-1396](https://financialtimes.atlassian.net/browse/OR-1396)

## Link to Figma designs

TBC

[OR-1396]: https://financialtimes.atlassian.net/browse/OR-1396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/dictionary/tokens/core/use-case/color.json
```diff
@@ -141,16 +141,20 @@
...
+          "type": "color"
+        },
         "error-background": {
-          "value": "rgba(204,0,0,0.06)",
+          "value": "#F7EBE5",
           "type": "color"
         },
-        "error-text": {
-          "value": "{o3.color.palette.crimson}",
+        "sucess-text": {
```
j-mes, line 152:
```suggestion
        "success-text": {
```

apps/dictionary/tokens/core/use-case/color.json
```diff
@@ -141,16 +141,20 @@
...
+          "value": "#00572C",
           "type": "color"
         },
-        "error": {
-          "value": "{o3.color.palette.crimson}",
+        "success-background": {
```
j-mes, line 156: Same comment as above.

```
{o3.color.use-case.success}
```

components/o3-foundation/src/css/tokens/core/_variables.css
```diff
@@ -182,8 +185,6 @@
   --o3-color-use-case-button-default: var(--o3-color-palette-teal);
   --o3-color-use-case-button-hover: var(--o3-color-palette-teal-40);
   --o3-color-use-case-button-pressed: var(--o3-color-palette-teal-20);
-  --o3-color-use-case-error-text: var(--o3-color-palette-crimson);
-  --o3-color-use-case-error: var(--o3-color-palette-crimson);
```
notlee, line 186: This could be a visually breaking change as we've removed public design tokens (and their CSS custom properties). However it's not a breaking API change, because CSS using this will continue to work. How do you propose approaching this:

1. Check Origami useage and make a call to remove (though it risks being visually breaking)
2. Add these CSS Custom Properties back manually and mark deprecated
3. Revert to the previous token naming

notlee: In this case in the early days of adoption I'd probably be lenient and go with (1). 

Another thing to consider is that Figma variables for these will also be deleted. Are we happy to publish a breaking change in Figma?

apps/dictionary/tokens/whitelabel/use-case/color.json
```diff
@@ -132,12 +132,16 @@
...
           "type": "color"
         },
-        "error-text": {
+        "error-foreground": {
           "value": "{o3.color.palette.crimson}",
           "type": "color"
         },
-        "error": {
-          "value": "{o3.color.palette.crimson}",
+        "success-background": {
```
notlee, line 139: I'm seeing whitelabel changes here in tokens but no whitelabel css updates – do tokens need to be rebuilt?

notlee: Does the whitelabel error background need to change too or is that ok?

components/o3-foundation/src/css/tokens/core/_variables.css
j-mes: Hmm, I think checking for Origami useage would be a good call here. However I do like the second option as well. It'd give people time to migrate off.

apps/dictionary/tokens/whitelabel/use-case/color.json
j-mes: The tokens were missing. I've remedied and added them in all of the affected places as well.

components/o3-foundation/src/css/tokens/core/_variables.css
j-mes: @MaryGodservant Are you happy with the breaking changes in Figma for this use case?

PR #2031, opened 2025-03-21
feat: add neutral theme to primary/ghost buttons

j-mes:
## Describe your changes

Add neutral theme for the button themes `primary` and `ghost` for Token Studio. 

The secondary buttons will be at a later date as I and @MaryGodservant feel it's too similar to the other themes and would warrant a further design review.

### Ad-hoc changes
I have also included `link` icon to be able to be used by `o3-button` now so it will allow people to use it onwards.

[OR-1231](https://financialtimes.atlassian.net/browse/OR-1231)

## Link to Figma designs

[Figma](https://www.figma.com/design/VVM0PixrY3IRZq2ZUTdWfU/branch/Mk7yARLvXaujjZwBeeZayf/%F0%9F%92%A0-Core---Origami-(o3)?m=auto&node-id=4059-1949&t=ysydoao5tfCVdEWK-1)

[OR-1231]: https://financialtimes.atlassian.net/browse/OR-1231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


notlee:
@j-mes The colours and tokens look good to me! There are more things to do to enable this theme:


1. Update the CSS to add the required CSS declarations using these custom properties See `main.css`. However, as this is core brand specific, we should add to `src/brands/core.css` to avoid bloating other brand's css.
2. Buttons tsx props need updating to accept the new theme.
3. Button Storybook config needs updating to display the new theme. For some reason I can't remember the option is configured manually. Although the tsx can't be branded, and the theme will show as an option for all tsx users, could we at least only show this option in the core brand storybook?
4. The website needs updating also to show this theme, just for the core brand!

Let me know if you have any questions about those things. Of course over time we should also look to reduce these steps which I think we could do with Storybook changes and token structure improvements 😄 but not for now

notlee:
In addition, it's the only theme that doesn't support all button types. We should:
* Update the readme and make the theme/brand support clear

Wild idea. It would fall back to the standard secondary theme as it's not supported. We don't want people using this. What if we added CSS to highlight the error of secondary / neutral being used together?
Screenshot 2025-03-21 at 14 44 52

```css
/* SECONDARY NEUTRAL: Unsupported */
.o3-button.o3-button--secondary[data-o3-theme='neutral'],
:where([data-o3-theme='neutral']) .o3-button.o3-button--secondary {
	border: 1px dashed var(--o3-color-palette-crimson);
	opacity: 0.6;
	pointer-events: none;
}
```

Unsure how I feel about shipping effectively debug code to users... we wouldn't want to do this much if at all...

j-mes:
> In addition, it's the only theme that doesn't support all button types. We should:
> 
> * Update the readme and make the theme/brand support clear
> 
> Wild idea. It would fall back to the standard secondary theme as it's not supported. We don't want people using this. What if we added CSS to highlight the error of secondary / neutral being used together? Screenshot 2025-03-21 at 14 44 52
> 
> ```css
> /* SECONDARY NEUTRAL: Unsupported */
> .o3-button.o3-button--secondary[data-o3-theme='neutral'],
> :where([data-o3-theme='neutral']) .o3-button.o3-button--secondary {
> 	border: 1px dashed var(--o3-color-palette-crimson);
> 	opacity: 0.6;
> 	pointer-events: none;
> }
> ```
> 
> Unsure how I feel about shipping effectively debug code to users... we wouldn't want to do this much if at all...

I think it's a good deterrent for now. I would presume if the documentation is clear with that this theme is currently not supporting secondary and that it will be revisited at a later date then no reason not to have something like that above because an engineer will proof-read their work and notice that happening then realise that they need to do something about it. Better that than shipping something in production without actual CSS/behaviour for it.

components/o3-button/src/types/index.ts
```diff
@@ -18,6 +18,7 @@ export interface ButtonProps {
 		| 'search'
 		| 'refresh'
 		| 'cross'
+		| 'link';
```
frshwtr, line 21:
```suggestion
		| 'link'
```

PR #2048, opened 2025-03-26
fix: improve specificity on neutral theme for primary/ghost

j-mes:
## Describe your changes

An issue has come up with CSS specificity. The `whitelabel` CSS is further up the source order due to using the styling in the `main.css` file and is overriding the styling of the neutral theme when being loaded in by either `core.css` or `professional.css` which is further down the import order (as spotted by @notlee).

I have added `TODO` comments in the CSS so that we can review and change things in the setup further down the line.

Also some additional changes to the documentation regarding using the neutral theme with the primary or ghost buttons courtesy of @notlee's excellent feedback.

Also it was missing some components for the website to be able to display neutral theme correctly.

Also some bug fixes with the secondary button theme dropdown for the website.


components/o3-button/main.css
```diff
@@ -289,6 +329,35 @@
 	}
 }
 
+/* GHOST NEUTRAL */
+/* This neutral theme is currently supported only in two brands: Core and Professional. */
+/* TODO: Review moving back to the Core & Professional brand CSS files in the future */
```
notlee, line 334: Can we remove `TODO`? We don't track work here and may never do it. TODOs from 5 years ago aren't fun 😅

j-mes: Alrighty I'll remove the `TODO` comments :)

notlee (approved):
Approved! Let's follow up with:

1. Updating the default ghost colour to be transparent (like a ghost), as with other themes.
2. Removing the `neutral` selector from the `secondary` theme dropdown.

Screenshot 2025-03-26 at 17 24 21

Screenshot 2025-03-26 at 17 25 09

Issue #2050, opened 2025-03-27
Sticky header not working for simplified headers (o-header)


petrklus:
We are using a simplified version of the `o-header` and we are experiencing issues with the sticky behaviour.

We have narrowed the issue down to the following line:

https://github.com/Financial-Times/origami/blob/027b7167f3a49a2b637f33e25d302da44320d039/components/o-header/src/js/sticky.js#L29

As we do not have any element with ID `o-header-search-sticky`, the `stickyHeaderContainer` is null and thus `stickyHeaderContainer.getAttribute('aria-hidden');` fails.

We have fixed this temporarily by adding an element with said ID into our sticky header markup:

```
<div id="o-header-search-sticky" style="display: none;"></div>
```

Which fixes the sticky header initialisation issue.

PR #2055, opened 2025-03-27
fix: Prevent an error if no sticky search button is used.

notlee:
subs.ft.com heavily modifies the header. It uses a sticky header without search interface.

Generally we don't support this level of customisation for components, and would expect things could break in minor/patch releases without notice. I'd have recommended creating a custom component here from scratch. We might want to consider an Origami header component that centres a logo and allows for custom items left/right, but it's unlikely to make it on our roadmap soon. https://origami.ft.com/getting-started/customisation/

Despite that, this is a small issue, so let us be helpful and check for the presence of search and fail gracefully. We have recommended to pin to the patch release to reduce risk of future changes breaking unsupported customisations.

Closes: https://github.com/Financial-Times/origami/issues/2050

### DIFF
--- a/components/o-header/src/js/sticky.js
+++ b/components/o-header/src/js/sticky.js
@@ -1,14 +1,19 @@
-import { debounce } from '@financial-times/o-utils';
+import {debounce} from '@financial-times/o-utils';
 
-function init (headerEl) {
+function init(headerEl) {
 	if (!headerEl.hasAttribute('data-o-header--sticky')) {
 		return;
 	}
 
-	function hideStickyHeaderContainer ({stickyHeaderContainer, searchIcon, isActive, isHeaderExpanded}) {
+	function hideStickyHeaderContainer({
+		stickyHeaderContainer,
+		searchIcon,
+		isActive,
+		isHeaderExpanded,
+	}) {
 		if (!isActive && isHeaderExpanded) {
 			stickyHeaderContainer?.setAttribute('aria-hidden', !isActive);
-			stickyHeaderContainer.classList.remove('o-toggle--active')
+			stickyHeaderContainer.classList.remove('o-toggle--active');
 			searchIcon?.setAttribute('aria-expanded', isActive);
 		}
 	}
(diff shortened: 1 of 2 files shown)


components/o-header/src/js/sticky.js
```diff
@@ -18,35 +23,52 @@ function init (headerEl) {
...
 		const stickyHeaderId = '#o-header-search-sticky';
 		const scrollDepth = window.pageYOffset || window.scrollY;
 		const isActive = scrollDepth > viewportOffset;
 		const stickyHeaderContainer = headerEl.querySelector(stickyHeaderId);
-		const searchIcon = headerEl.querySelector(`[aria-controls="${stickyHeaderId.slice(1)}"]`)
-		const isHeaderExpanded = stickyHeaderContainer.getAttribute('aria-hidden');
+		const searchIcon = headerEl.querySelector(
+			`[aria-controls="${stickyHeaderId.slice(1)}"]`
+		);
+		const isHeaderExpanded = stickyHeaderContainer
```
notlee, line 36: Prettier. This is a one line change, here.

components/o-header/src/tsx/sticky.tsx
```diff
@@ -82,53 +82,5 @@ const StickySearch = () => {
...
-								</span>
-								<span>Close</span>
-							</button>
-						</span>
-					</label>
-				</form>
-			</div>
-		</div>
-	);
+	return <></>;
```
frshwtr, line 85: Should all of this been removed?

notlee: Oh woops 😅 Yeah

j-mes:
@notlee is this still needed? 😅

notlee:
@j-mes Yeah, in that it's more resilient, but no one is waiting on it. Thanks for the reminder. I'm probably not going to get to this before leaving so feel free to pick up or close

joelcarr:
Closing this PR and noted it in our backlog: https://financialtimes.atlassian.net/browse/OR-2116

PR #2056, opened 2025-03-27
chore: add filter for secondary button on neutral theme

j-mes:
## Describe your changes

Add a filter on the dropdown for the secondary button preview to exclude the `neutral` theme from the options.

## Additional context

| JIRA ticket                                                    | Figma design                                                                |
| -------------------------------------------------------------- | --------------------------------------------------------------------------- |
| [OR-xxxx](https://financialtimes.atlassian.net/browse/OR-xxxx) | [Figma: Name](https://www.figma.com/design/path/to/your/figma/sharing/link) |

### DIFF
--- a/apps/for-everyone-website/src/components/button/preview/Types-Secondary.tsx
+++ b/apps/for-everyone-website/src/components/button/preview/Types-Secondary.tsx
@@ -11,6 +11,11 @@ function ButtonPreview() {
 	);
 }
 
-export const themes = buttonThemes;
+export const themes = Object.fromEntries(
+	Object.entries(buttonThemes).map(([key, value]) => [
+		key,
+		value.filter(theme => theme !== 'neutral'),
+	])
+);
 export const filePath = 'src/components/button/preview/Types-Secondary.tsx';
 export {ButtonPreview as preview};


apps/for-everyone-website/src/components/button/preview/Types-Secondary.tsx
```diff
@@ -11,6 +11,11 @@ function ButtonPreview() {
 	);
 }
 
-export const themes = buttonThemes;
+export const themes = Object.fromEntries(
```
notlee, line 14: I see `buttonThemes.ts` is used in a bunch of Previews, but this filter only applies for `Secondary`. I think this could lead us to reintroduce the bug in a future Preview, or cause more confusion regarding button Previews behaving differently. 

Is there a more generic approach we could take?

`ButtonPreview.astro` wraps each individual Preview with Figma and Storybook links, it ultimate sets the `themes` prop. Which it gets from individual button previews, which all in turn import `buttonThemes.ts`. Perhaps we could delete `buttonThemes.ts` and remove them `themes` export from all the individual button preview? I'm thinking `ButtonPreview.astro` could define supported themes. It could also have a function that accepts type and brand to return supported themes, since it has all that information.

j-mes: I shall investigate and see if I can figure this out too!

PR #2057, opened 2025-03-27
feat: Add support-inverse colour use-case.

notlee:
This aligns with our other body colour usecases, which have an inverse alternative.

I've updated our Style Dictionary config to ensure a hex is generated from a token which is a colour mix.

Tested across brands on the website locally:
Screenshot 2025-03-27 at 16 58 07

### DIFF
--- a/apps/dictionary/src/config/styledictionary.js
+++ b/apps/dictionary/src/config/styledictionary.js
@@ -7,11 +7,16 @@ import {transformSVG} from '../transforms/transformSVG.js';
 import {tintGroup} from '../transforms/tint-group.js';
 import {privatePrefix} from '../transforms/private-prefix.js';
 
-register(StyleDictionaryPackage);
+register(StyleDictionaryPackage, {
+	'ts/color/modifiers': {
+		format: 'hex',
+	},
+});
 
-StyleDictionaryPackage.format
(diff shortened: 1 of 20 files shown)


apps/dictionary/src/config/styledictionary.js
```diff
@@ -7,11 +7,16 @@ import {transformSVG} from '../transforms/transformSVG.js';
 import {tintGroup} from '../transforms/tint-group.js';
 import {privatePrefix} from '../transforms/private-prefix.js';
 
-register(StyleDictionaryPackage);
+register(StyleDictionaryPackage, {
+	'ts/color/modifiers': {
+		format: 'hex',
```
notlee, line 12: Colour mix tokens should be hex not rgba

apps/dictionary/tokens/core/use-case/color.json
```diff
@@ -65,28 +65,32 @@
...
-        "heading-inverse": {
+        "support-inverse": {
           "text": {
             "value": "{o3.color.palette.white}",
-            "type": "color"
+            "type": "color",
+            "$extensions": {
+              "studio.tokens": {
+                "modify": {
+                  "type": "mix",
```
notlee, line 87: A mix of white and slate!

PR #2066, opened 2025-04-04
feat: Add a pagination pattern to show the current page only.

notlee:
Pagination: Due to concerns with bot scraping and costs, we need to avoid the current Origami pagination pattern which includes multiple page links and a link to the last page. This impacts stream and search pages. Since it's going to impact multiple pages, we decided to quickly implement a new pagination variant in Origami with a forward/back button, and current page only. We prefer this from a reader perspective also so win win win.

This could be improved to improve efficiency – we're still calculating desktop/mobile pages to show when we're only showing the current page – and provide an api which doesn't need to know _all of the pages_ in advance. However due to time constraints and getting into a bit of a Storybook mess, I have chosen a more simple route of adding a param to the current JSX component.

Screenshot 2025-04-04 at 17 07 11

### DIFF
--- a/apps/for-everyone-website/src/components/button/preview/ButtonPagination-inverse.tsx
+++ b/apps/for-everyone-website/src/components/button/preview/ButtonPagination-inverse.tsx
@@ -5,6 +5,7 @@ function ButtonPaginationPreview() {
 		<>
 			<meta itemProp="@preview" />
 			<ButtonPagination
+				currentPageOnly={false}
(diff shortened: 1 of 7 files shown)


frshwtr:
Visually breaking change? Unsure of usage so might be able to sneak it as a feature change?

notlee:
@frshwtr This isn't visually breaking, by default Pagination is the same. There is a new prop to change the behaviour to optionally show a single page only.

frshwtr (approved):

PR #2082, opened 2025-04-17
docs: Fix o3 storybook build

notlee:
Correct figma links and remove some redundant code


components/o3-foundation/stories/story-templates.tsx
```diff
@@ -106,7 +98,7 @@ export function ColorPaletteMetaGenerator(brand: BrandType): Meta {
 		component: ColorPalette,
 		decorators: [
 			Story => (
-				<div data-o3-brand={brand}>
+				<div data-o3-brand={brand.split('-')[0]}>
```
frshwtr, line 101: This breaks the spacing demo for `sustainable-views`, `data-o3-brand` ends up being `sustainable`
image

notlee: Booo! Good catch

frshwtr (approved):

PR #2087, opened 2025-04-23
feat: Add sparkles helper class for o3-button

notlee:
This is used for Ask AI / FT Professional features.

I removed a list of icons from the README which was out of date. Instead rely on Storybook and the TSX template.


MarjanMagharehi (approved):
Appreciate the addition!

PR #2092, opened 2025-04-28
fix: handle email in use

leannecornish-ft:
## Describe your changes
We have found a bug that causes users that have previously had an FT account to comment with their old coral account. We will be adding an additional check to the auth endpoint in next-comments-api, which will return a 403 when an old coral account is found. This PR checks for this status and returns a user-friendly error to the client.

## Additional context

| JIRA ticket                                                    | Figma design                                                                |
| -------------------------------------------------------------- | --------------------------------------------------------------------------- |
| Parent: [CI-2823](https://financialtimes.atlassian.net/browse/CI-2823) |  |
| Sub: [CI-2827](https://financialtimes.atlassian.net/browse/CI-2827) | |

[CI-2823]: https://financialtimes.atlassian.net/browse/CI-2823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CI-2827]: https://financialtimes.atlassian.net/browse/CI-2827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/o-comments/src/js/stream.js
+++ b/components/o-comments/src/js/stream.js
@@ -33,6 +33,7 @@ class Stream {
 		this.isSubscribed = false;
 		this.isTrialist = false;
 		this.onlySubscribers = opts.onlySubscribers;
+		this.emailInUse = false;
 	}
 
 	init () {
@@ -112,6 +113,7 @@ class Stream {
 				this.isSubscribed = response?.isSubscribed;
 				this.isTrialist = response?.isTrialist;
 				this.isRegistered = response?.isRegistered;
+				this.emailInUse = response?.emailInUse;
 			})
 			.catch(() => {
 				return false;
@@ -326,7 +328,22 @@ class Stream {
 					View our full ${this.options.linkSubscribe ? `<a class="linkSubscribe" href='${this.options.linkSubscribe}'>subscription packages</a>` : `subscription packages` } to join the conversation.
 				</p>
 			`;
-			customMessageContainer.innerHTML = this.isTrialist ? messageForTrial : this.isRegistered ? messageRegistered : messageForAnonymous;
+			const messageForEmailInUse = `
+			<h3>There is a problem configuring your commenting account</h3>
+				<p>
+					Please contact ${this.options.linkContact ? `<a class="linkContact" href='${this.options.linkContact}'>Customer Care</a>` : `Customer Care`} referencing issue code E403.
+				</p>
+			`;
+			if (this.isRegistered) {
+				customMessageContainer.innerHTML = messageRegistered;
+			} else if (this.isTrialist) {
+				customMessageContainer.innerHTML = messageForTrial;
+			} else if (this.emailInUse) {
+				customMessageContainer.innerHTML = messageForEmailInUse;
+			} else {
+				customMessageContainer.innerHTML = messageForAnonymous;
+			}
+			  
 			// this content is attached after oTracking.init is called therefore set data-trackable to the links doesn't work, therefore we raise an event when click on it
 			customMessageContainer.querySelector('.linkSubscribe')?.addEventListener('click', (event) => {
 				event.preventDefault();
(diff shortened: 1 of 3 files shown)


components/o-comments/src/js/stream.js
```diff
@@ -326,7 +328,22 @@ class Stream {
 					View our full ${this.options.linkSubscribe ? `<a class="linkSubscribe" href='${this.options.linkSubscribe}'>subscription packages</a>` : `subscription packages` } to join the conversation.
 				</p>
 			`;
-			customMessageContainer.innerHTML = this.isTrialist ? messageForTrial : this.isRegistered ? messageRegistered : messageForAnonymous;
+			const messageForEmailInUse = `
+			<h3>There is a problem configuring your commenting account</h3>
+				<p>
+					Please contact ${this.options.linkContact ? `<a class="linkContact" href='${this.options.linkContact}'>Customer Care</a>` : `Customer Care`} referencing issue code E403.
+				</p>
+			`;
```
umbobabo, line 336: Thought: not your fault but all these consts instantiated and not even used are a missed optimisation. Everytime a const is create it's using memory and then taking resources because of the Garbage collection that needs to follow. See comment below...

components/o-comments/src/js/stream.js
```diff
@@ -326,7 +328,22 @@ class Stream {
...
+			if (this.isRegistered) {
+				customMessageContainer.innerHTML = messageRegistered;
+			} else if (this.isTrialist) {
+				customMessageContainer.innerHTML = messageForTrial;
+			} else if (this.emailInUse) {
+				customMessageContainer.innerHTML = messageForEmailInUse;
```
umbobabo, line 342: As per my previous comment, this could code could be structured differently...

```suggestion
				customMessageContainer.innerHTML = `<h3>There is a problem configuring your commenting account</h3>
				<p>
					Please contact ${this.options.linkContact ? `<a class="linkContact" href='${this.options.linkContact}'>Customer Care</a>` : `Customer Care`} referencing issue code E403.
				</p>
			`;
```

components/o-comments/src/js/stream.js
```diff
@@ -370,7 +403,13 @@ class Stream {
...
 					asset_type: this.options.assetType,
 					uuid: this.options.articleId,
-					user_type: this.isTrialist ? 'trialist' : this.isRegistered ? 'registered' : 'anonymous'
+					user_type: this.emailInUse
+						? 'registered-email-in-use'
+						: this.isTrialist
+							? 'trialist'
+							: this.isRegistered
+							? 'registered'
+							: 'anonymous'
```
umbobabo, line 412: Thought: Again, this was like this before...it would likely benefit of creating a model and use that pushing the logic in the model and keep this part dumb till you reach something like the following:
```
user_type: user.type << A user model
```

umbobabo (approved):

components/o-comments/src/js/utils/auth.js
```diff
@@ -19,13 +19,19 @@ export default {
...
+				// TODO: CI-1493 to remove after subscriber only is not behind a flag - check on Q&A usage, see below
+				// response for when onlySubscribers is false and user is signed in but has no display name
+				// also used in live Q&A when a user is signed in but has no display name
 				if (response.status === 409) {
 					return { userHasValidSession: true };
 				}
 
+				// some users have coral accounts associated to old FT accounts that are using the same email address
+				// we need to remove this email before they can sign in to the new account
+				if (response.status === 403) {
```
umbobabo, line 31: Thought: Using only status code may not be a very scalable solution. What if tomorrow we introduce another condition that generates a `403` but we want a different message.
We could share the error code as part of a payload in the response.
Pseudo response...
```
    status-code: 403
    payload: {
       error-code: `CommentError403`, << Or whatever
       }
 ```
 In this way is easy to track the error between customer care, frontend and backend, you can even search the repos for this.

leannecornish-ft:
Closing in favour of https://github.com/Financial-Times/origami/pull/2102

PR #2111, opened 2025-05-16
feat: Add an outside-page icon helper to o3-button

notlee:
Initially used within a chrome extension


notlee:
We have this already!

PR #2134, opened 2025-06-09
fix: Allow 200+ financier display weights.

notlee:
So the editorial team can apply a very special 242 weight instead of 300.


rowanbeentje (approved):
Nice, so loosens the clamp and allows values from 200 to 800 inclusive 👍🏻

PR #2136, opened 2025-06-10
feat: Add font-size icon

notlee:
This is used on the app for font size options within the app article

Screenshot 2025-06-10 at 09 31 41


apaleslimghost (approved):
thank uuuuuuu ❤️

PR #2138, opened 2025-06-10
feat: Enable named exports for esm tokens.

notlee:
- Uses snake case: kebab case isn't support as a variable name in JS. Camel case makes for some awkward token names e.g. `o3_font_size_metric2_0` vs. `o3FontLineheightMetric20`
- Non-breaking by adding an extra export under `[brand]-named.js` as in named export.
- These a tree-shakeable for reduced bundle sizes.
- Separate JS token exports by icons and all other branded tokens: Technically this is breaking, as icons are removed from the branded token JS, but we have very high confidence this is not used in production.


rowanbeentje (approved):
This matches the usecases of the app pretty neatly, and I think having named exports for the icons is also something we might use! Given the search concerns you have I think this isn't a bad pattern to encourage - use the non-named exports generally unless size will be an issue, which it would only be likely to be for the icons 👍🏻

Thank you so much 🤩

PR #2151, opened 2025-06-17
fix: Update teaser topic tag typography

notlee:
- tag-suffix, tag, tag-prefix now all use body-base.
- If a tag prefix or suffix is a link, they use body-highlight.


j-mes (approved):
LGTM :+1:

apaleslimghost (approved):

PR #2373, opened 2026-02-26
[OR-2177] Migrate Typography docs

j-mes:
This work migrates Typography documentation from the website into a docs folder on o3-foundation.


j-mes:
> Do we think we can summarise the variables available into https://github.com/Financial-Times/origami/tree/main/components/o3-foundation#typography?

I think we can do that instead. Also maybe condense the documentation quite a lot. I think it'd be more helpful to have a table like how Spacing is but for Typography.

PR #2422, opened 2026-04-21
OR-2324:remove origami dictionary

frshwtr:
* removes Origami Dictionary 
* removes Origami Dictionary's supporting components, `o3-figma-sb-link` and `o3-tooling-token`


j-mes (approved):
Approved for the timely release later…

